PR #35, opened 2019-05-06
Checkpointing/msgpack

chrisfernandes18:
Check out the sample-code directory I made and look at the load.c and save.c modules


borjasotomayor:
@chrisfernandes18: When creating a pull request where you are soliciting a review from someone before merging into your dev branch, make sure to add those people as reviewers, not as assignees (unless I'm mistaken, it looks like you want @jarvislam1999 to take a look at this code). Assignees are the people responsible for creating the pull request (and for making sure it is completed). Reviewers are the people who will actually be doing the code review.

mpcsh:
@chrisfernandes18 @jonahdf: while it's fine for now, please do not check-in a git repository within a git repository (in this case, the `msgpack-c` directory is the culprit). For including another project's code, use *submodules*. You can read about them here: https://git-scm.com/book/en/v2/Git-Tools-Submodules

That said, since it looks like the team has decided on using `protobuf`, let's actually revert this merge and leave only the `protobuf` samples on the team dev branch. If we're not intending to send this code to `master`, it shouldn't be merged into `checkpointing/dev` (you should view the team dev branch as a staging ground).

PR #37, opened 2019-05-06
Protobuf

schin99:


chrisfernandes18:
Your code is so much shorter than mine, but it looks like protobuf uses much more high-level packing and unpacking than msgpack. I think it's probably better than msgpack

mpcsh:
@chrisfernandes18: make sure to leave an approving review if you do indeed approve of this PR! (it's okay to do so retroactively; but do get used to doing that beforehand)

@schin99: be sure to not commit binary files (i.e. files produced by `gcc`, `make`, etc). This PR includes the `serialize` binary. Once we have a Makefile, we will to set up a `.gitignore` file so that git simply ignores binaries. But for now, please push a commit directly on `checkpointing/dev` that deletes said binary (by using `git rm`).

PR #54, opened 2019-05-07
Amatai/add to authors

a-matai:
pls accept my name in this game


lydiafilipe:
Hi Anant, this isn't quite what we're looking for. Firstly, the add to authors pull request should ONLY be adding your name to AUTHORS.md This pr seems to include code from this first sprint, which is a separate thing and not necessarily ready for being merged into master.

Another thing- make sure that the people you want to review it are under "Reviewers", not "Assignees". Assigness would be who worked on it- which in this case is you.

It might be easiest to just close this pr and do another one with just the AUTHORS.md changes

PR #61, opened 2019-05-09
Cli/ryangold/recommand

Gestalt-R:
Needs some comments cleared up, but it is otherwise stable


Tianlel:
After talking to Ryan, things make much more sense to me. 
Since we cannot assume a fixed list of actions, enum probably should not be a field in our struct.
Ryan will implement the dispatch table later.
I think it's good to be merged.

lydiafilipe:
Issue Score: ✓-

Comments:
Good work on the pr. However, the title and summary are unclear, and there is no closing statement. (Sorry for grading this late, I know this is from an earlier sprint and you all have since incorporated feedback like this :) )

PR #68, opened 2019-05-10
 Updated action structs to reflect Sprint 3 Discussion (5/10)

yoonsjeong:
There were a number of issues that we decided on during Sprint 3 such as:

- splitting actions into "kinds", or types so that we could better manage them
- placing the synonyms and canonical names within the struct so that they could be accessed by other teams
- creating a "get_supported_actions()" function

I've added these functions to the document and would appreciate any feedback!

### DIFF
--- a/src/action_management/include/actionmanagement.h
+++ b/src/action_management/include/actionmanagement.h
@@ -45,7 +45,7 @@ enum object_type {
 
 /* An action struct that contains the following: 
  * - act: the encoded enum name for this action
- * c_name - the 'canonical' string that should call the enum
+ * - c_name: the 'canonical' string that should call the enum
  * synonyms - the synonyms that would also be allowed to call action
  * parameters - an order-sensitive linked list of object_type enums
 */
--- a/src/action_management/include/actionmanagement.h
+++ b/src/action_management/include/actionmanagement.h
@@ -32,22 +32,20 @@ enum action_type {
 		"turn on", "turn off", and "go" */
   ACT_INVENTORY /* includes "take", "drop", "consume" */
 };
-  
+
+enum object
+
 
 /* An action struct that contains the following: 
  * act: A tag corresponding to the action 
  * type: A tag describing the type of action
- * num_arg: the number of actions required for the action to be valid
+ * parameters - array of enum inputs
 */
-struct action_struct {
+typedef struct {
   enum actions act;
   enum action_type type;
-  int num_arg; 
-};
-
-
-/* A type declaration of the action struct to action_t. */
-typedef struct action_struct action_t;
+  int *parameters;
+} action_t;
 
 
 /* 
--- a/src/action_management/include/actionmanagement.h
+++ b/src/action_management/include/actionmanagement.h
@@ -46,7 +46,7 @@ enum object_type {
 /* An action struct that contains the following: 
  * - act: the encoded enum name for this action
  * - c_name: the 'canonical' string that should call the enum
- * synonyms - the synonyms that would also be allowed to call action
+ * - synonyms: the synonyms that would also be allowed to call action
  * parameters - an order-sensitive linked list of object_type enums
 */
 typedef struct {
--- a/src/action_management/include/actionmanagement.h
+++ b/src/action_management/include/actionmanagement.h
@@ -5,48 +5,58 @@
 #include <stdlib.h>
 
 
-/* An enumeration of all supported actions. */
-/* All synonyms will be provided to CLI. */
+/* An enumeration of all supported actions.
+ * TYPE 1 ACTIONS - ACTION <item>
+ * TYPE 2 ACTIONS - ACTION <direction>
+ * TYPE 3 ACTIONS - ACTION <npc>
+ * TYPE 4 ACTIONS - ACTION <item> <npc> */
 enum actions {
-  OPEN, /* world action */
-  CLOSE, /* world action */
-  PUSH, /* world action */
-  PULL, /* world action */
-  EXAMINE, /* synonymous to "look at", is a world action */
-  TURN_ON, /* world action */
-  TURN_OFF, /* world action */
-  TALK_TO, /* NPC action */
-  GIVE, /* requires 2 arguments separated by 'to',  NPC action */
-  GO, /* world action */
-  TAKE, /* synonymous with "pick up", is an inventory action */
-  DROP, /* inventory action */
-  CONSUME, /* synonymous with "use", "drink", "eat", is an inventory action */
-  USE_ON, /* requires two arguments separated by 'on', world and NPC action  */
+
+/* TYPE 1 ACTIONS */
+	OPEN,
+	CLOSE,
+	PUSH,
+	PULL,
+	EXAMINE,
+	TURN_ON,
+	TURN_OFF,
+	TAKE,    // synonyms: "pick up"
+	DROP, 
+	CONSUME, // synonyms: "use" "drink" "eat"
+
+/* TYPE 2 ACTIONS */
+	GO, 	 // synonyms: "walk" "enter"
+
+/* TYPE 3 ACTIONS */
+	TALK_TO,
+
+/* TYPE 4 ACTIONS */
+	GIVE_TO,
+	USE_ON
 };
 
-
-/* An enumeration of all possible action types. */
-enum action_type {
-  ACT_NPC, /* includes "give", "talk to" */
-  ACT_WORLD, /* includes "look at", "examine", "open", "close", "push", "pull", 
-		"turn on", "turn off", and "go" */
-  ACT_INVENTORY /* includes "take", "drop", "consume" */
+/* Outlines the three different objects that can be action'd upon */
+enum object_type {
+	ITEM_OBJ, 	   // e.g. door, potion
+	DIRECTION_OBJ, // e.g. north, east
+	NPC_OBJ 	   // e.g. Borja, wizard
 };
-
-enum object
 
 
 /* An action struct that contains the following: 
- * act: A tag corresponding to the action 
- * type: A tag describing the type of action
- * parameters - array of enum inputs
+ * act - the encoded enum name for this action
+ * c_name - the 'canonical' string that should call the enum
+ * synonyms - the synonyms that would also be allowed to call action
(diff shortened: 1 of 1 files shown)


src/action_management/include/actionmanagement.h
```diff
@@ -5,50 +5,58 @@
...
+	PULL,
+	EXAMINE,
+	TURN_ON,
+	TURN_OFF,
+	TAKE,    // synonyms: "pick up"
+	DROP, 
+	CONSUME, // synonyms: "use" "drink" "eat"
+
+/* TYPE 2 ACTIONS */
+	GO, 	 // synonyms: "walk" "enter"
```
MTUC08: would "enter" necessarily be a synonym here? Since you can go north, but you can't really enter north.

src/action_management/include/actionmanagement.h
```diff
@@ -5,50 +5,58 @@
 #include <stdlib.h>
 
 
-/* An enumeration of all supported actions. */
-/* All synonyms will be provided to CLI. */
+/* An enumeration of all supported actions.
```
MTUC08: I like that you split the actions into categories based on what inputs they take, since one of the issues we had in sprint 1 was that there was a lot of overlap when categorizing.

src/action_management/include/actionmanagement.h
```diff
@@ -5,50 +5,58 @@
...
-/* An action struct that contains the following: 
- * act: A tag corresponding to the action 
- * type: A tag describing the type of action
- * num_arg: the number of actions required for the action to be valid
-*/
-struct action_struct {
-  enum actions act;
-  enum action_type type;
-  int num_arg; 
+/* Outlines the three different objects that can be action'd upon */
```
MTUC08: To clarify, is this a prospective object struct? Since Game State handles objects?

src/action_management/include/actionmanagement.h
```diff
@@ -5,50 +5,58 @@
...
-typedef struct action_struct action_t;
+/* An action struct that contains the following: 
+ * act - the encoded enum name for this action
+ * c_name - the 'canonical' string that should call the enum
+ * synonyms - the synonyms that would also be allowed to call action
+ * parameters - an order-sensitive linked list of object_type enums
+*/
+typedef struct {
+	enum actions act;   // e.g. CONSUME
+	char *c_name; 		// e.g. "eat"
```
MTUC08:
```suggestion
	char *c_name;  // e.g. "eat"
```

src/action_management/include/actionmanagement.h
```diff
@@ -5,50 +5,58 @@
...
+	ITEM_OBJ, 	   // e.g. door, potion
+	DIRECTION_OBJ, // e.g. north, east
+	NPC_OBJ 	   // e.g. Borja, wizard
 };
 
 
-/* A type declaration of the action struct to action_t. */
-typedef struct action_struct action_t;
+/* An action struct that contains the following: 
+ * act - the encoded enum name for this action
```
MTUC08:
```suggestion
 * - act: the encoded enum name for this action
```

src/action_management/include/actionmanagement.h
```diff
@@ -5,50 +5,58 @@
...
+ * c_name - the 'canonical' string that should call the enum
```
MTUC08:
```suggestion
 * - c_name: the 'canonical' string that should call the enum
```

src/action_management/include/actionmanagement.h
```diff
@@ -5,50 +5,58 @@
...
+ * synonyms - the synonyms that would also be allowed to call action
```
MTUC08:
```suggestion
 * - synonyms: the synonyms that would also be allowed to call action
```

src/action_management/include/actionmanagement.h
```diff
@@ -5,50 +5,58 @@
...
+ * parameters - an order-sensitive linked list of object_type enums
```
MTUC08:
```suggestion
 * - parameters: an order-sensitive linked list of object_type enums
```

MTUC08: What is the significance of order sensitivity?

src/action_management/include/actionmanagement.h
```diff
@@ -5,50 +5,58 @@
...
+*/
+typedef struct {
+	enum actions act;   // e.g. CONSUME
+	char *c_name; 		// e.g. "eat"
+	list_t *synonyms;   // e.g. "drink" -> "use"
+	list_t *parameters; // e.g. ITEM_OBJ -> NPC_OBJ
```
MTUC08: Can you explain this example a bit more? I'm not entirely sure what it means.

src/action_management/include/actionmanagement.h
```diff
@@ -59,7 +67,7 @@ typedef struct action_struct action_t;
  * Returns:
```
MTUC08: I can't tell if the new Parameters have been added to this function (GitHub is not allowing me to click and edit)

src/action_management/include/actionmanagement.h
```diff
@@ -89,227 +97,55 @@ int action_init(action_t *a, enum actions act);
...
- * - o: a pointer to an object struct given by game state
- * - r: a pointer to a room struct given by game state
- * 
- * Returns: 
- * - 0 if success, 1 if failure
- * - a message to stderr indicating whether the action was a success or failure
- */ 
-int act_w_turn_on_off(action_struct *a, object_struct *o, room_struct *r);
+/* A function that executes TYPE 1 actions (ACTION <item>) 
+ * Returns 0 on success, 1 on failure. */
```
MTUC08: Make sure to include parameters and what it returns

src/action_management/include/actionmanagement.h
```diff
@@ -89,227 +97,55 @@ int action_init(action_t *a, enum actions act);
...
- * Returns: 
- * - 0 if success, 1 if failure
- * - a message to stderr describing the object
- * - a message to stderr indicating whether the action was a success or failure
- */ 
-int act_w_examine(action_struct *a, object_struct *o);
+/* A function that returns all supported actions
+ * Returns:
+ * a linked list of action_t structs */
+list_t *get_supported_actions();
```
MTUC08: Why is this function necessary?

MTUC08: Also make sure to include parameters (I think it may need an object or something to pass otherwise how does it know what is supported?). Or is this returning all of the actions?

src/action_management/include/actionmanagement.h
```diff
@@ -89,227 +97,55 @@ int action_init(action_t *a, enum actions act);
...
- * - o: a pointer to an object struct given by game state
- * - n: a pointer to an npc struct given by game state
- * 
- * Returns: 
- * - 0 if success, 1 if failure
- * - messages to stderr providing information held by the npc
- * - a message to stderr indicating whether the action was a success or failure
- */ 
-int act_n_talk_to(action_struct *a, object_struct *o, npc_struct *n);
+/* A function that executes TYPE 2 actions (ACTION <direction>)
```
MTUC08: Might be good to add parameters and returns in the same format is before (stylistic)

lucyyutingli:
I know the title says it all but all issues need a description @yoonsjeong @ellychoi @MTUC08

src/action_management/include/actionmanagement.h
yoonsjeong: that's true -- in that case should we have another type 2 action called ENTER?
for commands like, say "ENTER house" or "ENTER cave"

src/action_management/include/actionmanagement.h
yoonsjeong: this one is a function Borja recommended that would be used by other groups to see what actions are available. I don't think it would require an input, since we would simply be returning a list of the actions that we support which does not depend on an input

src/action_management/include/actionmanagement.h
yoonsjeong: yes I think it would go in the object struct. I wasn't sure where to put it though since it seemed relevant to our 'types'

src/action_management/include/actionmanagement.h
MTUC08: I was thinking that if there was a possibility to enter, the game would automatically move the player into the new room

src/action_management/include/actionmanagement.h
yoonsjeong: ACTION <item> <npc> is different from 
ACTION <npc> <item> so the list would have to reflect the order

src/action_management/include/actionmanagement.h
yoonsjeong: we have a canonical name, which would be the associated name that is primary to the action, then synonyms which could be accepted or not for specific items "DRINK potato" would not work despite DRINK being synonym to CONSUME.
This is really just for other groups' convenience, so that they can call and see what synonyms are available to each action.

MTUC08: In particular what is parameters? Is it the inputs to the function?

src/action_management/include/actionmanagement.h
yoonsjeong: if there were multiple things one could enter though that could cause problems. if there was say a house and a cave but we wanted to enter the house how would we do that?

src/action_management/include/actionmanagement.h
yoonsjeong: oh yeah, it would be a list of enumerated type OBJ, (the new struct that contains ITEM_OBJ NPC_OBJ and DIRECTION_OBJ)

so for this action "ACTION item npc" it would be a list that looks like
(ITEM_OBJ) -> (NPC_OBJ)

src/action_management/include/actionmanagement.h
yoonsjeong: I'll do that!

src/action_management/include/actionmanagement.h
yoonsjeong: ill do this

src/action_management/include/actionmanagement.h
MTUC08: Yeah that’s fair let’s add an enter action then

PR #70, opened 2019-05-12
Checkpointing/protofile

jonahdf:
Jarvis and I created the .proto file specifying the structures needed to save the game. For guidance, we used the dev branch of Game State, along with the newer "room-connections" branch. Since Game State is still determining the structures, there are some incomplete parts, but it should be a good base for future changes. We also avoided infinite recursion errors by replacing pointers to rooms, paths, etc. with IDs (instead of the room/path itself). The biggest challenges ahead will be extracting from hash tables and linked lists.


borjasotomayor:
@schin99: In the future, please explicitly approve the review request before merging (merging does not automatically indicate that you've approved the request). Otherwise, the pull request will show up as having a pending review.

PR #72, opened 2019-05-13
Merging Type 3,4,5 i.e. action_npc, action_item_npc, action_item_item

MTUC08:
Just finished the implementation for action_npc, action_item_npc, and action_item_item. Can you take a look and let me know if you'd like me to change anything?

### DIFF
--- a/src/action_management/include/actionmanagement.h
+++ b/src/action_management/include/actionmanagement.h
@@ -42,11 +42,11 @@ enum actions {
 
 /* Each enum corresponds to a different "KIND" of action */
 enum action_kind {
-  ITEM, // ACTION <item> i.e. Action Type 1
-  DIRECTION, // ACTION <direction i.e. Action Type 2
-  NPC, // ACTION <npc> i.e. Action Type 3
-  ITEM_NPC, // ACTION <item> <npc> i.e. Action Type 4
-  ITEM_ITEM // ACTION <item> <item> i.e. Action Type 5
+  ITEM = 1, // ACTION <item> i.e. Action Type 1
+  DIRECTION = 2, // ACTION <direction i.e. Action Type 2
+  NPC = 3, // ACTION <npc> i.e. Action Type 3
+  ITEM_NPC = 4, // ACTION <item> <npc> i.e. Action Type 4
+  ITEM_ITEM = 5 // ACTION <item> <item> i.e. Action Type 5
 };
 
 
@@ -197,8 +197,8 @@ int action_direction(game_t *g, action_t *a, direction_t *d);
  * - n: An npc struct
  * 
  * Returns
- * - 1 on success
- * - 0 on failure
+ * - 0 on success
+ * - 1 on failure
  */
 int action_npc(action_t *a, npc_t *n);
 
@@ -206,31 +206,31 @@ int action_npc(action_t *a, npc_t *n);
 /* A function that executes KIND 4 actions (ACTION <item> <npc>)
  * 
  * Parameters:
- * - p: A player struct containing an inventory
+ * - g: A game struct containing a field for the current player
  * - a: An action struct
  * - i: An item struct
  * - n: An npc struct containing an inventory
  * 
  * Returns
- * - 1 on success
- * - 0 on failure
+ * - 0 on success
+ * - 1 on failure
  */
-int action_item_npc(player_t *p, action_t *a, item_t *i, npc_t *n);
+int action_item_npc(game_t *g, action_t *a, item_t *i, npc_t *n);
 
 
 /* A function that executes KIND 5 actions (ACTION <item> <item>)
  * 
  * Parameters:
- * - p: A player struct containing an inventory
+ * - g: A game struct containing a field for the current player
  * - a: An action struct
  * - direct: An item struct containing the direct object (the "actor")
  * - indirect: An item struct containing the indirect object (the "actee")
  * 
  * Returns
- * - 1 on success
- * - 0 on failure
+ * - 0 on success
+ * - 1 on failure
  */
-int action_item_item(player_t *p, action_t *a, item_t *direct, item_t *indirect);
+int action_item_item(game_t *g, action_t *a, item_t *direct, item_t *indirect);
 
 
 #endif
--- a/src/action_management/include/actionmanagement.h
+++ b/src/action_management/include/actionmanagement.h
@@ -3,7 +3,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
-
+#include "game.h"
 
 /* An enumeration of all supported actions.
  * KIND 1 ACTIONS - ACTION <item>
@@ -34,19 +34,19 @@ enum actions {
 
 /* KIND 4 ACTIONS */
 	GIVE_TO,
-
+	
 /* KIND 5 ACTIONS */
-    USE_ON
+	USE_ON
 };
 
 
 /* Each enum corresponds to a different "KIND" of action */
 enum action_kind {
-    ITEM = 1, // ACTION <item>
-    DIRECTION = 2, // ACTION <direction
-    NPC = 3, // ACTION <npc>
-    ITEM_NPC = 4, // ACTION <item> <npc>
-    ITEM_ITEM = 5 // ACTION <item> <item>
+  ITEM, // ACTION <item> i.e. Action Type 1
+  DIRECTION, // ACTION <direction i.e. Action Type 2
+  NPC, // ACTION <npc> i.e. Action Type 3
+  ITEM_NPC, // ACTION <item> <npc> i.e. Action Type 4
+  ITEM_ITEM // ACTION <item> <item> i.e. Action Type 5
 };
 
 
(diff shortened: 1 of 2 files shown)


src/action_management/include/actionmanagement.h
```diff
@@ -34,19 +34,19 @@ enum actions {
...
-    ITEM = 1, // ACTION <item>
-    DIRECTION = 2, // ACTION <direction
-    NPC = 3, // ACTION <npc>
-    ITEM_NPC = 4, // ACTION <item> <npc>
-    ITEM_ITEM = 5 // ACTION <item> <item>
+  ITEM, // ACTION <item> i.e. Action Type 1
+  DIRECTION, // ACTION <direction i.e. Action Type 2
+  NPC, // ACTION <npc> i.e. Action Type 3
+  ITEM_NPC, // ACTION <item> <npc> i.e. Action Type 4
+  ITEM_ITEM // ACTION <item> <item> i.e. Action Type 5
```
yoonsjeong: is there a particular reason you changed this so the enums aren't represented by the appropriate action kind number?
I felt like that would make it more intuitive to use

src/action_management/include/actionmanagement.h
```diff
@@ -193,46 +193,44 @@ int action_direction(game_t *g, action_t *a, direction_t *d);
...
  * 
  * Parameters:
- * - g: A game struct consisting of the game state
  * - a: An action struct
  * - n: An npc struct
  * 
  * Returns
- * - 0 on success
- * - 1 on failure
+ * - 1 on success
```
yoonsjeong: i think generally 0 is used for success, whereas 1, 2, 3, etc can be used for alternative error codes
is it alright if we stick to using 0 as the success code for consistency?

src/action_management/include/actionmanagement.h
```diff
@@ -193,46 +193,44 @@ int action_direction(game_t *g, action_t *a, direction_t *d);
...
  */
-int action_npc(game_t *g, action_t *a, npc_t *n);
+int action_npc(action_t *a, npc_t *n);
 
 
 /* A function that executes KIND 4 actions (ACTION <item> <npc>)
  * 
  * Parameters:
- * - g: A game struct consisting of the game state
+ * - p: A player struct containing an inventory
```
yoonsjeong: I think we should stick to the previous use of game_struct rather than player_struct, since the player struct exists within the game_struct (see game state's code:)
https://github.com/uchicago-cs/chiventure/blob/game-state/dev/src/game-state/include/game.h

this follows for all the other functions

src/action_management/src/actionmanagement.c
```diff
@@ -1,50 +1,120 @@
...
-    action_t *a = malloc(sizeof(action_t));
-
-    if(a == NULL)
+		     list_t *synonyms, enum kind)
+{
+  action_t *a = malloc(sizeof(action_t));
+  
+  if(!a)
     {
       error("Could not allocate memory");
```
yoonsjeong: I think all the error() funcs should be perror() instead -- I couldn't find an error() function for c but I did find a perror() func:
https://www.geeksforgeeks.org/error-handling-c-programs/

src/action_management/src/actionmanagement.c
```diff
@@ -1,50 +1,120 @@
...
-  list_t *synonyms, enum kind)
-  {
-    action_t *a = malloc(sizeof(action_t));
-
-    if(a == NULL)
+		     list_t *synonyms, enum kind)
+{
+  action_t *a = malloc(sizeof(action_t));
+  
+  if(!a)
```
yoonsjeong: for documentation purposes it seems like it would be better to keep it as (a == NULL)

src/action_management/src/actionmanagement.c
```diff
@@ -1,50 +1,120 @@
...
+    error("The action provided is not of the correct type");
+    return 0;
+  }
+  /* function needs to be implemented by game state 
+   * takes in a player, npc, and item
+   * removes object from inventory
+   * frees object if necessary
+   * returns 1 if success, 0 if failed
+   */
+  int given = remove_inventory_object(p, i);
```
yoonsjeong: we could write it as
int given = remove_inventory_object(p, i) == SUCCESS;
but the way it's currently written requires that remove return a non zero value

src/action_management/include/actionmanagement.h
MTUC08: A) I wasn't sure if that was allowed
B) By assigning numbers to each enum isn't that defeating the purpose of the enum?
I wasn't really sure what the Direction = 1 etc was doing

src/action_management/include/actionmanagement.h
MTUC08: I just thought that it made it more intuitive when reading the code, i.e. int action = action_item_item; if (item) <code>

src/action_management/include/actionmanagement.h
MTUC08: I think that cli would handle passing the correct player to us, so our function only has to deal with a certain player. As action management, I don't think we'd ever need to deal with the entire game struct ourselves

src/action_management/src/actionmanagement.c
MTUC08: the error function is straight from the other labs we've done but I can change it to fprintf(stderr) if you want?

src/action_management/src/actionmanagement.c
MTUC08: yeah that's fine

src/action_management/src/actionmanagement.c
MTUC08: im not sure what you mean

src/action_management/include/actionmanagement.h
yoonsjeong: enums do have the capability of being represented by other numbers, see this:
https://www.geeksforgeeks.org/enumeration-enum-c/

The reason for the numbering was because of the action "kind" each represented; so for example, DIRECTION = 2 because action kind 2 is "ACTION direction"
This would make it so that if I wanted to check if I was working with action kind 3, i could use either (action_kind x == NPC) or I could use (action_kind x == 3)

in the larger scope of the project it's a small thing so I won't push for it if you'd rather not, but I do think there is some merit to using this system

src/action_management/include/actionmanagement.h
yoonsjeong: this was actually addressed by Borja during sprint 3, which is why I ended up changing it despite our previous agreement to use player struct prior to this.
he told us a lot of games tend to take a "context variable" as input which would allows the function to access and change all kinds of information within in thegame

src/action_management/src/actionmanagement.c
MTUC08: ok I see what you're saying, I'll change it

src/action_management/include/actionmanagement.h
MTUC08: Ok yeah that makes sense, I had never seen it done before so I was confused. I'll change it back

src/action_management/include/actionmanagement.h
MTUC08: never mind I realized what you were saying

src/action_management/include/actionmanagement.h
MTUC08: but then we need a player index so we know which player's inventory is being altered

MTUC08: and wouldn't that do the same as if cli did it? except instead of them doing pass game->player[2] they'd do pass game, pass 2

src/action_management/src/actionmanagement.c
yoonsjeong: yeah that might be best. I don't personally remember using error() and it seems it isn't a standard function 
https://stackoverflow.com/questions/39137342/what-library-is-the-error-function-in

src/action_management/include/actionmanagement.h
yoonsjeong: I would assume that we could take the NPC_struct input to figure out which inventory to alter (and if there is no NPC_struct input, we can assume we're altering the main player inventory, which would probably have an unchanging index)
in this case I'd rather err on the side of following what was discussed during the sprint meeting

MTUC08: I wasn't present at the spring meeting, and without an additional input there's now way to code for the correct inventory to access

MTUC08: resolved by assuming that game_t will have a current_player field

MTUC08:
Closing pull request to implement changes without having to merge unfinished code to actions/dev

MTUC08:
Trying to merge changed files into actions/dev

mpcsh:
PR Score: ✓-

Comments:
The title shouldn't include "Merging". In general, don't duplicate in PR titles what is evident through Github's UI (i.e. anything about "merging" or the fact that this is a PR, branch names, etc). The title and summary should also be a little more high-level; the specific kind numbers and names aren't really needed. For the review, Yoon brought great feedback, but should've used "Request changes" instead of "Comment without approval". Next, the close and reopen weren't necessary. Finally and most importantly, the PR should not have been merged before the reviewer gave their explicit approval.

PR #73, opened 2019-05-13
Altered a bit of existing code, added "get" functions, and "kind 2" function

yoonsjeong:
I added the get functions to actionmanagement.c.
There were a couple of small things that I noticed needed fixing, feel free to comment on whether or not they have been addressed properly.
I added a kind 2 function, but also slightly altered the kind 3 function after it -- I would recommend checking on that too.


yoonsjeong:
Oh -- also I noticed that there were some functions that were tabbed by 2 spaces, so if you notice a lot of changes in certain areas without much alteration in text, it's probably the addition of more spaces

src/action_management/include/actionmanagement.h
```diff
@@ -193,14 +193,15 @@ int action_direction(game_t *g, action_t *a, direction_t *d);
...
+ * - g: A game struct
  * - a: An action struct
  * - n: An npc struct
  * 
  * Returns
  * - 0 on success
  * - 1 on failure
  */
-int action_npc(action_t *a, npc_t *n);
+int action_npc(game_t *g, action_t *a, npc_t *n);
```
MTUC08: I don't think this should take in a game_t struct because it doesn't affect the game state at all. It only spits out whatever dialogue an npc has, and can do that multiple times, and doesn't change anything else in the game, so the game_t struct wouldn't be used within the function at all. 
```suggestion
int action_npc(action_t *a, npc_t *n);
```

src/action_management/src/actionmanagement.c
```diff
@@ -8,52 +8,100 @@
...
+    assert(a);
+    assert(a->c_name);
+    assert(a->synonyms);
+    free(a->c_name);
+    free(a->synonyms);
+    free(a);
+    return SUCCESS;
+}
+
+/* A function that returns the canonical name of an action as a string */
```
MTUC08:
```suggestion
/* See actionmanagement.h */
```

src/action_management/src/actionmanagement.c
```diff
@@ -8,52 +8,100 @@
...
+char *get_action_cname(action_t *a)
+{
+    return a->c_name;
 }
 
+/* A function that returns the canonical name of an action as a string */
```
MTUC08:
```suggestion
/* See actionmanagement.h */
```

src/action_management/src/actionmanagement.c
```diff
@@ -8,52 +8,100 @@
...
+list_t *get_action_synonyms(action_t *a)
+{
+    return a->synonyms;
+}
+
+/* A function that returns the KIND of an action as an integer
```
MTUC08:
```suggestion
/* See actionmanagement.h */
```

src/action_management/src/actionmanagement.c
```diff
@@ -8,52 +8,100 @@
...
+ * Returns 1 for KIND 1, 2 for KIND 2, etc.
```
MTUC08:
```suggestion
 
```

src/action_management/src/actionmanagement.c
```diff
@@ -8,52 +8,100 @@
...
+ */
```
MTUC08:
```suggestion

```

src/action_management/src/actionmanagement.c
```diff
@@ -8,52 +8,100 @@
...
+int *get_action_kind(action_t *a)
+{
+    return a->kind;
+}
+
+/* A function that returns all supported actions
```
MTUC08: It should all be /* See actionmanagement.h */ for stylistic consistence

src/action_management/src/actionmanagement.c
```diff
@@ -8,52 +8,100 @@
...
+ * Parameters:
+ * - kind: The enumerated value of KIND, specifying which action KIND necessary
+ * Returns:
+ * - a linked list of action_t structs
+ */
+list_t *get_supported_actions(enum action_kind *kind)
+{
+    int kind = get_action_kind(kind);
```
MTUC08: from above, get_action_kind takes in an action_t struct, not an enum.

MTUC08: What does this line do exactly?

src/action_management/src/actionmanagement.c
```diff
@@ -8,52 +8,100 @@
...
+    switch(kind)
+    {
+        case ITEM:
+            list_t *actions_kind1 = get_actions_kind1();
```
MTUC08: do we assume that get_actions_kind1, etc. are created by game state? Or are you adding those as well?

src/action_management/src/actionmanagement.c
```diff
@@ -65,38 +113,67 @@ int action_item(game_t *g, action_t *a, item_t *i)
...
 	}
 	/* function needs to be implemented by game state
 	 * takes in a player, npc, and item
 	 * removes object from inventory
 	 * frees object if necessary
-	 * returns 1 if success, 0 if failed
+	 * returns 0 if success, 1 if failed
 	 */
 	int given = remove_inventory_object(g->current_player, i);
+    // THIS IS A non-existent FUNCTION we depend on GAME STATE FOR
```
MTUC08: Doesn't action_item also have to account for toggling (i.e. pull push, open, close), and adding to inventory (pick up) etc?

src/action_management/src/actionmanagement.c
```diff
@@ -65,38 +113,67 @@ int action_item(game_t *g, action_t *a, item_t *i)
...
+    } else {
+        fprintf(stderr, "Player was unable to move.\n");
+        return FAILURE;
+    }
+}
+
 //KIND 3
 /* See actionmanagement.h */
-int action_npc(action_t *a, npc_t *n)
+int action_npc(game_t *g, action_t *a, npc_t *n)
```
MTUC08: Again, I don't think game_t is necessary b/c it isn't ever used within the function
```suggestion
int action_npc(action_t *a, npc_t *n)
```

src/action_management/src/actionmanagement.c
```diff
@@ -65,38 +113,67 @@ int action_item(game_t *g, action_t *a, item_t *i)
...
 {
+    assert(g);
```
MTUC08:
```suggestion
   
```

src/action_management/include/actionmanagement.h
ellychoi: I agree that we shouldn't have a game_t parameter because it doesn't affect the game state. However, I do think we should at least have a player struct so that we can pass in the player and how the player can react to dialogue.

src/action_management/src/actionmanagement.c
```diff
@@ -8,52 +8,100 @@
...
+    return a->kind;
+}
+
+/* A function that returns all supported actions
+ * Parameters:
+ * - kind: The enumerated value of KIND, specifying which action KIND necessary
+ * Returns:
+ * - a linked list of action_t structs
+ */
+list_t *get_supported_actions(enum action_kind *kind)
```
ellychoi: is the parameter being passed not the kind that we should be using instead of calling get kind?

src/action_management/src/actionmanagement.c
```diff
@@ -65,38 +113,67 @@ int action_item(game_t *g, action_t *a, item_t *i)
...
 		return FAILURE;
 	}
 	/* function needs to be implemented by game state
 	 * takes in a player, npc, and item
 	 * removes object from inventory
 	 * frees object if necessary
-	 * returns 1 if success, 0 if failed
+	 * returns 0 if success, 1 if failed
 	 */
 	int given = remove_inventory_object(g->current_player, i);
```
ellychoi: Once we talk with game state further, I can think we can refine kind 1 so that it can account for something that simply changes the game state and something that specifically changes the inventory and how the game reacts to it.

src/action_management/src/actionmanagement.c
ellychoi: I agree with this, but maybe we should still have a player parameter.

src/action_management/src/actionmanagement.c
```diff
@@ -65,38 +113,67 @@ int action_item(game_t *g, action_t *a, item_t *i)
...
   assert(n);
   if (a->kind != NPC) {
-    fprintf(stderr, "The action provided is not of the correct type");
+    fprintf(stderr, "The action provided is not of the correct type.\n");
     return FAILURE;
   }
-  fprintf(stderr, "%s", n->dialogue);
-  return SUCCESS;
+
+  int talked = player_talk(n); // THIS IS A non-existent FUNCTION we depend on GAME STATE FOR
```
ellychoi: We should discuss with game state how we can have conversations and/or dialogue to choose from to respond to a given statement.

src/action_management/src/actionmanagement.c
```diff
@@ -65,38 +113,67 @@ int action_item(game_t *g, action_t *a, item_t *i)
...
 
 }
 
+//KIND 2
+/* See actionmanagement.h */
+int action_direction(game_t *g, action_t *a, direction_t *d)
+{
+    assert(g);
+    assert(a);
+    assert(n);
```
ellychoi: what exactly is n referring to?

src/action_management/src/actionmanagement.c
```diff
@@ -65,38 +113,67 @@ int action_item(game_t *g, action_t *a, item_t *i)
...
+    if (a->kind != DIRECTION) {
```
ellychoi: we should also check to make sure that it is a valid direction(so that we dont go out of bounds or wacky directions). this will also be refined once we figure out how the room is mapped.

src/action_management/src/actionmanagement.c
```diff
@@ -140,12 +217,12 @@ int action_item_item(player_t *p, action_t *a,
   assert(direct);
   assert(indirect);
   if (a->kind != ITEM_ITEM) {
-    fprintf(stderr, "The action provided is not of the correct type");
+    fprintf(stderr, "The action provided is not of the correct type.\n");
     return FAILURE;
   }
   int moved = remove_inventory_object(g->current_player, i);
```
ellychoi: we need another function that tells us if the action can be done again (ex once you use a key on a door can it be used again?)

yoonsjeong: I would imagine that the door's game state would be altered such that you could no longer perform the action on it. So rather than having a function tell us if the action can be performed again, just having the function check this state might be better

src/action_management/src/actionmanagement.c
yoonsjeong: n refers to the NPC variable, one of the inputs to this function

src/action_management/src/actionmanagement.c
yoonsjeong: this would be checked in the player_move function which comes out later in the function

src/action_management/include/actionmanagement.h
yoonsjeong: the player struct is within the game struct though; also, the game state should be affected, since talking to certain NPCs should be able to alter the game in some way

src/action_management/src/actionmanagement.c
MTUC08: I think game state would handle that

src/action_management/src/actionmanagement.c
yoonsjeong: yeah this should not be in the code, it's a mistake

src/action_management/src/actionmanagement.c
MTUC08:
```suggestion
  int talked = player_talk(g->current_player, n); // THIS IS A non-existent FUNCTION we depend on GAME STATE FOR
```

src/action_management/src/actionmanagement.c
yoonsjeong: no--this is where I had questions about whether we should have a separate get_actions.c file, and also wanted to clarify with borja on how we would be creating this kind of function

yoonsjeong: for now it's just a filler function

src/action_management/src/actionmanagement.c
MTUC08: I think I see what you were saying with the player. So by passing game_t we can pass the current player and have a conversation

src/action_management/src/actionmanagement.c
yoonsjeong: yes, I didn't add that yet since for this chunk I was just working on the Kind 2 funcs and the Get funcs and I figured it would require more discussion

src/action_management/src/actionmanagement.c
MTUC08: Also I'd expect most games to have a one use key

src/action_management/src/actionmanagement.c
MTUC08: I'm not sure what you mean

src/action_management/include/actionmanagement.h
MTUC08: resolved lower down

src/action_management/src/actionmanagement.c
yoonsjeong: we don't yet fully know what will have to be altered when you talk to an npc, but I say it's safe and also consistent with the other functions just to leave the game state in.

also, if you only have the npc as an input you won't be able to check for validity to see if that npc is actually in the room with you -- the game struct holds all the rooms you are in

src/action_management/src/actionmanagement.c
yoonsjeong: this one is specifically the enum so that you could call the function
> get_supported_actions(3)
if say, I wanted all the actions of kind 3

src/action_management/src/actionmanagement.c
yoonsjeong: it seems actually like I added this in error, i'll fix it in the next pull request

yoonsjeong:
I've fixed some of the comments made here and will be making a follow-up pull request shortly with some appropriate changes and additions.

mpcsh:
@yoonsjeong any particular reason you're opening a new PR rather than just making the updates here?

Issue #56, opened 2019-05-07
Improving the validation from WDL proof-of-concept code


mewthew:
With a more detailed WDL specification file, we can now validate more specific attributes within each object. For example, we can verify that a room has an id, short description, long description, and connections. For this task, we will seek to expand what we have written already in #24 to comply with these new specifications.

mewthew:
Tentative To Dos:
1. Update the implementation of validation functions for game, rooms, items. Also add implementation for player.
2. We are making an extract object function to get the list of objects for rooms, items, and players.
3. Use the new validation functions with extract object to automate the main

mewthew:
Update:
We decided to split up the parse module that we created for validation into two modules: validate and parse. Now, the validate module will contain verify and print functions while the parse module will contain extracting objects, mapping objects to specific game_state structs, and verifying attributes.

victor-a-a:
Update:
We decided on a method for validating and returning objects to be placed in the Game State structs.  In essence, validation is divided into two stages.  After we load an object, we first type check it to verify all fields are present and of the right type.  After type checking, we check the content of the object for any restrictions.  In this case, we have been working towards checking that referenced object ids (e.g when a room has the id for a connected room) are valid, and checking that no two objects of the same type (e.g. room, item, player) have the same id.

Currently, we have written a function to return lists of room, item, and player objects, `get_obj_list()`.  We also wrote a function to apply type checking tests to a list of objects, `list_type_check()`.  We are in the process of updating our `verify_<obj>()` functions to `<obj>_type_check()` functions.

The objective is to integrate all of this into a catch all function `extract_objects()` which will return a validated list of objects of a certain type given the document object.

victor-a-a:
In addition to writing `get_obj_list()` and `list_type_check()`, we also wrote the function `list_print()` which applies a print function to a list of corresponding objects.  We have successfully updated our `<obj>_type_check()` and `print_<obj>()` functions.  We wrote a `main.c` file to test our functions.

lydiafilipe:
Issue Score: ✓+

Comments:
Nice work! Keep it up with the progress updates especially. Small thing - in your conclusion the module link is broken, the file may have been moved or changed

PR #78, opened 2019-05-14
merge wdl/improved validation into wdl/dev

mewthew:
We have updated the validation functions to further comply with the new WDL specification and created a more automated process for validating and verifying objects, all found in issue #56. 

Specifically, we made the following changes:
* implemented `get_obj_list()` and `list_type_check()` functions, which respectively obtains a list of a given object and verifies the types of its attributes
* implemented `list_print()`, which applies a print function to a list of corresponding objects
* we have successfully updated our `obj_type_check()` and `print_obj()` functions
* We wrote a main.c file to test our functions

The module for our updates can be found [here](https://github.com/uchicago-cs/chiventure/tree/wdl/improved_validation/src/libwdl).


src/libwdl/include/validate.h
```diff
@@ -0,0 +1,120 @@
...
+
+#ifndef INCLUDE_VALIDATE_H
+#define INCLUDE_VALIDATE_H
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include "parser.h"
+
+/* print_item
```
bansals10: General question about the usage of printing: Are the print functions used for visual testing? Will there be automated validation for the contents of an object being read in besides through the type checking functions or will this have to be checked manually by the developer?

src/libwdl/src/validate.c
```diff
@@ -0,0 +1,242 @@
...
+        id = false;
+    if (obj_get_type(obj, "direction") != TYPE_STR)
+        direction = false;
+    if (obj_get_type(obj, "through") != TYPE_STR)
+        through = false;
+
+    return (id && direction && through);
+}
+
+/* connection_type_check()
```
bansals10: Are you only checking to make sure that the connections are associated with rooms here? It may be helpful to communicate with UI about where this validation is happening in relation to when a room may be rejected for having an invalid connection.

src/libwdl/include/validate.h
mewthew: Yes, print functions are used for visual testing. We will also have restriction testing, which will make sure that the required attributes of some objects (e.g. a room has a description) exist.

src/libwdl/src/validate.c
mewthew: As of now, we are only checking whether the types of rooms are inputted correctly and that they exist (So, yes kinda to answer your question). I think the game_state structs (like their path module) should coincide with whatever UI is doing, so I think we should be OK, although more clarification would be good with UI and gamestate.

lydiafilipe:
Issue Score: ✓

Comments:
Nice work! However this didn't include a closing statement. Also, the title isn't very informative. (Sorry for grading this late, I know this is from an earlier sprint and you all have since incorporated feedback like this :) )

PR #79, opened 2019-05-14
Add starting structure/code for CLI component

Tianlel:
Our component includes five modules: cmd, operations, shell, parser, validation.

- The cmd module introduces a data structure for commands

- The operations module holds miscellaneous action functions that call functions provided by the game state team and the action management team. 

- The shell module holds generic shell operations (e.g. help, quit)

- The parse module parses an input string

- The validation module checks the validity of a command input by checking whether an action is supported and whether an item is accessible. 
   Note that we don't check for comand logic. ("drink potato" is considered a valid input if drink is a supported action and potato is accessible to the player)

The functions we provide are meant to be called in a REPL loop. We implemented a sample REPL loop in cli/example/testshell.c (this loop uses the GNU readline library). To run the example interface, do `make test` and `./example/testshell` in the cli/ directory.


src/cli/example/testshell.c
```diff
@@ -0,0 +1,71 @@
...
+#include "../include/shell.h"
+
+#define BUFFER_SIZE 256
+
+/* trim_newline: duplicates the string and replace the newline
+ * at the end with '\0' if it exists.
+ */
+char *trim_newline(char *s)
+{
+    int n = strlen(s);
```
lydiafilipe: It is worth noting that there are "safer" string functions in c like strnlen, strncmp, strndup etc that take a parameter n where you supply the number of bytes, and this helps prevent against buffer overflow. You don't necessarily have to change these now for this pr but it's worth keeping that in mind and trying to use those.

src/cli/include/operations.h
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * Returns:
+ *  - Some system message
+ *
+ * Note that this command literally does nothing right now.
+ */
+char *quit_operation(char *tokens[TOKEN_LIST_SIZE]);
+
+
+/*
```
lydiafilipe: A lot of these functions are not properly documented - e.g. missing parameters, return values etc - please follow the documentation guidelines in the style guide

src/cli/include/shell.h
```diff
@@ -0,0 +1,46 @@
...
+ * FROM THE CMSC 15200 (WIN 2019) COURSE TOUGHT BY ADAM SHAW 
+ * (University of Chicago)
+ */
+
+/* ========================= */
+/* === interface actions === */
+/* ========================= */
+
+/* 
+ * Prints out a help menu that contains a list of supported commands
```
lydiafilipe: This file also has some functions that are not properly documented, please follow the documentation guidelines

src/game-state/include/object.h
```diff
@@ -0,0 +1,98 @@
...
+    union object_union t;
+} tagged_objs;
+
+/* this object struct will include a door object to be used between rooms,
+i.e. included in the room struct in its list of exits */
+typedef struct object {
+    /* hh is used for hashtable, making this struct hashable, as provided in uthash.h*/
+    UT_hash_handle hh;
+    char *object_id;
+    char *short_desc;
```
yoonsjeong: Our initial thinking about the object struct we requested from CLI was that it would be any generic struct for any word that could be action'd upon (open DOOR, go EAST, pull SCONCE), but I think this may have to change moving forward which we realized after communicating with Game State.

The primary things we need from CLI is to convert a string, say "go east" into something that can be put into our functions. At the moment, game state has made their own solutions to some of this, so I am thinking that CLI would have to convert from "ball" into item_t ball, a struct of Game State's, before giving it to Action Management in one of our functions. The same goes for a direction like "east", which would be converted into Game State's path_t struct, and then passed to us, Action Management.

This is something we could discuss in further detail since this is just a rough idea of what we thought might need to happen.

yoonsjeong: oh, I just realized I was looking at gamestate code. I think we should still talk about having a system of converting input strings into game state's structs though!

src/cli/src/validate.c
```diff
@@ -0,0 +1,61 @@
...
+
+/* Forward declaration. Need to collaborate with game state */
+game_t *curr_game;
+bool is_in_room(char *object);
+
+/* See validate.h */
+cmd *assign_action(char **ts)
+{
+    cmd *output = cmd_new(ts);
+    if(strcmp(ts[0],"QUIT")==0) output->functionofcommand = quit_operation;
```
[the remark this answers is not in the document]
Tianlel: Yeah we actually planned to implement the hashtable in this sprint #103

src/cli/include/operations.h
```diff
@@ -0,0 +1,87 @@
...
+ * The functions below take in a validated parsed command, package the 
+ * command into structs that can be passed onto the corresponding action 
+ * functions provided by the action management team.
+ *
+ * Will be further clarified. 
+ *
+ */
+char *type1_action_operation(char *tokens[TOKEN_LIST_SIZE]);
```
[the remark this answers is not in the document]
Tianlel: This function converts string input into action structs and object structs that can be passed to the action functions your team provides, so there's some work involved in the linking process which is integral to the command struct. Our struct also supports system actions that are not handled by the action management team (such as quitting and saving..), so I think it is still necessary. But I agree there is some redundancy in the way we are handling the operation functions and we should figure out how to connect these systems more efficiently.

src/cli/src/validate.c
```diff
@@ -0,0 +1,61 @@
...
+    else if(strcmp(ts[0],"HELP")==0) output->functionofcommand = help_operation;
+    else if(strcmp(ts[0],"HIST")==0) output->functionofcommand = hist_operation;
+    else if(strcmp(ts[0],"OPEN")==0) output->functionofcommand = type1_action_operation;
```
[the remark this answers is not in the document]
Tianlel: Yeah this is a good idea

src/game-state/include/object.h
Tianlel: Agree

src/cli/include/operations.h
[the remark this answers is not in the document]
Gestalt-R: By using a single data type for all kinds of commands, we can have a more modular interface. Sacrificing this for specificity means creating a different set of tools to handle each kind of input. This is a bad thing that makes the code harder to understand, edit, and work with.

Tianlel:
Closing statement:
After all the style fixes, the basic structure is good to go. The next major step will be implementing the hash table structure to store different actions.

lydiafilipe:
Issue Score: ✓+

Comments:
Good work!

PR #100, opened 2019-05-15
Merging ui into master

sebvil:
This branch contains the code necessary to run the UI. Right now, the UI is split in two windows (top and bottom), one which contains a dummy CLI, the other one which contains "info" (though mostly empty now, but could be used for images, score, map, inventory, etc.). To UI can be run from a main function, with a call to the function start_ui() (from ui.c). The info windows can be switched by pressing Alt+m, and the position of the windows can be switched by pressing Alt+s. To exit, press CTRL+D. The test_ui.c in examples shows how the Ui works.


src/ui/src/general_ui.c
```diff
@@ -0,0 +1,124 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <ncurses.h>
+#include <ctype.h>
+#include <stdbool.h>
+#include "general_ui.h"
+#include <signal.h>
+
+int score = 0;
```
mpcsh: As I mentioned in another comment, global variables need an extremely strong justification: basically, the only way I will consider merging code that uses global variables is if it's absolutely mandated by `ncurses`. What is the purpose of each of these?

src/ui/src/general_ui.c
```diff
@@ -0,0 +1,124 @@
...
+int moves = 0;
+int char_count = 0;
+int line = 1;
+int ch;
+
+/* Signal Handler for SIGINT */
+void sigintHandler(int sig_num)
```
mpcsh: So actually, besides needing to be in `snake_case`, this function definitely isn't necessary. Per the link below, it's sufficient to just have `signal(SIGINT, SIG_IGN)` once at the top of `start_ui()`.

src/ui/include/general_ui.h
```diff
@@ -0,0 +1,61 @@
+#ifndef GENERAL_UI_H
+#define GENERAL_UI_H
+
+typedef struct window {
+    WINDOW *w;
+    void (*print)(struct window *win);
```
mpcsh: Consider creating a `typedef` that allows you to conveniently refer to this function signature. As it stands, it gets repeated many times in arguments to other functions.

src/ui/src/general_ui.c
```diff
@@ -0,0 +1,124 @@
...
+    signal(SIGINT, sigintHandler);
+//    fflush(stdout);
+}
+
+/* see general_ui.h
+ */
+window_t *create_newwin(int height, int width, int starty,
+                        int startx, int show, void (*print)(window_t *wins)){
+    window_t *local_win;
+    WINDOW *w;
```
mpcsh: I'm not a fan of this style of declaring all variables at the top of functions, because then down below when you see `w = newwin(...)`, you have to backtrack up to here to find out what the type of `w` is. We should instead write `WINDOW *w = newwin(...)` and do away with these top declarations.

src/ui/src/general_ui.c
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/* see general_ui.h */
+void print_info(window_t *win){
+    mvwprintw(win->w, 1, 2, "Score: %i", score);
+    mvwprintw(win->w, 2, 2, "Moves: %i", moves);
+}
+
+/* see general_ui.h */
+void print_cli(window_t *win){
```
mpcsh: Will it be necessary to pass keystrokes to the CLI team like this? It should not be our responsibility to implement the behavior of the backspace key, etc; we should get that for free. Does ncurses not provide a way to just hand over control of a window?

src/ui/src/general_ui.c
```diff
@@ -0,0 +1,124 @@
...
+            break;
+        case KEY_RIGHT:
+            getyx(window, y, x);
+            // checks that cursor is not past the charcters inserted
+            if(x < char_count + 3){
+                wmove(window, y, x + 1);
+            }
+            break;
+        case KEY_DC:
+            // implements delete key
```
mpcsh: The comments in this switch statement are inconsistent. Either delete all such "implements X key" comments, or include them in all cases, and decide (and be consistent on) whether they go above or below the `case X:` lines.

src/ui/src/ui.c
```diff
@@ -0,0 +1,116 @@
...
+    height = LINES / 2;
+    width = COLS;
+
+
+
+
+    /* initializes the widows. there is one for score, one where maps could
+     * be displayed, and the cli window
+     */
+    score = create_newwin(height, width, cli_top * height, 0, true, print_info);
```
mpcsh: Once you create a `window_free` as requested in an earlier comment, you'll need to use it to free these windows.

src/ui/src/ui.c
```diff
@@ -0,0 +1,116 @@
...
+
+    // reads input from the cli window, allows scrolling
+    keypad(cli->w, TRUE);
+    scrollok(cli->w, TRUE);
+
+    // prints '>' in the cli window
+    mvwprintw(cli->w, line, 2, ">");
+
+    // prints the score and number of moves in the info window
+    info->print(info);
```
mpcsh: This pattern is a little awkward since we have to write `info` twice. Could we, perhaps, just write one generic function called `window_print` that takes a `window_t *` and calls the print function inside that `window_t *`? That way, we would just write `window_print(info)` here.

src/ui/src/ui.c
```diff
@@ -0,0 +1,116 @@
...
+        info->print(info);
+
+        // refreshes windows to reflect changes
+        wrefresh(info->w);
+        wrefresh(cli->w);
+
+
+    }
+
+    endwin();			/* End curses mode		  */
```
mpcsh: Comments belong on the line above the code they comment on. Also, for single line comments, please use `//` instead of `/* ... */`

mpcsh:
Also, two notes I forgot to include in the review:
1. Where's the `README`?
2. `general-ui.{h,c}` is a pretty misleading name. They should probably just become `window.{h,c}`. I'm definitely open to other ideas on that name but it definitely needs to change.

mpcsh:
Oh lastly @sebvil were you the only one that worked on this? I thought @Sshatzkin contributed to this too. If so, the assignees should be updated accordingly.

src/ui/Makefile
```diff
@@ -0,0 +1,44 @@
+# Makefile based on template at https://gist.github.com/xuhdev/1873316
+
+CC = gcc
+AR = ar
+CFLAGS = -fPIC -Wall -Wextra -O2 -g -I./include/ -I../common/include/
+RM = rm -f
+LIB = libui.a
+LDLIBS = -lyaml -lncurses
```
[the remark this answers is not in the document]
sebvil: I can remove it. I just left it there from the libobj Makefile

src/ui/src/general_ui.c
sebvil: I think we should be able to do that, but not exactly sure yet. I will be working with the CLI team this sprint to integrate their CLI with our UI

src/ui/examples/test_ui.c
```diff
@@ -0,0 +1,17 @@
...
+#include <stdio.h>
+#include <ncurses.h>
+#include <ctype.h>
+#include <stdbool.h>
+#include <signal.h>
+#include "general_ui.h"
+#include "ui.h"
+
+
+extern int line;
```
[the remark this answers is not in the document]
sebvil: `line` is actually not necessary, and `ch` is only necessary with the way the `print_cli()` function is written right now. Since we might not need the `print_cli` function as you mention in a comment below, I could get rid of it, and have the `print_cli` function just print ">" but not handle text input.

sebvil:
> Oh lastly @sebvil were you the only one that worked on this? I thought @Sshatzkin contributed to this too. If so, the assignees should be updated accordingly.

@Sshatzkin  worked on the maps, I worked on the UI.

sebvil:
@mpcsh  I made all the changes you suggested (except for the README, which I will add tomorrow). For the global variables, since they were mostly used for the printing functions (updating the score/moves, and the printing the characters to the dummy CLI), and this functionality is not necessary, I just removed them and simplified the print functions to only print one message.

src/ui/Makefile
```diff
@@ -4,7 +4,7 @@ CC = gcc
 AR = ar
 CFLAGS = -fPIC -Wall -Wextra -O2 -g -I./include/ -I../common/include/
 RM = rm -f
-LIB = libui.a
+LIB = ui.a
```
mpcsh: I believe this change was incorrect - shouldn't it be `libui.a`?

src/ui/examples/test_ui.c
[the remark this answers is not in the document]
mpcsh: It looks like you removed the global variables in `ui.c` but not here. Can they also be removed here?

src/ui/include/print_functions.h
```diff
@@ -0,0 +1,44 @@
...
+ * Prints the word map in the window (top left corner)
+ *
+ * Parameters:
+ *   - win : window where to print the info
+ *
+ * No value is returned
+ */
+void print_map(window_t *win);
+
+/* print_window
```
mpcsh: Oops - didn't catch this. This function should:
1. Live in `window.h`, not this file
2. Be called `window_print`

src/ui/Makefile
sebvil: I was looking at the Developer Guidelines and it said it should match the name of the component. I can change it back

sebvil:
> Looking much better. Regarding style: it looks like `window.c` and `print_functions.c` are formatted with Allman style, but everything else looks like Stroustrop. Did you run `astyle` on everything?

I did run it, and just run it again and it did not make any changes.

src/ui/include/print_functions.h
sebvil: Changed

src/ui/Makefile
mpcsh: Oh huh, never mind, you're right! I hadn't seen that part of the guidelines.

mpcsh:
🎉 this is good to go from me, we just need someone not from UI to review as well. Great work @sebvil!

mpcsh:
@sebvil can this branch be deleted or does it need to stick around?

sebvil:
> @sebvil can this branch be deleted or does it need to stick around?

Can be deleted. Will do so now.

mpcsh:
Issue Score: ✓

Comments:
Good work - needed a more descriptive title.

Issue #105, opened 2019-05-15
Validating Commands and Objects exists


ecoble:
Validate that a command exists in the list of commands laid out by action management, then validate that the object exists in a room using game state's functions

//Please add more description as necessary

efraimdahl:
UPDATE: Added a few instances of object validification in operations.c, They are using a placeholder at the moment.

THINGS TO WORK ON: Actually using game-states function instead of a placeholder;
Changing the error messages to be more detailed, containing the item that was not found etc.

Tianlel:
The validation part currently in the operation functions should be refactored into the validate module

Tianlel:
UPDATE: all game-state related placeholders have been replaced by game-state macros. We are waiting on #106 to call action-management functions in our operation functions

efraimdahl:
CLOSING NOTICE
Action Validation now happens within the hash-table lookup function
Object Validation (for appropriate actions) finds place within the operation functions, by iterating through the macros provided by game state. 

We will want to update the object validation to utilize the get_item_in_room function that will be part of game-states next PR

Other than this minor issue the goals of this Issue within this sprint are achieved.

lydiafilipe:
Issue Score: ✓+

Comments:
Good work!

PR #130, opened 2019-05-18
Hotfix: game-state build on master is broken

borjasotomayor:
game-state's build on `master` (from PR #105) is broken and needs to be fixed as soon as possible:

```
chiventure/src/game-state$ make
gcc -fPIC -Wall -Wextra -O2 -g -I./include/ -I../common/include/ -I./src/   -c -o src/game.o src/game.c
src/game.c:2:25: fatal error: common-room.h: No such file or directory
compilation terminated.
<builtin>: recipe for target 'src/game.o' failed
make: *** [src/game.o] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ all: $(BIN)
 #  - Directory: src/ui/
 #  - Makefile: src/ui/Makefile
 #  - Library: src/ui/ui.a
-COMPONENTS = libobj ui
+COMPONENTS = libobj ui cli game-state
 LIBS = $(foreach comp,$(COMPONENTS),src/$(comp)/$(comp).a)
 
 $(LIBS):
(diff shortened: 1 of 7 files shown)


a-matai:
Problem solved.

There were three key issues:

- path_id was renamed to direction in the path struct, but we didn't change all references `path->path_id` to `path->direction`. Renamed everywhere for consistency.

- The Makefile produced a game-state.so file instead of a game-state.a file which prevented the final build. This was the key error in the final build compilation.

- game.c included common-player and common-room headers, which unfortunately were not added in the final commit. These have been commented out for now but will be resolved once the hotfix is resolved.

borjasotomayor:
This looks good, I'm waiving the reviews to get this hotfix into master.

PR #147, opened 2019-05-20
Merging checkpointing/proto to checkpointing/dev

jonahdf:
Jarvis and I updated proto to reflect current game state (still haven't included action management) and deleted dummy struct


jarvislam1999:
A note that the proto file may change in the future depending on game state answers so we will still be working on the proto branch. However, we should get this merged to dev quick so we can begin working on the save and load functions.

src/checkpointing/game.proto
```diff
@@ -1,131 +1,67 @@
...
+    optional string long_desc = 3;
+    // optional bool can_take // reserved for future expansion
+    repeated Attribute attributes = 4; //Array of attributes (submessage)
+    optional int32 attributes_len = 5; //Length of array (helpful for loading)
 }
 
-//Room
+message Condition {
+    optional string item_id = 1; //Assuming item is already stored in a room or player, only ID is needed
+    optional Attribute attribute = 2; //May change when Game State explains how attributes work in paths
```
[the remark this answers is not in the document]
ellynliu: are we ignoring the linked list inside condition_t? Here's the full struct:

```
typedef struct condition {
    /* fields used for linked list of condition_t */
    struct condition *next;

    item_t *item;
    char *attribute;
    attribute_value_t value;
} condition_t;
```

src/checkpointing/game.proto
```diff
@@ -1,131 +1,67 @@
...
-message Room {
-	required string room_id = 1;
-	optional string short_desc = 2;
-	optional string long_desc = 3;
-	required int32 objs_len = 4; 
-	repeated Object objs = 5; //Array of all object ids in room
-	//repeated Path paths = 5; Array of all paths (submessage) ~future expansion~
+message Path {
+    optional string direction = 1; 
+    optional string destination = 2; //Room ID of destination
```
ellynliu: is there a particular reason why we call this destination while game state calls it dest? We should probably try to stay consistent with naming

src/checkpointing/game.proto
```diff
@@ -1,131 +1,67 @@
 syntax = "proto2";
 
-// Dummy Implementation
+//In Game-state implementation, attribute is a union. Proto doesn't support unions, so we use optional instead
+message Attribute_value {
+    optional double double_val = 1;
+    optional string char_val = 2; //NO CHAR TYPE SUPPORTED IN .PROTO, SO USING STRING INSTEAD
+    optional bool bool_val = 3;
+    optional string string_val = 4;
```
ellynliu: game state calls this str_val, but here you call it string_val

chrisfernandes18:
I agree with what Ellyn has been saying with renaming the proto variable names to match the ones from game struct so it'll be easier to know what that part of the struct reflects from the game-state header files.

src/checkpointing/game.proto
```diff
@@ -1,131 +1,67 @@
...
-//Game
 
 message Game {
-	repeated Player players = 1; //Array of all players (submessage)
-	required int32 players_len = 2;
-	repeated Room rooms = 3; //Array of all rooms (submessage)
-	required int32 rooms_len = 4;
-	optional string curr_room = 5; //ID of current room
-	optional int32 start_time = 6;
+    repeated Player players = 1; //Array of all players (submessage)
```
chrisfernandes18: Game state named this all_players

src/checkpointing/game.proto
```diff
@@ -1,131 +1,67 @@
...
+    required int32 players_len = 2;
+    repeated Room rooms = 3; //Array of all rooms (submessage)
```
chrisfernandes18: Game state named this all_rooms

src/checkpointing/game.proto
```diff
@@ -1,131 +1,67 @@
...
+    optional bool bool_val = 3;
+    optional string string_val = 4;
+    optional int32 int_val = 5;
+}
 
-// Object
+message Attribute {
+    required string attribute_key = 1; // Presumably ID for attribute hashtable
+    required string attribute_tag = 2; //Tag of which type attribute value will be
+    required Attribute_value val = 3; //Value of attribute (submessage)
```
chrisfernandes18: game state names this attribute_value

src/checkpointing/game.proto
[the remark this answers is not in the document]
jarvislam1999: Hi I talked to Game State and I think we reached a verdict where they will change the attribute_value_t struct in condition to attribute_t, which solved our other problems. I made some changes under Condition which reflects our agreement but there may be minor changes in the future so keep an eye out for those.

We don't need to include the linked list inside condition_t because we will be making those into an array. We will have to iterate through it in our load/save functions, however, which would be my task so no worries. 

I will resolve this comment for now.

src/checkpointing/game.proto
jarvislam1999: No reason. Updated the file to your request.

src/checkpointing/game.proto
jarvislam1999: Good catch! Fixed.

src/checkpointing/game.proto
jarvislam1999: Nice catch, issue fixed.

src/checkpointing/game.proto
jarvislam1999: Changed as requested.

src/checkpointing/game.proto
jarvislam1999: Fixed this also.

PR #149, opened 2019-05-20
updated save functions

ellynliu:
I wrote new functions for save_attribute_value, save_attribute, save_condition, and save_path in save.h and save.c. I also updated the old functions to match the new proto file. However, I ignored anything that uses hash tables by leaving a bunch of comments where they would belong.

### DIFF
--- a/src/checkpointing/src/save.c
+++ b/src/checkpointing/src/save.c
@@ -4,33 +4,156 @@
 #include "game.pb-c.h"
 #include "save.h"
 
-int save_object(object_t *o_t, Object *o)
+int save_attribute_value(attribute_value_t *av_t, Attribute_value *av)
 {
-    if (o_t == NULL){
-        return -1;
-    }
-  
-    o->object_id = o_t->object_id;
-  
-    if (o_t->short_desc == NULL) {
-	o->short_desc = NULL;
-    } else {
-	o->short_desc = o_t->short_desc;
-    }
-  
-    if (o_t->long_desc == NULL) {
-	o->long_desc = NULL;
-    } else {
-	o->long_desc = o_t->long_desc;
+    if (av_t == NULL) {
+	fprintf(stderr, "Given a NULL attribute_value struct in save_attribute_value");
+	return -1;
     }
 
+    if (av_t->double_val == NULL) {
+	av->double_val = NULL;
+    } else {
+	av->double_val = av_t->double_val;
+    }
+
+    if (av_t->char_val == NULL) {
+	av->char_val = NULL;
+    } else {
+	av->char_val = av_t->char_val;
+    }
+
+    if (av_t->bool_val == NULL) {
+	av->bool_val = NULL;
+    } else {
+	av->bool_val = av_t->bool_val;
+    }
+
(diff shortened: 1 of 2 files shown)


src/checkpointing/src/save.c
```diff
@@ -4,33 +4,156 @@
...
+	return -1;
+    }
+
+    if (c_t->item->item_id == NULL) {
+	c->item_id = NULL;
+    } else {
+	c->item_id = c_t->item->item_id;
+    }
+
+    // THIS MIGHT BE attribute_key
```
jarvislam1999: Yes, I think attribute_key will no longer be there but you should change it to expected_attribute instead of just attribute (I will try to push game state to implement this change, chances are that we may have to use dummy placeholders, however)

src/checkpointing/src/save.c
```diff
@@ -4,33 +4,156 @@
...
+    }
+
+    if (p_t->direction == NULL) {
+	p->direction = NULL;
     } else {
-	o->long_desc = o_t->long_desc;
+	p->direction = p_t->direction;
     }
 
+    if (p_t->dest == NULL) {
```
jarvislam1999: Change destination to just dest, sorry our proto file is causing you so much trouble

src/checkpointing/src/save.c
```diff
@@ -47,64 +170,61 @@ int save_room(room_t *r_t, Room *r)
...
+    for (int i = 0; i < len; ++i) {
+	items[i] = malloc(sizeof(Item));
+	item__init(items[i]);
+	int transfer_item_success = save_item(r_t->items[i], items[i]);
     }
 
-    r->objs = objs;
-  
+    r->items = items;
+    */
```
jarvislam1999: Add a similar comment section as item for path? Sorry, I'm just lazy, but it would help me remember to implement path. Just dismiss this if there's no time

src/checkpointing/src/save.c
```diff
@@ -118,37 +238,21 @@ int save_player(player_t *p_t, Player *p)
...
 
 int save_game(game_t *g_t, Game *g)
 {
     if(g_t == NULL){
-	fprintf(stderr, "Given a game_t struct that is NULL in transfer_game.\n");
+	fprintf(stderr, "Given a game_t struct that is NULL in save_game.\n");
 	return -1;
     }
-  
+
```
jarvislam1999: Change players to all_players, rooms to all_rooms to reflect the new proto structs

src/checkpointing/src/save.c
```diff
@@ -194,20 +300,19 @@ int save_game(game_t *g_t, Game *g)
     }
 
     g->rooms = rooms;
-  
+    */
     if (g_t->curr_room == NULL) {
 	g->curr_room = NULL;
     } else {
 	g->curr_room = g_t->curr_room;
     }
```
jarvislam1999: I think you can delete start_time and everything similar like has_start_time since it's no longer in Master or proto file

ellynliu: I already did?

PR #155, opened 2019-05-21
Added .gitignore for .pb-c

jonahdf:
Created a .gitignore to ignore the protobuf generated files ending with .pb-c


jonahdf:
I made new commits to dev without realizing they would also apply to this PR. Gonna close this for now.

PR #168, opened 2019-05-25
Checkpointing/loading

ellynliu:
I updated and wrote new functions in load.h and load.c to reflect save.h and save.c. I didn't write anything that involved hash tables. Let me know what changes I need to make, especially regarding writing into a game struct instead of mallocing a new game struct. Right now, the functions just return an int for success or failure and use protobuf to unpack. But I'm not writing directly into a game struct.

### DIFF
--- a/src/checkpointing/src/load.c
+++ b/src/checkpointing/src/load.c
@@ -23,12 +23,74 @@ size_t read_file(char *filename, unsigned max_length, uint8_t *out)
 }
 
 
-int load_attribute_value(Attribute_value *av, attribute_value_t *av_t);
+int load_attribute_value(Attribute_value *av, attribute_value_t *av_t)
+{
+    if (av_t == NULL) {
+	fprintf(stderr, "given null attribute_value_t struct\n");
+	return -1;
+    }
+
+    if (av->double_value != NULL) {
+	av_t->double_value = av->double_value;
+    } else {
+	av_t->double_value = NULL;
+    }
+
+    if (av->char_val != NULL) {
+	av_t->char_val = av->char_val;
+    } else {
+	av_t->char_val = NULL;
+    }
+
+    if (av->bool_val != NULL) {
+	av_t->bool_val = av->bool_val;
+    } else {
+	av_t->bool_val = NULL;
+    }
+
+    if (av->str_val != NULL) {
+	av_t->str_val = av->str_val;
+    } else {
+	av_t->str_val = NULL;
+    }
+
+    if (av->int_val != NULL) {
+	av_t->int_val = av->int_val;
+    } else {
+	av_t->int_val = NULL;
+    }
+
+    return 0;
+}
 
 
-int load_attribute(Attribute *a, attribute_t *a_t);
+int load_attribute(Attribute *a, attribute_t *a_t)
+{
+    if (a_t == NULL) {
+	fprintf(stderr, "given null attribute_t struct\n");
+	return -1;
+    }
 
+    a_t->attribute_key = a->attribute_key;
 
+    if (strcmp(a->attribute_tag, "DOUBLE") == 0) {
+	a_t->attribute_tag = DOUBLE;
+    } else if (strcmp(a->attribute_tag, "BOOLE") == 0) {
+	a_t->attribute_tag = BOOLE;
+    } else if (strcmp(a->attribute_tag, "CHARACTER") == 0) {
+	a_t->attribute_tag = CHARACTER;
+    } else if (strcmp(a->attribute_tag, "STRING") == 0) {
+	a_t->attribute_tag = STRING;
+    } else {
+	a_t->attribute_tag = INTEGER;
+    }
+
+    load_attribute_value(a->attribute_value, a_t->attribute_value);
+
+    return 0;
+}
+
+    
 int load_item(Item *i, item_t *i_t);
 {
     if(i_t == NULL) {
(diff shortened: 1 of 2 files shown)


src/checkpointing/src/load.c
```diff
@@ -164,17 +280,17 @@ int load_game(Game *g, game_t *g_t)
     g_t->rooms = rooms;
   
     g_t->rooms_len = g->rooms_len;  
-
+    */
     if(g->curr_room != NULL) {
```
jarvislam1999: so g->curr_room and g->curr_player stores the room id, not the pointer itself. Need to fix

src/checkpointing/src/load.c
```diff
@@ -22,87 +23,148 @@ size_t read_file(char *filename, unsigned max_length, uint8_t *out)
...
-	p_t->xp = p->xp;
+
+    // bool condition reserved for future expansion
+
+    // repeated Attribute HERE
+    
+    return 0;
+}
+
+
```
jarvislam1999: We can comment load_condition out, but should attempt to write it.

src/checkpointing/src/load.c
```diff
@@ -22,87 +23,150 @@ size_t read_file(char *filename, unsigned max_length, uint8_t *out)
...
+    if (av_t == NULL) {
+	fprintf(stderr, "given null attribute_value_t struct\n");
 	return -1;
     }
 
-    o_t->object_id = o->object_id;
-  
-    if (o->short_desc != NULL) {
-	o_t->short_desc = o->short_desc;
+    if (av->double_value != NULL) {
```
jonahdf: Shouldn't we use the tagged union instead of checking if each type is NULL? We can make one of the parameters the attribute tag. 

Correct me if I'm wrong, but won't we only ever call this function from the load_attribute function anyway?

src/checkpointing/src/load.c
```diff
@@ -22,87 +23,150 @@ size_t read_file(char *filename, unsigned max_length, uint8_t *out)
...
+    // bool condition reserved for future expansion
+
+    // repeated Attribute HERE
+    //ITER_ALL_ITEMS
+    return 0;
+}
+
+// won't need condition for semi-incremental saving
+/*
+int load_condition(Condition *c, condition_t *c_t)
```
jonahdf: We shouldn't have commented out code when we merge to master. This looks sloppy.

src/checkpointing/src/load.c
```diff
@@ -22,87 +23,150 @@ size_t read_file(char *filename, unsigned max_length, uint8_t *out)
...
-	inventory[i] = malloc(sizeof(object_t));
-	load_object(p->inventory[i], inventory[i]);
+    // something here? idk depends on game state
+
+    // optional Attribute here but might change
+
+    return 0;
+    }*/
+
+
```
jonahdf: Paths will be loaded by WDL and will never change. We do not need this.

src/checkpointing/src/load.c
```diff
@@ -124,27 +188,79 @@ int load_room(Room *r, room_t *r_t)
     } else {
 	r_t->short_desc = NULL;
     }
-  
-    object_t **objs = malloc(sizeof(object_t*) * r->objs_len);
+
+    // hash table stuff
```
jonahdf: Let's avoid this type of comment. Replace with "merging repeated item field with WDL," or just remove this section entirely. Since this task is for non-hash functions, we can just remove the other ones entirely.

src/checkpointing/src/load.c
```diff
@@ -124,27 +188,79 @@ int load_room(Room *r, room_t *r_t)
...
+	p_t->health = NULL;
+    }
+  
+    if (p->xp != NULL) {
+	p_t->xp = p->xp;
+    } else {
+	p_t->xp = NULL;
+    }
+
+    // hash table stuff for inventory
```
jonahdf: //Load player's inventory

No need to mention "hash table stuff"

src/checkpointing/src/load.c
```diff
@@ -164,17 +282,27 @@ int load_game(Game *g, game_t *g_t)
     g_t->rooms = rooms;
   
     g_t->rooms_len = g->rooms_len;  
+    */
 
-    if(g->curr_room != NULL) {
-	g_t->curr_room = g->curr_room;
+
+    if(g->curr_room != -1) {
```
jonahdf: This is because curr_room is our room_id, correct? We should comment this to clarify.

src/checkpointing/src/load.c
```diff
@@ -164,17 +282,27 @@ int load_game(Game *g, game_t *g_t)
...
+	ITER_ALL_ROOMS(g_t, curr_r) {
+	    if (curr_r->room_id == g->curr_room) {
+		g_t->curr_room = curr_r;
+	    }
+	}
     } else {
 	g_t->curr_room = NULL;
```
jonahdf: Is setting curr_room to NULL necessary? I feel like if we haven't saved a room, we should just start with whatever WDL loads in as the current room.

src/checkpointing/src/load.c
```diff
@@ -164,17 +282,27 @@ int load_game(Game *g, game_t *g_t)
     g_t->rooms = rooms;
   
     g_t->rooms_len = g->rooms_len;  
+    */
 
-    if(g->curr_room != NULL) {
-	g_t->curr_room = g->curr_room;
+
+    if(g->curr_room != -1) {
+	ITER_ALL_ROOMS(g_t, curr_r) {
```
jonahdf: Where did you define curr_r? I shouldn't be your compiler—you should probably see if it compiles before you request a code review.

src/checkpointing/src/load.c
```diff
@@ -164,17 +282,27 @@ int load_game(Game *g, game_t *g_t)
...
+	}
     } else {
 	g_t->curr_room = NULL;
     }
-    
-    if(g->has_start_time == 1) {
-	g_t->start_time = g->start_time;
+
+    if (g->curr_player != -1) {
+	ITER_ALL_PLAYERS(g_t, curr_p) {
```
jonahdf: Never defined curr_p

src/checkpointing/src/load.c
```diff
@@ -164,17 +282,27 @@ int load_game(Game *g, game_t *g_t)
...
+	    if (curr_p->player_id == g->curr_player) {
+		g_t->curr_player = curr_p;
+	    }
+	}
     } else {
-	g_t->start_time = -1;
+	g_t->curr_player = NULL;
```
jonahdf: Same comment as curr_room

src/checkpointing/src/load.c
ellynliu: In save.c, since the tagged union elements are optional, we set them to NULL if they were empty. So  while loading, I think we need to check if an element is null first before trying to access them. And correct, this function is only used in load_attribute

src/checkpointing/src/load.c
ellynliu: A lot of these comments mentioning hash table or repeated fields are for Jarvis since he'll be taking my code and writing the hash table functions. These comments are mainly to indicate where the hash table functions belong, hopefully making his job a little easier. As we did in save.c, he'll delete these comments once he's done

ellynliu:
Functions for load_path and load_condition have been deleted, as they are immutable. We will take what is already present in game state from WDL and just write over the structs that are mutable. Several functions had to be fixed to use iterators as provided by game state as I did not take into consideration overwriting a pointer in a struct. This code has comments in it for where repeated fields need to be filled in.

PR #173, opened 2019-05-27
Merging save into cli/dev

dbarer:
Just want someone to do a quick look to make sure the logic here makes sense. Need to have the updated save function from checkpointing to actually test that it works.

### DIFF
--- a/src/cli/include/operations.h
+++ b/src/cli/include/operations.h
@@ -64,7 +64,7 @@ char *hist_operation(char *tokens[TOKEN_LIST_SIZE], game_t * game);
  * - pointer to game to be saved,
  *
  * Returns:
- * - A NULL pointer
+ * - A success or error message
  */
 char *save_operation(char *tokens[TOKEN_LIST_SIZE], game_t *game);
 
--- a/src/cli/include/operations.h
+++ b/src/cli/include/operations.h
@@ -56,6 +56,19 @@ char *help_operation(char *tokens[TOKEN_LIST_SIZE], game_t * game);
 char *hist_operation(char *tokens[TOKEN_LIST_SIZE], game_t * game);
 
 /*
+ * Saves the current state of a game to a .dat file
+ * Prints an Error message if filename improperly specified
+ *
+ * Paramters:
+ * - tokens: parsed input string
+ * - pointer to game to be saved,
+ *
+ * Returns:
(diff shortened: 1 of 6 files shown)


src/cli/include/validate.h
```diff
@@ -77,6 +77,18 @@ bool validate_prep(cmd *c);
  */
 bool validate_ind_objects(cmd *c);
 
+/*
+ * Checks that a specified filetype is a .dat file
+ *
+ * Parameters:
+ * - A string of the filneame
```
Tianlel: typo

src/cli/include/operations.h
```diff
@@ -55,6 +55,19 @@ char *help_operation(char *tokens[TOKEN_LIST_SIZE], game_t * game);
...
+/*
+ * Saves the current state of a game to a .dat file
+ * Prints an Error message if filename improperly specified
+ *
+ * Paramters:
+ * - tokens: parsed input string
+ * - pointer to game to be saved,
+ *
+ * Returns:
+ * - A NULL pointer
```
Tianlel: should be error message or success message?

src/cli/src/operations.c
```diff
@@ -24,13 +24,15 @@ char *hist_operation(char *tokens[TOKEN_LIST_SIZE], game_t * game)
...
 char *save_operation(char *tokens[TOKEN_LIST_SIZE], game_t *game)
 {
   if(tokens[1] == NULL){
     printf("Save Error, No filename specified");
   }
-  int sv = save(game, tokens[1]);
+  if (validate(tokens[1]) == true){
+    int sv = save(game, tokens[1]);
+  }
   return NULL;
```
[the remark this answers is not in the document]
dbarer: Checkpointing handles all the error and success messages inside their save function. I just needed to add the error case where the filename is improperly formatted.

Tianlel:
Closing statement:
To successfully use the save function, we need to build under the main directory. However, right now checkpointing has dummy structures that conflict with the game state structs. So we commented out the newly implemented parts for now. These parts can be used and tested as soon as checkpointing updates their module.

lydiafilipe:
Issue Score: ✓

Comments:
Good work overall! However the title and summary are a bit vague, and you shouldn't be merging without all reviewers' explicit approval

Issue #131, opened 2019-05-18
Map Drawing: Refactor coordinate assigning code using Game State structs/functions


omorkved:
Refactor the code in coordinate.c and DFS.c to make use of the relevant functions provided by Game State and to no longer rely on dummy structs. (Especially accessing adjacent rooms within a room struct)

omorkved:
Current structs that need to be changed
coordinate_t in coordinate.h
room_t in coordinate.h
coord_t in map.h (same as coordinate_t but with three fields. duplicate because came from another feature branch)
room_t in map.h (duplicate)

To deal with the room_t structs, I will delete these and use the game state structs instead. There might be a few changes to make in handling rooms within functions depending on how game state has written the current struct.

To deal with the coordinate structs: I will keep the coordinate_t in coordinate.h, as I feel like it makes the most sense to have the struct defined in this module. I will also have to make a decision about whether to support two dimensions (x, y) or three (x, y, z). It looks like Sam already has a system to handle visualization of multiple floors (by using key handling to switch between floors). The coordinate struct acts as the key to the hash table, and uthash works by literally comparing all of the bytes in a struct to see if two structs are identical, so adding a third int field into the struct should not affect how uthash stores and recieves keys. Therefore, I am currently planning to add in a z-direction. I will ask Game State and WDL to support this third direction, but even if they don't, at least we will have the infrastructure in place in case this is supported in the future. (If a third-direction/z-axis is not implemented in this sprint, then I will just set the z value to be 0 in coord_init() in coordinate.c for every coordinate as a temporary measure).

This task also involves switching over to gamestate functions, particularly a function for finding whether a direction exists in a room, and if so, what is room pointer matches that direction. (This will be used extensively in DFS.c)

omorkved:
Ran into something unexpected -- game-state is using a char * for room IDs instead of integers. In DFS.c, we had originally been checking equality of room IDs to figure out whether an assigned coordinate was already assigned to the same room that we were checking (as opposed to attempting to double-assign a coordinate to a different room). However, I realized this could be solved by using strcmp() on the char * IDs so it's not a big deal. This should be fine as long as rooms are required to have _unique_ room IDs.

omorkved:
Update: This code was successfully merged with the rest of the code in Sebastian's cli-integration branch and with our context branch, and has just been merged into master. We have not run into any other issues integrating with Game State.

[...]

PR #174, opened 2019-05-27
Merge Ui/refactor structs to Ui/dev

omorkved:
See discussion in #131 for explanation of the changes I made. I also updated the makefile to include game-state headers--this code compiles, creates the ui.a library, and the coord_example executable runs correctly (with the text file generating containing all of the correct print statements). The only thing is in compiletime I get a warning that one of the game-state header files redefines SUCCESS/FAILURE, which are already definted in common.h. I did not want to change game-state's work, so I left as-is.


omorkved:
There were some conflicts with map.h in dev but I fixed them (conflicts arose because I had deleted the dummy coord and room structs in map.h on my feature branch, and Sam had updated the map struct in the meanwhile)

sebvil:
@omorkved this code does not compile. map.c is using the old room structs and is also missing the coordinate header, and therefore the build fails. Run make in both the UI directory and the root directory to make sure everything is compiling correctly.

omorkved:
> @omorkved this code does not compile. map.c is using the old room structs and is also missing the coordinate header, and therefore the build fails. Run make in both the UI directory and the root directory to make sure everything is compiling correctly.

Mmm I see, I only ran it in the UI directory--will do.

omorkved:
comments on map.c: I am also working on the map-from-coords branch with Sam, but I made clear comments so that when we have merge conflicts in Dev, it should be obvious which version to keep. The main problem was that the map.c code was using fields we had assumed room_t would have in our dummy struct. A lot of that info is now stored in the hashmap, which is what the map-from-coords branch is focused on integrating with. 

@sebvil

PR #175, opened 2019-05-27
Replaced all item related placeholders with game-state macro

Tianlel:
All changes are made in operation.c.


Tianlel:
Closing statement:
Since game state hasn't provided us with the get_item function, we are using the available macros to iterate through items. After merging the branches, we can build our static library successfully. However, we ran into issues while building the test. It only works if the game state header files use absolute paths instead of relative paths. We need to fix this in the next sprint.

lydiafilipe:
Issue Score: ✓

Comments:
Nice closing statement and title, however the summary is not very descriptive and there is not a ton of detailed commentary. Also you should not be merging without all reviewers' explicit approval

Issue #161, opened 2019-05-22
Create UI context struct


sebvil:
In this task, we want to create an UI context struct that will hold important information relevant to the UI that other teams might need to use in their code. Such information includes window structs that CLI might need to access to print info or modify the UI. 

Other work in this task will include modifying the main context struct (and related functions) so that they include our UI context struct, and modifying our existing UI functions so that they take the main context struct as a parameter.

omorkved:
Explanation of the coord_t field in ui_ctx:

We need to know the current _coordinate_ location of the player to know where to center the map. The current thinking is that, as the player moves through the map, we update the actual coordinate instead of reading the updated room *curr_room out of game state's curr_room field. (This is because our hash uses coords as keys to room *s, so finding a coordinate _from_ a room * would involve just searching from top to bottom of the hash map looking for that room *--seems inefficient, which is why we think it is better to just track the current coordinate location of the player).
In short, this would look like 1. writing a quick function that adds/subtracts one to either the x or y field (never both at the same time) (also this is more complicated if portals are supported) and 2. telling action management how to call this function with the appropriate fields whenever they allow the player to pass into a new room.

The alternative to this is to read the curr_room field out of game state everytime we redraw the map (which happens with any key press to change windows + every time the player moves) and then iterating through the hash to lookup which coord key corresponds with that specific room.
(Having coordinates as keys made sense for the purposes of DFS creating the map to efficiently check uniqueness of coord assignments)

omorkved:
Here are the current three fields in our ctx struct:

 UI context struct.  Contains the following fields (explanation of player_loc in my comment above):
```
/*
 *  - coord_hash: a pointer to the head of a hash map that matches
 *                coordinate keys to room pointers
 */
typedef struct ui_ctx
{
    map_t *map;
    coord_t *player_loc;
    coord_record_t *coord_hash;
```

However, previous commits will show that we had initially included pointers to windows in our actual ctx struct:

```
/*
 *  - map_win : contains the map
 *  - main_win : main window. May contain images or other info about the game
 *  - cli_win : window that contains the CLI. It's always displayed
 *  - displayed_win : window that is diplayed along with the CLI.
*/
```
We are still on the fence for this. @sebvil , could you post a summary of why we initially included windows and now are not including them?

sebvil:
[...]

PR #187, opened 2019-05-28
Ui/context structs

sebvil:
Merging ui context structs. All libraries successfully compile.


omorkved:
#161

src/ui/include/ui_ctx.h
```diff
@@ -0,0 +1,69 @@
...
+ *  - map_win : contains the map
+ *  - main_win : main window. May contain images or other info about the game
+ *  - cli_win : window that contains the CLI. It's always displayed
+ *  - displayed_win : window that is diplayed along with the CLI.
+ *  - coord_hash: a pointer to the head of a hash map that matches
+ *                coordinate keys to room pointers
+ */
+typedef struct ui_ctx
+{
+//    map_t *map;
```
Sshatzkin: ui_ctx needs to contain the map_t struct  in order for the current map implementation to work.  There are a lot of curcial map features stored in that struct.

src/ui/src/ui.c
```diff
@@ -5,16 +5,16 @@
 #include <stdbool.h>
 #include <signal.h>
 #include <string.h>
-#include "window.h"
+#include "ctx.h"
 #include "ui.h"
+#include "window.h"
 #include "print_functions.h"
-#include "map.h"
```
Sshatzkin: Why is map.h being removed from ui.c? Was this for compilation purposes? If so, I can amend this in my branch.

src/ui/src/ui_ctx.c
```diff
@@ -0,0 +1,100 @@
...
+#include <ctype.h>
+#include <stdbool.h>
+#include <signal.h>
+#include <assert.h>
+#include "ui_ctx.h"
+#include "game.h"
+#include "coordinate.h"
+#include "window.h"
+#include "print_functions.h"
+//#include "map.h"
```
Sshatzkin: This one too?

src/ui/src/ui_ctx.c
```diff
@@ -0,0 +1,100 @@
...
+    if (init == FAILURE) {
+        return NULL;
+    }
+
+    return ui_ctx;
+}
+
+// see ui_ctx.h
+int ui_ctx_init(ui_ctx_t *ui_ctx, game_t *game)
+{
```
Sshatzkin: Again, the map_t isn't initialized here. I think it's something that could be added here but  I can also add it later in my branch.

src/ui/src/ui.c
sebvil: Yes, the map stuff did not work correctly yet, so I removed/commented it out for now to make sure everything else compiled.

src/ui/src/ui_ctx.c
omorkved: it's at the bottom of the function, but commented out (for compilation)! Will let you uncomment in map-from-coords branch, since i know we are in the process of changing map_init()

PR #195, opened 2019-05-28
Checkpointing/loadtesting to Dev

chrisfernandes18:
All of our tests are done more or less. We need another pair of eyes to look at them. Then we need to have a discussion about whether we are going to load attribute and attribute_values.


jarvislam1999:
Sorry I did not spot this request, will get to it tonight

src/checkpointing/tests/test_load.c
```diff
@@ -0,0 +1,227 @@
+#include <criterion/criterion.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include "load.h"
+#include "game.h"
+
+Test(load, attribute_value)
```
jarvislam1999: I believe load_attribute_value() and load_attribute() is no longer needed in load and don't exist as functions. However, I noticed that save_attribute() is not tested in save, which may lead to some problems. Would you guys be able to switch?

src/checkpointing/tests/test_load.c
```diff
@@ -0,0 +1,227 @@
...
+    c_v->char_val = 'T';
+    Attribute_value *b_v;
+    a_v->bool_val = true;
+    Attribute_value *s_v;
+    s_v->str_val = "string";
+    Attribute_value *i_v;
+    i_v->int_val = 5;
+  
+    atttribute_value_t *d_t;
+    int succ_d = load_attribute_value(d_v, d_t);
```
jarvislam1999: assert_equal cannot be used on double so you may want to use the equivalent

src/checkpointing/tests/test_load.c
```diff
@@ -0,0 +1,227 @@
...
+    attribute_t *test_t;
+    Attribute *str;
+    Attribute_value *strv;
+    str->attribute_key = "string";
+    str->attribute_tag = "STRING";
+    attribute_value__init(strv);
+    strv->str_val = "this is a string";
+    str->attribute_value = strv;
+    
+    success = load_attribute(str, test_t);
```
jarvislam1999: assert not asser

src/checkpointing/tests/test_load.c
```diff
@@ -0,0 +1,227 @@
...
+    Attribute_value *dv;
+    d->attribute_key = "double";
+    d->attribute_tag = "DOUBLE";
+    attribute_value__init(dv);
+    dv->double_val = 2.5;
+    d->attribute_value = dv;
+    
+    success = load_attribute(d, test_t);
+    
+    cr_asser_eq(success, 0, "load_attribute failed");
```
jarvislam1999: assert not asser

src/checkpointing/tests/test_load.c
```diff
@@ -0,0 +1,227 @@
...
+                "load_attribute failed to load sttribute_value");
+}
+
+Test(load, item)
+{
+    Item *candy;
+    candy->item_id = "1234";
+    candy->short_desc = "salt water taffy";
+    candy->long_desc = "glues mouth shut";
+
```
jarvislam1999: attribute is missing in item

src/checkpointing/tests/test_load.c
```diff
@@ -0,0 +1,227 @@
...
+		 "load_item: loading long_desc failed");
+}
+
+Test(load, room)
+{
+    Room *dorm;
+    dorm->room_id = "5";
+    dorm->short_desc = "college dorm";
+    dorm->long_desc = "there are clothes and empty ramen cups everywhere";
+
```
jarvislam1999: Maybe add more than one item? And I don't think path is supported

src/checkpointing/tests/test_load.c
```diff
@@ -0,0 +1,227 @@
...
+     */
+}
+
+Test(load, player)
+{
+    Player *chad;
+    chad->health = 1;
+
+    player_t *player_t;
+
```
jarvislam1999: We need an inventory here

src/checkpointing/tests/test_load.c
```diff
@@ -0,0 +1,227 @@
...
+    cr_assert_eq(success, 0, "load_player failed");
+    cr_assert_eq(player_t->health, 1, "load_player: loading health failed");
+}
+
+Test(load, game)
+{
+    Game *game;
+    game->curr_room = "5";
+    game->curr_player = "chad";
+
```
jarvislam1999: We should add 1-2 other rooms here

src/checkpointing/tests/test_load.c
chrisfernandes18: Since we don't load attribute wouldn't it be loaded from the WDL file?

src/checkpointing/tests/test_load.c
schin99: attribute stuff deleted

src/checkpointing/tests/test_load.c
jarvislam1999: We still load attribute as though the number of attributes don't change, it's value might? I should ask action management and WDL

jarvislam1999: The load function that we wrote does take care of attribute, though there's no load_attribute() function. However, there's a save_attribute function so you may want to add test for that.

PR #200, opened 2019-05-28
Checkpointing/loading

jarvislam1999:
Have been editing the save.c and Makefile in this loading/ branch, whose version we want to use instead of in load. Should we delete the Makefile and save.c in load/ to avoid merge conflicts?


jarvislam1999:
This was a minor merge into a different branch so no reviews were needed.

Issue #198, opened 2019-05-28
Update Makefile to compile into static library


SnowballAntrobus:
Currently our Makefile compiles into a dynamic library. The project specification says that it must be a static library. The goal of this issue is to make our Makefile in line with the project spec.

SnowballAntrobus:
So the problem was that before we were generating a dynamic library and just sticking a .a extension on it like it was static. The new makefile will generate a .a file.

SnowballAntrobus:
The changes are currently in the branch actions/bugging which will now be merged with actions/test

SnowballAntrobus:
Decided to change this fix to a hotfix directly to master.

SnowballAntrobus:
Closing Statement:
During our integration with CLI we noticed that our library was compiling incorrectly. We fixed it so that it followed the project standard of being a static library. In future work we will have to further change the makefile to accommodate integration with other groups such as UI and GS for the purposes of using their functions or just testing.

mpcsh:
Issue Score: ✓+

Comments:
Great work!

PR #210, opened 2019-05-29
Hotfix: Changed makefile to make static library

SnowballAntrobus:
Currently our Makefile compiles into a dynamic library. The project specification says that it must be a static library. The goal of this issue is to make our Makefile in line with the project spec. At the request of CLI we are pushing this directly to master as a hotfix. This is the PR for #198 .

### DIFF
--- a/src/action_management/Makefile
+++ b/src/action_management/Makefile
@@ -4,7 +4,7 @@
 
 CC = gcc
 SRCS = src/actionmanagement.c src/get_actions.c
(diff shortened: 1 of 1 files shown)


SnowballAntrobus:
Closing Statement:
We fixed our makefile so that it followed the project standard of being a static library. In future work we will have to further change the makefile to accommodate integration with other groups such as UI and GS for the purposes of using their functions or just testing.

mpcsh:
Issue Score: ✓+

Comments:
Great work!

PR #211, opened 2019-05-29
Cli/integrate with am

ecoble:
This PR integrates our code with action management. We are now calling action management's functions in our type action operations, which are associated with different commands through the hashtable. We have also implemented the add_action_entries for the hashtable. This also includes a Makefile fix so that we can use other teams functions, and we no longer need to use absolute paths in our include statements. (These include statements have already been updated in this branch). Currently calling any actions will result in the action error operations, but this will be fixed once action management gets their hotfix through.

### DIFF
--- a/src/cli/example/Makefile
+++ b/src/cli/example/Makefile
@@ -1,6 +1,6 @@
 CC = gcc
 CFLAGS = -fPIC -Wall -Wextra -O2 -g -I../include/ -I../../common/include/ -I../../game-state/include/ -I../../action_management/include/
-LDFLAGS = -L../ -Wl,-rpath,.
+LDFLAGS = -L../ -L../../action_management/ -Wl,-rpath,../ -Wl,-rpath,../../action_management/
 RM = rm -f
 LDLIBS = -lreadline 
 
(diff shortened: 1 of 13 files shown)


src/cli/example/Makefile
```diff
@@ -1,18 +1,25 @@
...
 
-testshell: testshell.o ../cli.a 
-	$(CC) $(STC_LDFLAGS) $^ -o$@ -l:cli.a -lreadline
+$(BIN): $(OBJS)
+	$(CC) $(LDFLAGS) $(CFLAGS) $(OBJS) $(LDLIBS) -o$(BIN) -l:cli.a -l:action_management.a
+
+$(SRCS:.c=.d):%.d:%.c
+	$(CC) $(CFLAGS) -MM $< >$@
+
+include $(SRCS:.c=.d)
```
efraimdahl: GOOD Job reworking that Makefile -> That was a pain

lydiafilipe:
Issue Score: ✓-

Comments:
Nice work with the summary and referencing issues. However, there is no closing statement or much detailed comments/feedback. Also, you should not merge without all reviewer's explicit approval

PR #212, opened 2019-05-29
PR to merge new actions in item struct into game-state/dev

lindsey-hsi:
Work done in this branch (branched from dev shortly after adding a basic action struct to attributes and modifying relevant Makefiles to ensure compilation):
- wrote action new, init, and free functions, so that when actions are added to the attributes hash, they are initialized as a string key with a corresponding pointer to an action struct as its value
- added type-specific functions to accommodate getting/setting/changing action attributes, in line with the existing functions
- modified existing functions to accommodate for actions (ex. add_allowed_action, attribute_free)
- wrote/modifying corresponding tests --> at this point, everything compiles with 100 tests passing. 

Work to be done once merged with dev:
- figure out why the build does not pass so that this work can be included in the current gigantic PR to master

Work to be done in Sprint 4:
- write an action_equal function to be incorporated into attribute_equal (this was left out now because I wanted to get just a simple action component working so that we can move towards a working prototype)
- further develop action component in close collaboration with WDL and action management


src/game-state/include/item.h
```diff
@@ -50,17 +50,40 @@ int item_init(item_t *new_item, char *item_id, char *short_desc, char *long_desc
...
-// action type struct modeled from action management
-// typedef struct action_type {
-//     char *c_name;
-//     enum action_kind kind;
-// } atype;
+// ACTION STRUCTURE DEFINITION + BASIC FUNCTIONS ------------------------------
 
 typedef struct action {
+    char *action_name;
     action_type_t *action_type;
```
patrickkoopmans: It would be helpful to indicate that action_type_t is from AM's modules

src/game-state/include/item.h
```diff
@@ -165,6 +189,16 @@ int set_char_attr(item_t* item, char* attr_name, char value);
...
+/* set_act_attr() sets the value of an attribute of an item to the given int
+ * Parameters:
+ *  a pointer to the item
+ *  the attribute name with value to be changed
+ *  the action attribute value to be set
+ * Returns:
+ *  SUCCESS if successful, FAILURE if failed
+ *  returns SUCCESS if given value is already the attribute value
+ */
+int set_act_attr(item_t* item, char* attr_name, action_type_t *value);
```
patrickkoopmans: Is there a reason why setting the action (set_act_attr) is a different operation from creating the action (action_new)? I think that it might be more intuitive to have it bundled as one operation, unless you have been asked to separate them by AM. Making setting and creating the action one function call would be more similar to the way we have been dealing with attributes. Also, if I understand correctly, actions are associated with items, so I am not entirely sure as to why one would be made to exist on its own using action_new().

src/game-state/include/item.h
```diff
@@ -50,17 +50,40 @@ int item_init(item_t *new_item, char *item_id, char *short_desc, char *long_desc
...
-// ACTION STRUCTURE DEFINITION ------------------------------------------------
-// action type struct modeled from action management
-// typedef struct action_type {
-//     char *c_name;
-//     enum action_kind kind;
-// } atype;
+// ACTION STRUCTURE DEFINITION + BASIC FUNCTIONS ------------------------------
 
 typedef struct action {
+    char *action_name;
```
patrickkoopmans: It would be helpful to specify that the action_type struct is coming from AM's modules. I was initially confused by this.

src/game-state/src/item.c
```diff
@@ -207,11 +239,13 @@ int set_act_attr(item_t* item, char* attr_name, game_action value)
 
     else
     {
-        res->attribute_value.act_val = value;
+        // this needs to be discussed
+        res->attribute_value.act_val = action_new(attr_name, value);
```
patrickkoopmans: Regarding your comment, the second line is definitely more correct as res->attribute_value.act_val is of type game_action_type rather than action_type.

src/game-state/src/item.c
```diff
@@ -38,6 +38,38 @@ item_t *item_new(char *item_id, char *short_desc, char *long_desc)
...
+    assert(new_action != NULL);
+
+    strcpy(new_action->action_name, act_name);
+    new_action->action_type = act_type;
+
+    return SUCCESS;
+}
+
+/* see item.h */
+game_action_t *action_new(char *act_name, action_type_t *act_type)
```
patrickkoopmans: It may be helpful to name this game_action_new rather than action_new to distinguish it from action management's action_new

src/game-state/src/item.c
```diff
@@ -296,8 +330,9 @@ bool get_bool_attr(item_t *item, char* attr_name) {
     return res->attribute_value.bool_val;
 }
 
-game_action get_act_attr(item_t *item, char* attr_name) {
+game_action_t *get_act_attr(item_t *item, char* attr_name) {
```
patrickkoopmans: Should this return type game_action_t or action_type_t?

src/game-state/src/item.c
```diff
@@ -189,15 +221,15 @@ int set_bool_attr(item_t* item, char* attr_name, bool value)
...
+int set_act_attr(item_t* item, char* attr_name, action_type_t *value)
 {
     attribute_t* res = get_attribute(item, attr_name);
     if (res == NULL)
     {
         attribute_t* new_attribute = malloc(sizeof(attribute_t));
         new_attribute->attribute_key = (char*)malloc(100);
         new_attribute->attribute_tag = ACTION;
-        new_attribute->attribute_value.act_val = value;
+        new_attribute->attribute_value.act_val = action_new(attr_name, value);
```
patrickkoopmans: Regarding my first comment on bundling the operations of action_new and set_action, I see here that it action_new is mainly used to be called in set_action. Thus, I think we should remove action_new from item.h

src/game-state/include/item.h
```diff
@@ -50,17 +50,40 @@ int item_init(item_t *new_item, char *item_id, char *short_desc, char *long_desc
...
+/* action_init() initializes an action struct with given values
+   arguments are taken from action management
+ Parameters:
+    a memory allocated new action pointer
+    an action name
+    an action type struct
+ Returns:
+    FAILURE for failure, SUCCESS for success
+*/
+int action_init(game_action_t *new_action, char *act_name, action_type_t *act_type);
```
patrickkoopmans: Leave this out of the header file? Not sure but I think all opf our init funcs should be moved to internal header files so they can be tested but not face the public

src/game-state/include/item.h
lindsey-hsi: ah you're right I had a comment there but must've deleted it!

src/game-state/src/item.c
lindsey-hsi: you're definitely right that was also why I named the struct game_action

src/game-state/src/item.c
lindsey-hsi: I think it should be game_action_t since that's technically the attribute value. what do you think?

src/game-state/include/item.h
lindsey-hsi: yes! will do.

src/game-state/src/item.c
lindsey-hsi: sounds good, will do!

src/game-state/src/item.c
patrickkoopmans: I just think in the particular use cases, action management will want to deal with the structs they are familiar with (thus, action_type_t). Even though we technically wrap the attribute in the game_action_t struct, this is only for internal use reasons and increases the cognitive load of our module should we pass it as a return type.

lindsey-hsi: This definitely makes sense but since the struct is about to be expanded to include conditions and effects I think it should be game_action_t - AM can access type through it!

lydiafilipe:
Issue Score: ✓

Comments:
Good work! Great title, summary and discussions. Please remember the closing statement though. Also it looks like assignees was not filled out (you need this to get points for it).

PR #228, opened 2019-05-30
Hotfix: Memset everything immediately after malloc & fix errors for desc strings in room

liuyifan-eric:
- Memset everything to 0 (NULL) immediately after malloc each struct, so that ui team doesn't get unexpected errors
 - Fix errors when init description strings of rooms: malloc and strcpy for the id, long_desc, & short_desc

### DIFF
--- a/src/game-state/src/game.c
+++ b/src/game-state/src/game.c
@@ -5,8 +5,7 @@
 /* see game.h */
 game_t *game_new() {
     game_t *game = malloc(sizeof(game_t));
-    game->all_players = NULL; //helper fxn to get list of players
-    game->all_rooms = NULL;
+    memset(game, 0, sizeof(game_t));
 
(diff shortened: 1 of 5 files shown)


liuyifan-eric:
These updates should resolve the bugs that happened when the UI team call our functions. 
All malloc'd memories are initialized now;
When storing description strings in the structs, we now make sure that we copy the strings to the new space malloc'd, instead of directly change the pointer.

liuyifan-eric:
This hotfix pull request fixed two bugs:
 - memset everything in the structs to NULL when memory is malloc'd
 - malloc & strcpy (instead of directly changing the pointer) the desc strings when initilizing a room struct

lydiafilipe:
Issue Score: ✓+

Comments:
Nice work!

Issue #181, opened 2019-05-27
Make tests for do_path_action


ellychoi:
The object of this issue is to make multiple tests for do_path_action, covering all possible outcomes of the function in order to assure that it works correctly. Since the functions at this time are still fairly simple, the outputs of this tests will show whether they are printing the correct strings or not.

This was a branch off issue #177

SnowballAntrobus:
Created task #191 to add new tests after function has been updated to use game state functions.

SnowballAntrobus:
There is now a PR for this issue into dev #243 .

SnowballAntrobus:
Closing statement:
We wrote basic tests for the function do_path_action. These tests focused on making sure that the correct string is being printed when correct and incorrect action kinds are used. In future work we would make testing simpler by making the return value of the function an error code int. Additionally, we will make more tests for this function once it is updated to change game state.

mpcsh:
Issue Score: ✓

Comments:
Essentially no progress updates.

PR #186, opened 2019-05-28
New item functionality for AM and WDL + Tests written for all existing and new code within game-state modules

patrickkoopmans:
This pull request is made to merge the work of game-state during Sprint 3 to the master branch. The large size of the pull request is due to the fact that we had many many dependencies in our code that were constantly being updated.

The updates/ additions include: 
1. Tests for rooms, items, player, and game found within the game-state/tests directory (Issue #135, #136, #137, #138, #158)
2. Functions for AM and WDL to assign actions to items (Issue #157 and #172)
3. A function for UI to get items within a room (Issue #152 )
4. Several functions for UI that finds a room given different inputs (e.g. a path, or a room + direction, or a game and a room id) (Issue #151)
5. A function that creates a connection between two rooms (Issue #163)
6. Other minor changes including the addition of an "intro" member to the game struct (Issue #171), and a reworking of the internal usage of uthash macros across all modules

PR #243, opened 2019-05-31
Added tests for do_path_action

SnowballAntrobus:
We made multiple tests for do_path_action, covering all possible outcomes of the function in order to insure that it works correctly. Since the functions at this time are still fairly simple, the outputs of this tests will show whether they are printing the correct strings or not. This is the PR for issue #181 .


SnowballAntrobus:
Of course, we now need to check if it compiles post GS PR #186

src/action_management/tests/test_path_action.c
```diff
@@ -0,0 +1,70 @@
...
+#include "actionmanagement.h"
+#include "action_structs.h"
+#include "game.h"
+#include "path.h"
+
+bool execute_do_path_action(char *c_name, enum action_kind kind)
+{
+    char *dest = room_new("dummyroom", "a dummy room", "a placeholder room");
+    char *direction = "south";
+    game_t *g = game_new;
```
MTUC08: Should this be game_new()?

src/action_management/tests/test_path_action.c
```diff
@@ -0,0 +1,70 @@
...
+    
+
+    char *expected_output = malloc(100); // buffer
+    expected_output[0] = '\0';
+    strcat(expected_output, "Requested action ");
+    strcat(expected_output, c_name);
+    strcat(expected_output, " in direction ");
+    strcat(expected_output, p->direction);
+    strcat(expected_output, " into room");
+    strcat(expected_output, p->dest);
```
MTUC08: This is a carry-over typo from before I changed it in actions/dev, it should be p->dest->room_id.

MTUC08:
Note: make sure you don't merge to actions/dev until #244 goes through

src/action_management/tests/test_path_action.c
SnowballAntrobus: haha well it doesn't matter now but actually that is valid in C (CLI taught me this)

SnowballAntrobus:
Everything compiles and all tests pass

SnowballAntrobus:
Closing statement:
We wrote basic tests for the function do_path_action. These tests focused on making sure that the correct string is being printed when correct and incorrect action kinds are used. In future work we would make testing simpler by making the return value of the function an error code int. Additionally, we will make more tests for this function once it is updated to change game state.

mpcsh:
Issue Score: ✓+

Comments:
Good work, but the closing statement only came well after the fact.

Issue #241, opened 2019-05-31
Add tests for do_item_action checking for when an action is and isn't allowed on an item


SnowballAntrobus:
The goal is to make more tests for do_item_action that cover the possibility that the action cannot be performed on an item because it is not in its list of allowed actions.

SnowballAntrobus:
Closing statement:
We wrote more tests for the function do_path_action. These tests focused on making sure that the correct string is being printed when allowed and un-allowed actions are used on the item, which we call direct. In future work we would make testing simpler by making the return value of the function an error code int. Additionally, we will make more tests for this function once it is updated to change game state.

mpcsh:
Issue Score: ✓-

Comments:
No progress updates. Closing statement only well after the fact.

Issue #180, opened 2019-05-27
Make basic tests for do_item_action


lucyyutingli:
The object of this issue is to make multiple tests for do_item_action, covering all possible outcomes of the function in order to assure that it works correctly. Since the functions at this time are still fairly simple, the outputs of this tests will show whether they are printing the correct strings or not.

This was a branch off issue #177

SnowballAntrobus:
Created task #192 to add new tests after function has been updated to use game state functions in #190

lucyyutingli:
Since we forgot to add a closing statement, I'll be adding one now. 

Closing Statement:
As of now the these tests were created before we altered our functions to return ints instead of chars, so they're fairly simple and only test for the correct output. Later on, we'll be editing them so they test for correct conditions met as well, once we have Game State functions.

mpcsh:
Issue Score: ✓

Comments:
Progress updates were skimpy (and only improved well after the fact).

PR #249, opened 2019-05-31
Added tests for do_item_action

SnowballAntrobus:
We made multiple tests for function do_item_action. They cover the cases of using correct and incorrect kinds with the function. Tests to check for items being allowed by actions have been added. For more information refer to #180 and #241 respectively.


src/action_management/tests/test_item_actions.c
```diff
@@ -0,0 +1,84 @@
...
+
+    cr_assert_eq(rc, false,
+                 "execute_do_item_action returned true for wrong kind 2");
+}
+
+Test(item_actions, wrong_kind_3)
+{
+    bool rc = execute_do_item_action("dummy", ITEM_ITEM, "dummy", ITEM_ITEM);
+
+    cr_assert_eq(rc, true,
```
MTUC08: should this be false?

src/action_management/tests/test_item_actions.c
```diff
@@ -0,0 +1,84 @@
...
+    game_free(g);
+
+    return rc;
+}
+
+Test(item_actions, correct_kind_1_and_allowed_action)
+{
+    bool rc = execute_do_item_action("dummy", ITEM, "dummy", ITEM);
+
+    cr_assert_eq(rc, false,
```
MTUC08: Should this be true?

SnowballAntrobus: correct

src/action_management/tests/test_item_actions.c
SnowballAntrobus: correct lol oops

SnowballAntrobus:
Everything compiles and all tests pass

SnowballAntrobus:
Closing statement:
We wrote more tests for the function do_item_action. These tests focused on making sure that the correct string is being printed when the correct and incorrect kind is used and when allowed and un-allowed actions are used on the item, which we call direct. In future work we would make testing simpler by making the return value of the function an error code int. Additionally, we will make more tests for this function once it is updated to change game state.

mpcsh:
Issue Score: ✓+

Comments:
Good work, but the closing statement only came well after the fact.

Issue #202, opened 2019-05-28
Create functions to return short and long string descriptions from items


jessxia:
Write functions to return the short and long descriptions stored in our item structs. This will be used in WDL's tests of their load_item functions (issue #179). It will also be useful for returning descriptions for actions like LOOK.

jessxia:
Added said functions; then added condition so that NULL items to return NULL. Added tests for both created and NULL items in test_items.c.

lydiafilipe:
Issue Score: ✓+

Comments:
Nice work! It seems like this was a small issue so progress updates weren't super necessary but make sure to keep up with those on bigger issues.

PR #254, opened 2019-06-01
Adding get description from item functions to game-state/dev

jessxia:
Added two functions, get_sdesc_item and get_ldesc_item into item.h and item.c. These functions return the short and long descriptions stored in an item (for issue #202). Also added two tests into test_items.c to test these functions.

### DIFF
--- a/src/game-state/include/item.h
+++ b/src/game-state/include/item.h
@@ -74,6 +74,7 @@ int item_free(item_t *item_tofree);
  *
  * Returns:
  *  short description string
+ *  NULL if item is NULL
  */
 char *get_sdesc_item(item_t *item);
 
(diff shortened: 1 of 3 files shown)


a-matai:
Can you add a test for if there is no ldesc / if taken from a null item? 

Code is fine, but I'd suggest a check incase a null item is fed in?

jessxia:
NULL items now return NULL for both functions :))

a-matai:
Yay!

jessxia:
Get_sdesc_item and get_ldesc_item were added to our item module to return short and long descriptions of items, or NULL if NULL item. Tests were written and pass.

lydiafilipe:
Issue Score: ✓+

Comments:
Nice work! Though it's important that the reviewers approve explicitly on github.

PR #258, opened 2019-06-02
Cli integration with UI and new context struct

ecoble:
This pull request completes integration of UI and CLI from CLI's end. Changes made include:

- Adding the CLI hashtable to the chiventure context struct, and integrating our code with this struct. UI will need access to our hashtable since it is now a parameter to a couple of the functions they are using, so we have added it to the context struct. Places where functions used to take a game struct or hashtable as a parameter now take a chiventure context struct.

- Formal new, init, and free functions for cmd struct and hashtable struct. Most of these functions already existed or had different names, but there are now conventional new, init, and free functions for both of these structs.

- Printing to the UI window. All places where we were printing to the console have been replaced with a call to print to UI's window.

- Adding "MAP" and "SWITCH" as valid commands. User can now type "MAP" to show the map, or "SWITCH" to switch the position of the main window and the cli

Notes/Considerations:

- The "MAP" command currently does not print anything, but this should be fixed soon.

- We have removed the "HIST" command, which showed the user's command history, as this was causing difficulties both in this issue and for testing, CLI will try to support this command again in the future.

- For now we have also replaced the "QUIT" command with CTRL + D to quit the window, in the future we will also try to support "QUIT"

- The CLI example is broken in this branch. I have not made great efforts to fix it since all of our output is now being routed to UI's window anyways, so even if the example compiled it would not be of much use anyways. Going forward to test CLI output, UI's example should be used.

- I had to add a parameter "char *dummy" to UI's toggle_map and layout_switch commands in order to integrate it with our hashtable. This is because CLI's hashtable requires functions to map to operations that take both a string and a chiventure context struct as a parameter. When calling these functions it is fine to pass NULL as the parameter to the string, since it is not used anyways and exists only for type matching purposes. This is not a great solution, and in the future CLI would like to make this more flexible, but for now it is a fix to get everything up and running.

- We are unable to test most functions right now, since the backend is not complete. Commands such as "LOOK" or "GIVE" will just result in a seg fault, which is expected behavior. Commands that will give visible results include "HELP", which prints help text, "SWITCH", which switches the main window and cli position, and any invalid command (e.g. "DANCE"), which will print the action error message.


src/cli/include/cmd.h
```diff
@@ -27,7 +28,9 @@ typedef struct
     operation *func_of_cmd;
 } cmd;
 
+#include "ctx.h"
```
Tianlel: Probably should move these to the top?

src/cli/include/cmd.h
```diff
@@ -114,24 +117,56 @@ void delete_entry(char *command_name, lookup_t **table);
...
-/* Heap allocates a new cmd struct
+/* Frees resources associated with the lookup hashtable struct
  *
+ * Parameters:
+ * - pointer to a lookup hashtable
+ *
+ * Returns:
+ * - nothing
+ */
+void lookup_t_free();
```
Tianlel: I think this should take a lookup_t table as input. Since you still have the delete_entries function in this file, maybe you could just delete this and rename that function?

src/cli/include/cmd.h
```diff
@@ -114,24 +117,56 @@ void delete_entry(char *command_name, lookup_t **table);
 void delete_entries(lookup_t **table);
```
Tianlel: See the next comment

Tianlel:
Everything else looks good. I think we can delete the example directory since it won't be needed anymore.

src/cli/include/cmd.h
ecoble: These are actually intentionally after the structs, putting them at the top causes circular dependencies and the build will fail

src/cli/include/cmd.h
ecoble: My bad, fixed it in the .c file and forgot to change it in the .h. Thank you!

dbarer: You changed lookup_t_init to return an integer, but need to update the documentation to reflect that change. Just say Returns:
Success in all cases

src/ui/include/ui_ctx.h
```diff
@@ -86,7 +86,7 @@ int ui_ctx_free(ui_ctx_t *ui_ctx);
  * Does not return anything
```
lydiafilipe: It seems you guys are modifying some of ui's files. To confirm, was this done with them and are they on board with this?

src/cli/example/testshell.c
```diff
@@ -1,73 +0,0 @@
-/*
```
lydiafilipe: What's the motivation behind deleting these? If they still highlight how the functions are used they may be worth having.

src/ui/include/ui_ctx.h
ecoble: Yes, I have been in communication with UI while making these changes

src/cli/example/testshell.c
ecoble: We deleted the examples directory since running our example shell won't really do anything now that we print to the UI window, but I can put it back if you think it is valuable.

ecoble: UI however has a very similar block of code that they lifted from us in their print_functions.c file, so there is still an example of use of this code

lydiafilipe: I think it's always good to keep examples. Generally if you make a library, multiple other projects might use it, but it's still always good to have examples for reference

ecoble: Okay, I have added the examples back and also updated them to use the chiventure context struct

Gestalt-R:
#268 Aims to fix the operations issue introduced in this pull request. The code as is is stable, but unnecessarily messy.

ecoble:
CLOSING STATEMENT: CLI and UI are not integrated, with all CLI output printing to the UI window and the addition of two UI functions, 'MAP' and 'SWITCH' to the hashtable. A first step to this integration was integrating CLI code with the new chiventure context struct, and adding the CLI hashtable to the chiventure context struct, both of which were done in this PR.

Gestalt-R:
Also, when testing my code, noticed that the "quit" command is not integrated. This should be fixed before this is closed.

ecoble:
In response to the above comment about the quit command not being integrated, for now we are using CTRL + D to quit from the window, since CLI is no longer in control of the loop that runs the prompt. This was noted in the "Notes/Considerations" in the first comment. Supporting "QUIT" as a prompt can be a future issue.

ecoble:
CLOSING STATEMENT: CLI and UI are not integrated, with all CLI output printing to the UI window and the addition of two UI functions, 'MAP' and 'SWITCH' to the hashtable. A first step to this integration was integrating CLI code with the new chiventure context struct, and adding the CLI hashtable to the chiventure context struct, both of which were done in this PR.

lydiafilipe:
Issue Score: ✓+

Comments:
Great work!

PR #266, opened 2019-06-03
Wdl/wdl spec draft

thoranderskow:
Updated wdl.md specification to reinclude conditions and effects in the item document object. Also got rid of the "through" field.


lydiafilipe:
It looks like there are merge conflicts. Please resolve these first

bansals10:
The merge conflicts have been resolved within the WDL file.

docs/wdl.md
```diff
@@ -40,12 +41,20 @@ Please see this document for all possible actions and descriptions: https://gith
...
+   2. The player is located in a specific room
+    - Ex. in_room: "BEDROOM"
+
 ### GAME example:
 ```yaml
  - start: "KITCHEN"
  - intro: “Welcome to the virtual house. You have been wandering for quite some time,
    and you need to determine how to return to reality.”
-
+ - end:
```
mewthew: What do you guys think of instead of having the start, intro, and end being different items, we can have it just be one list starting from start, for example:

- start
  intro
  end
    - in_room  

That way it mirrors what we've done for rooms/items.

bansals10: We followed the syntax given in the distaff.wdl sample WDL file for game. Tick marks are placed in front of each field in this object because the start, intro, and end are distinct from one another.

docs/wdl.md
```diff
@@ -63,10 +72,8 @@ Please see this document for all possible actions and descriptions: https://gith
     - to: `<ROOM  ID>` which lists a valid place the player can reach in one action from this room by ID
 
       direction: `<CARDINAL DIRECTION>` which states the direction that connection is in. Only six directions are available for use in the game: north, east, south, west, up, down. 
-
-      through: `<ITEM ID>` if applicable, the item that the player must go through to go in that direction
-
-    ###### NOTE: a valid connection has to have an ID that exists. 
+
```
jessxia: Can you explain your reasoning for removing the "through" field for room connections? Game-state has (tentatively) added a pointer to an item in our path struct in our game-state/develop_actions branch to point to the item that is being moved through. Should we change that?

bansals10: As per our conversation with Borja, the "through" field is an externality that is not critical to this portion of the project. We have created a backlog issue to update the spec to include this field, but for the sake of simplicity in this sprint of the project, we are not including the "through" field. We consider the "to" and "direction" fields as being sufficient at the current stage. Game-state can continue to have a pointer to an item within the path struct, but this will be set to NULL when the room structs are loaded until further notice.

mewthew:
Master's wdl doc is updated to include conditions and effects in the item document object. We have also gotten rid of the "though" field.

lydiafilipe:
Issue Score: ✓+

Comments:
Nice work! The title is quite vague, though everything else is good.

PR #282, opened 2019-06-05
Checkpointing/temp

schin99:
we thought dev was updated with our most recent code but we were wrong. this is to fix it


schin99:
this was a mistake, first need to merge master into temp

Issue #288, opened 2019-06-05
Update do_item_item_action tests to reflect #245 


MTUC08:
In #245 , we added a new out parameter to hold the status string returned by our do_item_item_action function, and changed the return format to return an int reflecting the status of the action. However, this change will break our original do_item_item_action tests, so the goal of this issue is to update/fix our do_item_item_action tests to reflect the new parameters and return format.

SnowballAntrobus:
Complies and all tests pass in branch actions_change_return_format_tests

SnowballAntrobus:
Closing Statement:
In #245 we finally changed the return value of our function do_item_item_action to output an error code int instead of a string. This makes testing much more easy and reliable since we no longer compare strings, instead just checking the return value. In future work we will update the function once again this time to change game state, i.e move the player, and check for conditions before executing. We will be writing tests for these new features as well.

SnowballAntrobus:
It is important to note that we defined various macros for the error code ints the function outputs. This function is particular in that it has an additional macro for when the function fails on a not allowed action in the direct vs in the indirect.

mpcsh:
Issue Score: ✓+

Comments:
Great work!

Issue #245, opened 2019-05-31
Refactor AM modules for ease of unit testing


MTUC08:
Refactor action functions (do_item_action, do_path_action, do_item_item_action) to pass the string to CLI in an output parameter, so we can have the functions return SUCCESS(0) or FAILURE(1) for ease of testing. This prevents complete test failure if we alter the passing string within the function. This will only alter the code in PR #244  so the functions do not deal with conditions or attributes yet. We will add conditions/attributes through issues #188 #190 and #191 .

Since this will directly impact CLI, this will be a joint venture with the CLI team. Their changes to preemptively fix the code this change will break will also be added to our feature branch actions/change_return_format, so that it can be carried into actions/dev and then master without creating a separate PR.

MTUC08:
Update: just changed the description of the functions within actionmanagement.h and the reformatting within actionmanagement.c

yoonsjeong:
I recall during the sprint meeting Borja proposed that we also have different error outputs -- for example, output the integer 1 when kind is incorrect, and integer 2 when not an allowed action.

MTUC08:
@borjasotomayor just to confirm, should we still have different integer outputs if we are additionally outputting a message? Currently, the function will output 0 upon success, with a success message, and 1 upon failure, with different error messages. Since our functions will be used by CLI, who will likely just pass the outputted string to the user, are different integers needed for different errors?

borjasotomayor:
Yes, different integers will allow your function to be more testable, because your tests can then check whether it is correctly distinguishing between the different types of errors.

MTUC08:
Update: updated the action management.h file with descriptions of which errors produce which ints as output, and defined 1 as WRONG_KIND and 2 as NOT_ALLOWED within the code in actionmanagement.c

ecoble:
UPDATE: Updated CLI's use of action management's functions, we now malloc a string to pass to each action management function, and then print this string after the function call.

Note: For now I have put in printf, but once the PR for CLI and UI integration is through I will be changing this to print to the UI window

ecoble:
UPDATE: Changed CLI use of action management's functions, before I was using malloc to create a string to pass to AM, but realized this was unnecessary since they malloc a string. We are now passing a null pointer.

borjasotomayor:
[...]

PR #204, opened 2019-05-28
Added tests for do_item_item_action

SnowballAntrobus:
Made multiple tests for function do_item_item_action. They cover the cases of using correct and incorrect kinds with the function. Tests to check for items being allowed by actions have been added. For more information refer to #182 and #242 respectively.

PR #289, opened 2019-06-05
Added tests for actionmanagement.c functions 

SnowballAntrobus:
Changes:

- Bug fixed action_type_free as it was preventing compilation when used
- Fixed do_item_item_action so that it now also checks to see if the direct item supports the action
- Created a new file test_action_new_init_free.c that includes tests for the new, init, and free action_type functions, using criterion.
- Created a new file test_item_action.c that includes tests for the do_item_action function. They test for correct action kind and allowed_action on the item, using criterion. 
- Created a new file test_item_item_action.c that includes tests for the do_item_item_action function. They test for correct action kind and allowed_action on the items, using criterion. 
- Created a new file test_path_action.c that includes tests for the do_path_action function. They test for correct action kind. 
- altered action functions to return an int descriptive of the status of the action called (#245 ) with a status string (success/failure) as an out parameter
- altered test c files to reflect the altered returned format (#286 , #287 , #288 )
- altered cli code to reflect changes from #245


SnowballAntrobus:
Yoon's comments on the do_item_item_action tests PR #204 that apply to other test files:

- malloc size in bytes was replaced with a macro scale variable BUFFER_SIZE in the actions/dev branch. I think you should merge actions/dev into this branch
- all "strcat"s have been replaced with sprintf()
- there will need to be two else if's given the new code in actions/dev, returns 1 if there is one particular error, returns 2 if there is another particular error. We will also need tests to check for failure corresponding to the error codes 1 and 2.
- I just realized that using booleans for return codes may be confusing.
SUCCESS = 0, FAILURE = 1; but true = 1, false = 0. Since we're using the failure return codes 1 and 2, I think you should change rc just to be an integer.

SnowballAntrobus:
Made changes in response to Yoon's comments that I listed in my previous comment.

src/action_management/src/actionmanagement.c
```diff
@@ -122,8 +120,16 @@ char *do_item_item_action(game_t *g, action_type_t *a,
         sprintf(ret_string, "The action type provided is not of the correct kind");
         return ret_string;
     }
-    // checks if the action can be used on the item
-    int allowed = allowed_action(indirect, a->c_name);
+    // checks if the action can be used on the direct item
+    int allowed = allowed_action(direct, a->c_name);
+    if (allowed != SUCCESS)
```
yoonsjeong: it appears there was a duplication error of some sort during the merge! (see line 133)

src/action_management/src/actionmanagement.c
```diff
@@ -122,8 +120,16 @@ char *do_item_item_action(game_t *g, action_type_t *a,
...
+    {
+        sprintf(ret_string, "Action %s can't be requested on item %s",
+                a->c_name, direct->item_id);
+        return ret_string;
+    }
+    // checks if the action can be used on the indirect item
+    allowed = allowed_action(indirect, a->c_name);
     if (allowed != SUCCESS) {
```
yoonsjeong: if statement here is identical to line 125

src/action_management/tests/test_item_actions.c
```diff
@@ -0,0 +1,101 @@
...
+    {
+        rc = SUCCESS;
+    }
+    else if (strcmp(do_item_action(g, a, item), kind_error) == 0)
+    {
+        rc = FAILURE;
+    }
+    else if (strcmp(do_item_action(g, a, item), allowed_error) == 0)
+    {
+        rc = 2;
```
yoonsjeong: might not hurt to define macros for 1, 2, and 4 as well since FAILURE can come in many forms -- I just checked our sprint 4 (not yet on dev) code,
we have
`#define WRONG_KIND 1`
`#define NOT_ALLOWED 2`
I think it wouldn't hurt to add this to the tests
`#define TEST_FAIL 4`
or something similar. is there a reason you used the rc = 4 rather than 3? (3 is currently not being used by any other returns)

src/action_management/tests/test_item_item_actions.c
```diff
@@ -0,0 +1,131 @@
...
+            a->c_name, indirect->item_id);
+
+    int rc;
+    if (strcmp(do_item_item_action(g, a, direct, indirect), expected_output) == 0) 
+    {
+        rc = SUCCESS;
+    }
+    else if (strcmp(do_item_item_action(g, a, direct, indirect), kind_error) == 0)
+    {
+            rc = FAILURE;
```
yoonsjeong: I do wonder if we could just have different macros for each test so they don't even depend on common.h
-- so for example at the beginning of each test framework
```
#define TEST_SUCCESS 0
#define KIND_ERROR 1
#define DIRECT_ERROR 2
#define INDIRECT_ERROR 3
#define OUTPUT_ERROR 4
```

yoonsjeong: and then since the test files don't depend on each other (at least, I believe they don't) this shouldn't lead to compilation errors

src/action_management/src/actionmanagement.c
MTUC08: This was originally meant to check that the action could be used on both indirect and direct items, but I agree that the direct check should be removed in later PRs since the game state implementation of conditions would have us just check that for example, the item key exists and is associated with the action use, and can be used on the door. For this PR, this is fine.

src/action_management/src/actionmanagement.c
MTUC08: See above

src/action_management/tests/test_item_item_actions.c
MTUC08: At that point, we might as well add a AM common file

yoonsjeong: @SnowballAntrobus thoughts?
I will approve the changes for now since I don't see anything really holding this PR back now

src/action_management/tests/test_item_actions.c
```diff
@@ -0,0 +1,101 @@
+#include <criterion/criterion.h>
```
ecoble: These tests (as well as test_path_actions and test_item_item_actions) feel like they are bordering more on integration tests than unit tests, especially with all the dummy structs and stuff you guys are creating. Not sure if this is a problem, but I know for CLI Borja told us to try and stick to unit testing. Also people are going to start working on a shared dummy game that people can use for testing in the future, so you might want to keep a look out for that or even help build it, and use that for these tests later on instead.

src/action_management/tests/test_action_new_init_free.c
```diff
@@ -0,0 +1,153 @@
...
+
+/* Checks that a new action type, kind 1 is initiated without interruption */	
+Test(action_type, init_ITEM)	
+{	
+    action_type_t a;	
+    int rc;	
+    char* c_name = "TAKE";	
+    enum action_kind kind = ITEM;	
+
+     rc = action_type_init(&a, c_name, kind);
```
ecoble: Small thing, but this line and the following 2 lines are not properly aligned

src/action_management/tests/test_action_new_init_free.c
```diff
@@ -0,0 +1,153 @@
...
+
+ /* Checks that a new action type, kind 2 is initiated without interruption */	
+Test(action_type, init_PATH)	
+{	
+    action_type_t a;	
+    int rc;	
+    char* c_name = "GO";	
+    enum action_kind kind = PATH;	
+
+     rc = action_type_init(&a, c_name, kind);
```
ecoble: same here

src/action_management/tests/test_action_new_init_free.c
```diff
@@ -0,0 +1,153 @@
...
+
+ /* Checks that a new action type, kind 3 is initiated without interruption */	
+Test(action_type, init_ITEM_ITEM)	
+{	
+    action_type_t a;	
+    int rc;	
+    char* c_name = "PUT_ON";	
+    enum action_kind kind = ITEM_ITEM;	
+
+     rc = action_type_init(&a, c_name, kind);
```
ecoble: same here

src/action_management/tests/test_item_actions.c
SnowballAntrobus: This has been addressed in the new PR #291 . Do you think this can wait? The reason I'm using 4 is because in the tests for item_item_action we use 3 as error code NOT_ALLOWED_INDIRECT.

src/action_management/tests/test_action_new_init_free.c
MTUC08: note: can be done by running the command astyle --style=stroustrup <filename> in a linux environment

src/action_management/tests/test_item_item_actions.c
SnowballAntrobus: I think it would be good to approve this for now and make adding a common file an issue. Because as I said in my last comment this was fixed in #291 .

src/action_management/tests/test_item_actions.c
SnowballAntrobus: It was my understanding from talking to Borja and Mark that these were our unit tests. But to be honest, now that you mention it I'm not entirely sure. It's just that the functions won't run without them.

src/action_management/tests/test_action_new_init_free.c
```diff
@@ -0,0 +1,150 @@
...
+
+    cr_assert_not_null(new_action, "action_type_new failed");
+
+    cr_assert_eq(new_action->c_name, "USE_ON", "action_type_new() didn't set new_action->c_name");
+    cr_assert_eq(new_action->kind, ITEM_ITEM, "action_type_new() didn't set new_action->kind");
+
+    action_type_free(new_action);
+}
+
+/* Checks that a new action type, kind 1 is initiated without interruption */
```
mpcsh: `s/initiated/initialized/`
(make sure that gets changed everywhere)

src/action_management/tests/test_item_actions.c
mpcsh: This definitely can't wait - please make this change as part of this PR.

src/action_management/tests/test_item_actions.c
mpcsh: I have kind of the opposite question - I think these are fine in terms of scope (defining test fixtures doesn't make something not a unit test), but I'm just not sure what exactly is being tested here. @SnowballAntrobus can you elaborate on what this is supposed to be checking?

src/action_management/tests/test_item_actions.c
MTUC08: @mpcsh Can you elaborate on what change you want us to make? The change to returning different macros is done in #245 and was going to be the PR done after this one. Should we instead add those files to this PR?

mpcsh: Yes - you can do that by just merging that branch into `dev`

MTUC08:
Update: I will add a PR to merge the changed return format functions/tests into actions/dev for this PR

src/action_management/tests/test_item_actions.c
MTUC08: Just created PR #294 for this purpose

src/action_management/tests/test_item_actions.c
SnowballAntrobus: @mpcsh we are testing our do_action functions sans the conditions and effects from sprint 4. So really we are testing that the correct string is outputted. Like the PR description says, hopefully it’s not terrible, this means we test for validations. This means validating kind and validating that the action is allowed by the item.

src/action_management/tests/test_item_actions.c
ecoble: Would it be possible to just add the macros here and leave the other changes as a separate PR? I wanted to put a CLI reviewer on it and I know the review is mostly done here so I don't want to hold it up.

SnowballAntrobus:
Addressed @mpcsh comments

src/action_management/tests/test_item_actions.c
mpcsh: Right, ok. Makes sense.

src/action_management/tests/test_item_actions.c
mpcsh: Looks like @SnowballAntrobus did just that

mpcsh:
@SnowballAntrobus one of my change requests wasn't addressed - please make sure that gets taken care of

src/action_management/tests/test_action_new_init_free.c
MTUC08: @mpcsh what exactly does this mean?

lucyyutingli: If I read this correctly it was to change initiated to initialized. I just changed them in the tests.

mpcsh:
:tada:

SnowballAntrobus:
Closing Statement:
This PR is mostly tests, including new, init, and free tests for the action struct, and testing that the functions for each of the three kinds of actions function properly (do_item_action, do_path_action, and do_item_item_action). We also updated these action functions to return an error code int (0 for success, 1 for incorrect kinds, 2 for invalid direct object and action pairings, 3 for invalid indirect object and action pairings, and 4 for incorrect strings). We updated any files using these functions accordingly, including our tests and CLI files that called these functions. There are also bug fixes for action_type_free and do_item_item_action. In the future we will also update these functions to check for effects and conditions once game state's new functions come out.

mpcsh:
Issue Score: ✓+

Comments:
Great work!

PR #292, opened 2019-06-05
Checkpointing/savetesting into dev

jarvislam1999:
I got all the tests for the save functions to pass by fixing both save.c and test_save.c. I think our save functions are complete and ready for master. Would love some reviews.

Notes: we may need to add code to deal with action_t in the future
We need to make the tests more rigorous and careful, by also testing the values being passed in instead of just the length of the hashtable


src/checkpointing/tests/test_save.c
```diff
@@ -61,32 +52,47 @@ Test(save, item)
...
+    char *s_long_desc = "looks unwashed";
+
+    // item_new is a function in game-states's item.h. It creates an item_t struct.
+    shirt_t = item_new(s_item_id, s_short_desc, s_long_desc);
+    cr_assert_not_null(shirt_t, "shirt_t is null in save_room");
+
     path_t *path_t;
+
     // path_new is a function in game-states's path.h. It creates a path_t struct.
     path_t = path_new(room_t, "east");
```
chrisfernandes18: Do we still need this?

chrisfernandes18: as this i mean path stuff

jarvislam1999: Deleted

Issue #193, opened 2019-05-28
Update do_path_action tests to reflect #191


SnowballAntrobus:
The goal is to add multiple tests for do_item_item_action, corresponding with the changes from #191 . The additional tests should cover checking conditions and attribute changes.

This is the successor to #181

yoonsjeong:
^ accidentally clicked the close issue button (I got too excited)

PR #302 has been merged successfully into `actions/dev`.
Within this pull request was the implementation of the following
- do_path_function using game-state functions and chiventure_ctx struct (as per issue #191)
- tests with the appropriate game-state functions (as per this issue)
- removal of the "ENTER" action_type_t
- refactoring of the do_item and do_item_item function to no longer take in the game_t struct
- refactoring of tests related to do_item and do_item_item

the tests are successful and don't crash!

mpcsh:
Issue Score: ✓+

Comments:
Great work!

Issue #192, opened 2019-05-28
Update do_item_action tests to reflect #190


SnowballAntrobus:
The goal is to add multiple tests for do_item_action, corresponding with the changes from #190 . The additional tests should cover checking conditions and attribute changes.

This is the successor to #180

SnowballAntrobus:
These tests are being made using the tests from do_item_item_action as a template. Therefore if a bug is found in either of the test files it is a good idea to check the other one for it too.

SnowballAntrobus:
Work for this issue is being done on the branch actions/updatedo_item

SnowballAntrobus:
Closing Statement:
We have now finished writing tests that check the full functionality of the do_item_action function. In this issue we added tests for effects and conditions. We will proceed by checking for compatibility as the GS functions make it to master.

mpcsh:
Issue Score: ✓

Comments:
Definitely missing some progress updates here

Issue #191, opened 2019-05-28
Update do_path_action to accommodate condition checking and attribute changes


SnowballAntrobus:
Build upon the do_path_action basic prototype ( #115 ) to include condition checks, i.e. whether the item has the necessary attributes for the action to be performed, and attribute changes, i.e. the consequences of the action being requested. This will be done with game state functions.

yoonsjeong:
I've started work on this task, on a branch called `actions/do_path_191`.

yoonsjeong:
Hi @MTUC08 @SnowballAntrobus @lucyyutingli I've defined a new macro called `#define NOT_ALLOWED_PATH (4)` just a heads up in case any other macros turn up.

yoonsjeong:
PR #302 has been merged successfully into `actions/dev`.
Within this pull request was the implementation of the following
- do_path_function using game-state functions and chiventure_ctx struct
- tests with the appropriate game-state functions (as per #193)
- removal of the "ENTER" action_type_t
- refactoring of the do_item and do_item_item function to no longer take in the game_t struct
- refactoring of tests related to do_item and do_item_item

mpcsh:
Issue Score: ✓+

Comments:
Great work!

Issue #190, opened 2019-05-28
Update do_item_action to incorporate condition checking and attribute changes


SnowballAntrobus:
Build upon the do_item_action basic prototype ( #115 ) to include condition checks, i.e. whether the item has the necessary attributes for the action to be performed, and attribute changes, i.e. the consequences of the action being requested. This will be done with game state functions.

lucyyutingli:
Once #186 is approved, we will be using the following functions:

- In order to change attributes, we will use set_attribute() family of functions
- In order to check conditions, we will use get_attribute() family of functions

lucyyutingli:
The functions have been changed for the most part but we'll have to wait for game-state's functions to be pulled to master before we can be sure that they're working properly. We'll be waiting on that before we can make any further progress.

MTUC08:
Closing comments: in this issue, we updated the do_item_action function to handle conditions and effects. This was pulled into master with #364 . Looking forward, next years class can continue to evolve this function by adding more conditions (inventory, location) and effects (teleportation). Thanks for a great quarter!

mpcsh:
Issue Score: ✓+

Comments:
Great work!

PR #294, opened 2019-06-06
Changed the return format of AM functions

MTUC08:
This PR is the culmination of issues #245 #287 #288 #289 . It includes an altered return format for the do_item_action, do_path_action, and do_item_item_action functions. Where we originally had them return status strings for the basic sprint 3 prototype, we now have them return an int macro for easier testing, with a status string as an out parameter. It also includes cli code and updated AM function tests reflective of the change in format.


MTUC08:
Closing Statement:
In this PR we updated the return format of our AM functions to return int rather than a string, which we moved to be an out parameter in order to ease testing troubles. It will be pulled into master in #289 .

Looking Forward:
We will update the functions to handle condition checking and effect application #188 #190 #191 and update the tests accordingly #189 #192 #193.

mpcsh:
Issue Score: ✓+

Comments:
Great work!

PR #297, opened 2019-06-06
Wdl/wdl spec draft

thoranderskow:
Updated wdl.md in docs to add that actions must be in uppercase, and directions must be in lowercase. Also added that only being in a certain room is an end condition.


thoranderskow:
Branch was fouled up, ignore this pull request

Issue #205, opened 2019-05-28
Create UI Homescreen


omorkved:
Create a user-friendly homescreen

- Chiventure logo should appear when first loaded
- instructions should appear "press _something_ for help, press _something_ to show the map..."

etc.

sebvil:
In order to be able to use the Unicode characters used in the banner, we must call the function `    setlocale(LC_ALL, "")` (must include locale.h) before `initscr` is called. Thus, this is done in the ui context struct initialization. Furthermore, the we must use `lncursesw` instead of `-lncurses` in order to support the utf-8 locale.

sebvil:
Closing Statement:  Printing the home screen was not as trivial as it seemed at first. The ncurses standard library does not support for Unicode characters present in the banner, so when linking, we had to change `-lncurses` to `lncursesw`. Additionally, I had to call the command `setlocale(LC_ALL, "")` before `initscr` to support the utf-8 locale.. 

After this task, when `chiventure` is run, one sees the banner centered  on the top window of the UI, with the message "Type 'HELP' to show help menu" under it.

mpcsh:
Issue Score: ✓+

Comments:
Great work!

PR #301, opened 2019-06-07
Created homescreen

sebvil:
In this task I implemented a home screen which displays the Chiventure banner and the message "Type 'HELP' to show help menu" below it. I also included the banner in the chiventure context struct so that the UI team would be able to access it in order to print it, and called the `start_ui` function from the main chiventure.c file so that the UI is actually shown when the game is run. See #205 for more info on steps necessary to properly display the home screen using ncurses.


src/ui/src/print_functions.c
```diff
@@ -19,7 +19,28 @@
 /* see print_functions.h */
 void print_info(chiventure_ctx_t *ctx, window_t *win)
 {
-    mvwprintw(win->w, 1, 2, "Main Window");
+    int x_pos = COLS / 2 - 48;
```
ecoble: What are these numbers based on? Might be helpful to define macros so they aren't just "magic numbers"

src/chiventure.c
```diff
@@ -20,13 +21,11 @@ int main(int argc, char **argv)
 {
     chiventure_ctx_t ctx;
 
-    chiventure_ctx_init(&ctx);
+    chiventure_ctx_init(&ctx, banner);
```
ecoble: Not necessarily an issue but I am wondering why you chose to make the banner part of the context struct? If it's only really used once at the beginning of the program couldn't it just be a parameter to the start_ui function and be printed there?

Makefile
```diff
@@ -4,7 +4,7 @@ CC = gcc
 AR = ar
 CFLAGS = -fPIC -Wall -Wextra -O2 -g -I./include/ -I./src/common/include -I./src/game-state/include -I./src/ui/include -I./src/cli/include -I./src/action_management/include
 RM = rm -f
-LDLIBS = -lyaml -lncurses -lreadline
+LDLIBS = -lyaml -lncursesw -lreadline
```
mpcsh: Is this a typo?

src/chiventure.c
mpcsh: Yes, agreed. We should only be modifying UI code here.

src/common/include/ctx.h
```diff
@@ -16,6 +16,7 @@
 typedef struct chiventure_ctx
 {
     /* Add component-specific structs here */
+    const char *banner;
```
mpcsh: See @ecoble's comment above - we shouldn't be modifying the big context for this

src/ui/Makefile
```diff
@@ -5,7 +5,7 @@ AR = ar
 CFLAGS = -fPIC -Wall -Wextra -O2 -g -I./include/ -I../common/include/ -I../game-state/include/ -I../cli/include/ -I../action_management/include/
 RM = rm -f
 LIB = ui.a
-LDLIBS = -lncurses -lreadline
+LDLIBS = -lncursesw -lreadline
```
mpcsh: Same as above - is this a typo?

src/common/include/ctx.h
mpcsh: And then all the changes below adding the banner as a parameter should be backed out

src/ui/src/print_functions.c
mpcsh: Agreed - there are more below (`LINES / 4 - 7`). If they don't make sense as compile-time constants, then at least include a comment.

mpcsh:
@ecoble don't be afraid to request changes! You brought valid, valuable feedback that should be addressed before the PR is merged.

src/ui/src/print_functions.c
```diff
@@ -19,7 +19,28 @@
 /* see print_functions.h */
 void print_info(chiventure_ctx_t *ctx, window_t *win)
```
adomiter: I would add comments to this function as this would help the rest of the team understand how to we can print additional information to the screen (i.e. the list of actions under 'help' and the 'inventory').

Makefile
adomiter: I believe this is an extended version of the ncurses library. I wandering why you chose this library specifically and what is it helping achieve that the regular ncurses can't.

src/ui/src/print_functions.c
adomiter: I would add comments to this function as this would help the rest of the team understand how we can print additional information to the screen (i.e. the list of actions under 'help' and the 'inventory').

Makefile
sebvil: Yes, it is an extension of the ncurses library. It essentially allows us to print unicode characters  present in the chiventure banner. See #205 for more info.

src/ui/src/print_functions.c
sebvil: The list of actions under help is taken care of by the CLI team, and similar with the inventory (though I don't know that that is currently implemented yet)

src/chiventure.c
sebvil: I had it so that the banner would always be printed in the main screen, which is why I needed it in the ctx struct (so I would not need to modify the print functions). I changed it so that now it is only printed once in the beginning until a key is pressed.

mpcsh:
Issue Score: ✓+

Comments:
Great work!

Issue #250, opened 2019-06-01
Refactor AM modules to use chiventure context struct


MTUC08:
Refactor action management functions to use the chiventure context struct. This allows for simpler passing of arguments between group functions and is anticipated to be the norm for sprint 4. This includes testing functions as well.

Additionally, CLI is being added to this issue so that they can refactor operations.c with our new parameters.

SnowballAntrobus:
I can help with the testing functions.

MTUC08:
So my current line of logic is that since do_path_action is the only one that takes in/uses a game_t struct, it's the only one that needs to be refactored to use a chiventure context struct.

MTUC08:
Update: Just refactored do_path_action c and h to take in and handle the chiventure context struct instead of a game struct. This was done on branch actions/chiventure_ctx_t and after Dante refactors the corresponding tests, will be merged into actions/do_path_191 so it can go into actions/dev with issue #302 .

SnowballAntrobus:
I have added a member of CLI so that they can refactor their code to our new function parameters.

ecoble:
UPDATE: Changes CLI's use of action management's do_item_action, do_path_action, and do_item_item_action to not pass the game struct as a parameter (for do_item_action and do_item_item_action), or replacing the game struct parameter with a chiventure context struct parameter (for do_path_action)

SnowballAntrobus:
We had an issue with a circular dependency where ctx.h was calling cmd.h which was calling actionmanagement.h which was calling ctx.h. I think we have now resolved this with help from CLI.

SnowballAntrobus:
Ok we should be done with the refactor both of tests and actionmanagment.c/.h using the new chiventure context struct. The main issue was fixing makefiles because now we have to link with cli and ui since they are used in the struct. We will now be making a PR into actions/do_path_191.

MTUC08:
In this issue, we changed the game_t input parameter to the do_path function to the shared chiventure context struct chiventure_ctx_t and removed the game_t input parameter from the other do functions as they were not being used to actually implement the function. We then refactored the do_path_function and tests from issue #191 and #193 to reflect the new refactorization. CLI also refactored their function calls based on this change. Finally, we fixed the makefiles to link to the cli, ui, and common libraries to reflect the new chiventure context struct. 

[...]

PR #302, opened 2019-06-07
Implemented the do_path_action function and associated tests using game-state structs, as per #191 #193

yoonsjeong:
As part of issue #191, I've used the game-state functions to implement an actually functional do_path function. Here are a list of some of the features
- added the macro variable `#define NOT_ALLOWED_PATH (4)`
- checks to ensure that destination room and path exist, returns NOT_ALLOWED_PATH otherwise
- alters ret_string to print what error code we get if move_room() fails
- changes ret_string to the short description if move_room() succeeds
- ALTERS path function so that rather than a game_t struct, it uses a chiventure_ctx_t struct (the context struct)
- tests above path function to make sure that these work
- removed all instances of the game struct in the do_item and do_item_item functions, and all associated tests
Let me know if there are any questions or slightly weird bugs!


src/action_management/src/actionmanagement.c
```diff
@@ -95,21 +96,45 @@ int do_path_action(game_t *g, action_type_t *a, path_t *p, char **ret_string)
...
+
+    if (move == SUCCESS) {
+        sprintf(string, "Moved into %s. Short description: %s", 
+                room_dest->room_id, room_dest->short_desc);
+        *ret_string = string;
+        return SUCCESS;
+    }
+    else {
+        sprintf(string, 
+                "Requested action %s in direction %s into room %s, but given error code %d by move_room()",
```
MTUC08: "given error code %d by move_room()" is not useful to the game player. Maybe change it to something like "Unable to move to room %s", room_dest->room_id

src/action_management/src/actionmanagement.c
```diff
@@ -95,21 +96,45 @@ int do_path_action(game_t *g, action_type_t *a, path_t *p, char **ret_string)
...
+    if ((path_found == NULL) || (room_dest == NULL)) {
+        sprintf(string, "The path or room provided was invalid.");
+        return NOT_ALLOWED_PATH;
+    }
+
+    /* PERFORM ACTION */
+    int move = move_room(g, room_dest);
+
+    if (move == SUCCESS) {
+        sprintf(string, "Moved into %s. Short description: %s",
```
MTUC08: This string should be more user-readable since it is going to the player of the game. Maybe remove the "Short description: " part so it becomes "Moved into %s. %s", room_dest->room_id, room_dest->long_desc. (Long Desc because you just moved to a new room, so it should show the entire description.

src/action_management/src/actionmanagement.c
yoonsjeong: I don't think this string is meant to be seen by the player -- it's an error that only occurs if there's something wrong with the code itself, not with an action the player requested. This is since move = move_room(), and move_room() is only executed once everything player-based has been validated

MTUC08: But the functionality of our strings are that cli will return it to the user interface, so the error should be something that would be useful to the game player, even if its just an error message.

MTUC08: @yoonsjeong please resolve this before I approve the merge.

MTUC08:
Update: Currently switching the game_t input to a chiventure_ctx_t input as per issue #250 . We are doing this on a branch called actions/chiventure_ctx_t that will be merged with actions/do_path_191 as soon as Dante finishes refactoring the tests so it can go into actions/dev along with the other files in this PR.

src/action_management/src/actionmanagement.c
yoonsjeong: got it -- I've fixed it

MTUC08:
Approved! Please wait for the PR with the new chiventure_ctx_t format to go through first before merging to actions/dev

MTUC08:
Update: Just added the PR for the new format #321

mpcsh:
Issue Score: ✓+

Comments:
Great work!

PR #305, opened 2019-06-07
Adding basic end conditions to game

patrickkoopmans:
This pull request to merge a small collaborative feature between WDL and Game State to add an end condition to games. 

As specified by WDL, the end condition will consist of a final_room specified by the developer. move_room() in the game module has been changed to return a specific value when the player moves to the final room. The game struct has been updated to include a final_room field. A function add_final_room_to_game() has been written to add a final room to the game and protect against bad inputs. All relevant documentation has been duly updated. 

Although this is a small feature, I am aiming to get this approved quickly so there is some way to end games before the demo next week.

### DIFF
--- a/docs/GAME-STATE.md
+++ b/docs/GAME-STATE.md
@@ -59,6 +59,9 @@ There is also a time_start field, but it does not have specific functionalily ri
         /* pointer to current room struct */
         room_t *curr_room;
 
+        /* pointer to room that, when entered, ends the game */
+        room_t *final_room;
+
         /* pointer to current player struct */
         room_t *curr_player;
 
@@ -85,7 +88,6 @@ This functions mallocs space for a new game struct. It does not take any argumen
 
  Frees everything in the game struct safely.
 
-
     void game_quit(game_t *game);
 + Parameters:
      - game struct
@@ -94,14 +96,27 @@ This functions mallocs space for a new game struct. It does not take any argumen
 
 Exits game safely (frees all memory before exits).
 
-    void move_to_room(game_t *game, room_t *new_room);
+    void move_room(game_t *game, room_t *new_room);
 + Parameters:
      - game struct
-     - room that we're changing to
+     - pointer to room that we're changing to
 + Returns:
-     - none
+     - 0 if successful, 1 if failed, 2 if game NULL, 3 if new_room NULL, 4 if new_room is final room
 
-Given the pointer to another room, changes the current room in game-state. This function does NOT check if the move is legal.
+Given the pointer to another room, changes the current room in game-state. 
+
+int add_final_room_to_game(game_t *game, room_t *final_room)
++  Parameters:
+    - game struct
+    - pointer to final room
++ Returns: 
+    - SUCCESS if successful, FAILURE if failed
+
+ Adds the final room to the given game
+
+
+
+
 
 ## The Player Module
 
--- a/docs/GAME-STATE.md
+++ b/docs/GAME-STATE.md
@@ -112,7 +112,7 @@ Given the pointer to another room, changes the current room in game-state.
 + Returns: 
     - SUCCESS if successful, FAILURE if failed
 
- Adds the final room to the given game
+ Adds the final_room field to the given game struct
 
 
 
--- a/docs/GAME-STATE.md
+++ b/docs/GAME-STATE.md
@@ -105,7 +105,7 @@ Exits game safely (frees all memory before exits).
 
 Given the pointer to another room, changes the current room in game-state. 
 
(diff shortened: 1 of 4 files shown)


src/game-state/include/game.h
```diff
@@ -110,6 +114,17 @@ int add_player_to_game(game_t *game, player_t *player);
  */
 int add_room_to_game(game_t *game, room_t *room);
 
+/* Adds the final room to the given game
+ *
+ * Parameters:
+ *  game struct
```
lydiafilipe: Note that these are pointers not just structs. It looks like this is a problem in a lot of the documentation in this file, please change all the function comments that have this.

src/game-state/src/game.c
```diff
@@ -55,6 +55,22 @@ int add_room_to_game(game_t *game, room_t *room) {
     return SUCCESS;
 }
 
+/* See game.h */
+int add_final_room_to_game(game_t *game, room_t *final_room) {
+    room_t *check;
+    HASH_FIND(hh, game->all_rooms, final_room->room_id, strnlen(final_room->room_id, MAX_ID_LEN),
```
lydiafilipe: Some of these lines are too long, please make sure they don't exceed 80 characters

src/game-state/src/game.c
```diff
@@ -103,6 +119,10 @@ int move_room(game_t *game, room_t *new_room) {
     }
     if(new_room == NULL)
         return 3;
+    if(new_room == game->final_room)
+    {
+        return 4;
```
lydiafilipe: See comment in game.h about this

src/game-state/include/game.h
```diff
@@ -63,6 +66,7 @@ game_t *game_new(char *start_desc);
  *  1 for failure
  *  2 if game null
  *  3 if new_room is null
+ *  4 if new_room is the game's final_room
```
lydiafilipe: Off of Lindsey's comment, it might actually be good to have meaningful macros for all of these in addition to SUCCESS and FAILURE. I believe in the large pr I let this go but it's important to add for clarity

lydiafilipe:
@lindsey-hsi generally it is good to leave comments in the code directly rather than as a comment on the entire issue. It's easier to pinpoint and have direct discussion on

docs/GAME-STATE.md
```diff
@@ -94,14 +96,27 @@ This functions mallocs space for a new game struct. It does not take any argumen
...
+Given the pointer to another room, changes the current room in game-state. 
+
+    int add_final_room_to_game(game_t *game, room_t *final_room)
++  Parameters:
+    - game struct
+    - pointer to final room
++ Returns: 
+    - SUCCESS if successful, FAILURE if failed
+
+ Adds the final room to the given game
```
thoranderskow: I think this comment could be more clear. Is there a way to make it clear that final room is referring to final_room within the game_t struct?

patrickkoopmans:
> Everything looks great for the most part! There are just a few small errors I saw, mostly in the documentation:
> 
> * move_room in game.h: the header comment reads that the function returns 0 for success and 1 for failure (alongside the other ints that specify different results), but the actual function returns 0 for FAILURE and 1 for SUCCESS, which is consistent with the conventions we’ve been following.
> * add_final_room_to_game in game.h: shouldn’t the error message read “this room already exists in the list of rooms”, since the condition is if check != NULL?
> * game-state.md: the same inconsistency applies as in the documentation for move_room in game.h — the description for move_room should read that it returns 0 for failure, 1 for SUCCESS, etc.
> 
> Other than those three things, your work looks great - it's well-written + concise, and the code and documentation follow the stylistic conventions we've been using.

Thanks for the feedback! 

1. Regarding your first point, I have changed the functions to return the correct SUCCESS and FAILURE macros
2. Actually, this check is to ensure that the room that WDL passes to load into the final_room() pointer in the game struct is an actual room in the game. This error message is to prevent WDL from making the final room a room that does not exist for the player to enter
3. Noted and changed!

src/game-state/src/game.c
```diff
@@ -99,10 +120,14 @@ room_t *find_room_from_game(game_t *game, char* room_id) {
 /* See game.h */
 int move_room(game_t *game, room_t *new_room) {
     if(game == NULL) {
-        return 2;
+        return GAME_NULL;
     }
     if(new_room == NULL)
```
lydiafilipe: very small thing but you should have braces in this if-statement. Also, it looks like you're using k&r style but below is allman, please make sure this is consistent

lydiafilipe:
Issue Score: ✓

Comments:
Nice work on this pr! However there is no closing statement here.

Issue #310, opened 2019-06-08
Catching errors when game pointer not initialized


Gestalt-R:
Currently, the functions will segfault if the game pointer isn't initialized. In normal operations, they should be initialized, but this helps for bugfixing.

I hope to add code to each operation that accesses a game to output an error message if it isn't initialized.

Gestalt-R:
This branch added in all the necessary code that was needed to output the errors in operations.c.

I consider the issue resolved with the merge.

lydiafilipe:
Issue Score: ✓

Comments:
Nice job! Generally it's good to have more detailed closing statements and updates but since this is a small issue this is okay. However it would have been nice as part of the closing statement or an update to specify a bit more exactly what you did for the fix.

PR #313, opened 2019-06-08
Catching errors occuring when game doesn't exist

Gestalt-R:
Issue #310. This is meant to give informative error messages when no game has yet been loaded. Essentially, just adding in tests for when there is no game to access, and error messages to occur instead of segfaults.

Files have been tweaked in operations.c in order to insert the messages.

### DIFF
--- a/src/cli/src/operations.c
+++ b/src/cli/src/operations.c
@@ -43,7 +43,8 @@ char *save_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 char *look_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 {
     game_t *game = ctx->game;
-    if(game == NULL || game->curr_room == NULL) return "Room not found! Error! We need a room to be loaded to LOOK!\n";
+    if(game == NULL || game->curr_room == NULL)
+        return "Room not found! Error! We need a room to be loaded to LOOK!\n";
     if(tokens[1] == NULL)
     {
         return game->curr_room->long_desc;
--- a/src/cli/src/operations.c
+++ b/src/cli/src/operations.c
@@ -44,7 +44,9 @@ char *look_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 {
     game_t *game = ctx->game;
     if(game == NULL || game->curr_room == NULL)
+    {
         return "Room not found! Error! We need a room to be loaded to LOOK!\n";
+    }
     if(tokens[1] == NULL)
     {
         return game->curr_room->long_desc;
@@ -98,7 +100,7 @@ char *kind2_action_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ct
     if(game == NULL || game->curr_room == NULL)
     {
         print_to_cli(ctx, tokens[0]);
-        return ( "Error! We need a loaded room to do the above action. \n");
+        return "Error! We need a loaded room to do the above action. \n";
     }
     lookup_t **table = ctx->table;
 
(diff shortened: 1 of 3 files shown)


Gestalt-R:
Oh, by the way. The changes look a lot bigger than they are, because astyle found a couple of empty lines to delete. It's really very small.

src/cli/src/operations.c
```diff
@@ -43,7 +43,7 @@ char *save_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 char *look_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 {
     game_t *game = ctx->game;
-
+    if(game == NULL || game->curr_room == NULL) return "Room not found! Error! We need a room to be loaded to LOOK!\n";
```
a-matai: Is it better style to add the string in parens

src/cli/src/operations.c
```diff
@@ -111,6 +121,11 @@ char *kind2_action_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ct
 char *kind3_action_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 {
     game_t *game = ctx->game;
+    if(game == NULL || game->curr_room == NULL)
+    {
+        print_to_cli(ctx, tokens[0]);
+        return ( "Error! We need a loaded room to do the above action. \n");
```
a-matai: Instead of repeating this code 3x could you write a function that you could just call that checks if game/curr_room is null? Makes it easier in case you change things later on.

src/cli/src/operations.c
ecoble: Please make this in the form
if(...)
{
    .....
}

Not a big deal but I think Lydia has requested this change from us before

src/cli/src/operations.c
Gestalt-R: It would be kinda weird to do that, because it terminates the running of the function there.

src/cli/src/operations.c
Gestalt-R: We didn't elsewhere, so I actually changed it for this spec.

Gestalt-R: Done!

lydiafilipe:
Please give this pr a more informative title

src/cli/src/operations.c
```diff
@@ -43,7 +43,8 @@ char *save_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 char *look_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 {
     game_t *game = ctx->game;
-
+    if(game == NULL || game->curr_room == NULL)
```
lydiafilipe: Please uses braces around if statements

src/cli/src/operations.c
```diff
@@ -63,6 +64,11 @@ char *look_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 char *kind1_action_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 {
     game_t *game = ctx->game;
+    if(game == NULL || game->curr_room == NULL)
+    {
+        print_to_cli(ctx, tokens[0]);
+        return ( "Error! We need a loaded room to do the above action. \n");
```
lydiafilipe: Why parentheses here? Not sure they're necessary. Same below

src/cli/src/operations.c
lydiafilipe: I don't think an additional function is necessary

lydiafilipe:
Issue Score: ✓

Comments:
Nice work! However there is no closing statement.

PR #321, opened 2019-06-08
Changed game_t parameter to chiventure_ctx_t as per #250 in AM functions

MTUC08:
We changed the game_t parameter to chiventure_ctx_t within the do_path_action function, which is the only one that uses game_t, and removed the parameter from the other do functions which don't use it. This change was also reflected in our changes to the path test function. Finally, we refactored our tests makefile to reflect the new chiventure_ctx_t input.


src/action_management/src/actionmanagement.c
```diff
@@ -91,36 +89,60 @@ int do_item_action(game_t *g, action_type_t *a, item_t *i, char **ret_string)
...
+        *ret_string = string;
+        return NOT_ALLOWED_PATH;
+    }
 }
 
 
 /* KIND 3
  * See actionmanagement.h */
-int do_item_item_action(game_t *g, action_type_t *a, item_t *direct,
+int do_item_item_action(action_type_t *a, item_t *direct,
```
yoonsjeong: I know currently both the do_item_item_actions and do_item_actions funcs are being worked on, but how do they work without a context variable? I assume for now there's some work around, but is it possible to change the game state without having the game as an input?

MTUC08: yes b/c the way they work is that conditions and effects are all in the passed in item functions so they just change those. It's more changing items within the game than the game itself

MTUC08:
Closing Comments:
In this PR we changed the game_t input in do_path to a chiventure context struct and removed the game_t input in the other two do_path functions as we found that they were extraneous. The changes will be carried into master with the path function and tests #325 .

mpcsh:
Issue Score: ✓+

Comments:
Great work!

PR #326, opened 2019-06-09
Tests for lookup_t struct, cmd struct, and parser, general cleanup and updating

ecoble:
This PR includes:

- Tests lookup_t_new, lookup_t_init, and lookup_t_free
- Tests for cmd_new, cmd_init, and cmd_free
- Tests for the parser with no input, 1, 2, 3, and 4 word input, invalid input (more than 4 words), and input with a lot of spaces around each word
- Refactoring: Removed the validate module and moved the two functions from validate.c and validate.h into operations.c and operations.h
- Changed lookup_t_free and cmd_free to return a success int instead of void
- Updated operation functions to use game state's get_item_in_room function
- Small bug fixes


src/cli/include/cmd.h
```diff
@@ -167,9 +167,9 @@ int cmd_init(cmd *c, char *tokens[TOKEN_LIST_SIZE]);
  * - pointer to a cmd struct
  *
  * Returns:
- * - nothing
+ * - always returns 0
  */
-void cmd_free(cmd *c);
+int cmd_free(cmd *c);
```
Gestalt-R: Why is it type int if it never returns 1?

src/cli/src/operations.c
```diff
@@ -4,8 +4,7 @@
 #include "operations.h"
 #include "shell.h"
 #include "assert.h"
```
dbarer: I don't think there's any cost to having this in. But I don't think anything in this file calls on the assert library.

src/cli/tests/Makefile
```diff
@@ -0,0 +1,32 @@
+CC = gcc
```
dbarer: It might make sense to run this Makefile when you run make tests in our cli directories makefile, or even do all the building tests from that Makefile. Though this is just a suggestion, and this obviously works as is as well, since the games users aren't going to be touching these tests.

lydiafilipe:
@Gestalt-R make sure that you request changes rather than just comment when you review

Gestalt-R:
@lydiafilipe I wanted help understanding the code before I made any attempts at review. 
Because I didn't think it needed changing necessarily, and looked intentional, I just wanted to know why it had been changed.

lydiafilipe:
It looks like there are merge conflicts, please pull from master and sort that out

src/cli/include/cmd.h
ecoble: I believe returning the 0 is the standard, if the free call were to fail it would just crash and we wouldn't get the success int back

src/cli/src/operations.c
```diff
@@ -4,8 +4,7 @@
...
-#include "validate.h"
-
+#include "room.h"
 
 // remove the comment as soon as checkpointing removes their dummy struct
```
lydiafilipe: Since this is all commented out now you should remove

src/cli/src/operations.c
```diff
@@ -179,10 +217,3 @@ char *switch_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
     layout_switch(ctx);
     return "Layout switched.";
 }
-
-//Because action managment does not support NPCs type 4 is not supported
-//char *kind4_action_operation(char *tokens[TOKEN_LIST_SIZE], game_t *game)
-//{
-//    printf("%s\n",tokens[0] );
-//    return "is an action!";
-//}
```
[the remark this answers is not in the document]
ecoble: I am not sure, maybe @lydiafilipe could help answer this. We have this commented code and something we were originally saving to support NPCs, it seems unlikely this will happen before the project is over but next year's project would probably work on this, should we leave the commented code or delete it?

src/cli/src/operations.c
```diff
@@ -28,6 +27,46 @@ char *hist_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
+ * - A string of the filename
+ *
+ *
+ * Returns:
+ * - TRUE if proper filename, FALSE if not
+ */
+bool validate_filename(char *filename)
+{
+    int len = strlen(filename);
+    if(len < 4)
```
lydiafilipe: What is the significance of 4 here? It is a bit like a magic number (see style guide for discussion of those). I would assign a meaningful variable name or macro for what it is

src/cli/src/operations.c
```diff
@@ -28,6 +27,46 @@ char *hist_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
 
+/*
+ * Checks that a specified filetype is a .dat file
+ *
+ * Parameters:
+ * - A string of the filename
+ *
+ *
+ * Returns:
+ * - TRUE if proper filename, FALSE if not
```
lydiafilipe: true and false are lowercase

src/cli/src/operations.c
```diff
@@ -28,6 +27,46 @@ char *hist_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
+        return false;
+    }
+}
+
+/* See operation.h */
+cmd *assign_action(char **ts, lookup_t ** table)
+{
+    cmd *output = cmd_new(ts);
+    output->func_of_cmd = find_operation(ts[0], table);
+    if(output->func_of_cmd == NULL) output->func_of_cmd = action_error_operation;
```
lydiafilipe: Please use braces in an if statement and put the next expression on the next line

src/cli/tests/Makefile
ecoble: I think we can leave the test makefile separate, that seems to be what other teams are doing and I don't see a reason to compile our tests when the whole chiventure program is being compiled

src/cli/src/operations.c
```diff
@@ -66,22 +107,21 @@ char *kind1_action_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ct
     lookup_t **table = ctx->table;
 
     if(tokens[1] == NULL)
-    {
-        return "You must identify an object to act on\n";
-    }
+        {
```
lydiafilipe: The spacing here is inconsistent, there shouldn't be indentation before the braces. Same with the if statement on 115

src/cli/src/operations.c
```diff
@@ -161,13 +189,23 @@ char *inventory_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
+        char str[10];
+        sprintf(str, "%d", i);
+        print_to_cli(ctx, str);
         print_to_cli(ctx, t->item_id);
         print_to_cli(ctx, t->short_desc);
     }
     return "This was your inventory";
 }
 
+//Because action managment does not support NPCs type 4 is not supported
```
lydiafilipe: I would actually delete this all together for now, it's not good to have commented out code in your files

ecoble:
@lydiafilipe Fixed the merge conflicts

src/cli/include/cmd.h
```diff
@@ -134,9 +134,9 @@ int lookup_t_init();
  * - pointer to a lookup hashtable
  *
  * Returns:
- * - nothing
+ * - always returns 0
```
lydiafilipe: To be consistent with the init function, it would be better to return SUCCESS in this and cmd_free rather than 0

src/cli/src/cmd.c
```diff
@@ -37,8 +36,8 @@ void add_action_entries(lookup_t **table)
             add_entry(curr_action->c_name, kind3_action_operation, table);
         }
 
-        all_actions = all_actions->next;
-    }
+         all_actions = all_actions->next;
```
lydiafilipe: why are these extra spaces being added?

src/cli/src/operations.c
```diff
@@ -48,17 +86,19 @@ char *look_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
-        return game->curr_room->long_desc;
-    }
-    item_t *curr_item;
-    ITER_ALL_ITEMS_IN_ROOM(game->curr_room, curr_item)
-    {
-        if (strcmp(curr_item->item_id,tokens[1]) == 0)
-        {
-            return curr_item->long_desc;
+    {   
+        if(game !=NULL){
```
SnowballAntrobus:
```suggestion
        if(game != NULL){
```

src/cli/src/operations.c
```diff
@@ -28,6 +26,46 @@ char *hist_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
+{
+    int len = strlen(filename);
+    if(len < 4)
+    {
+        return false;
+    }
+    const char *ending = &filename[len-4];
+    int cmp = strcmp(ending, ".dat");
+    if(cmp == 0)
+    {
```
lydiafilipe: I've noticed the braces style is a bit inconsistent please stick to a style (it looks like you've been using allman so make sure everything uses that) - please fix throughout the file

src/cli/src/operations.c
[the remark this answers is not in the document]
lydiafilipe: See my comment about commented-out code. Also, since stuff is under version control, if you ever want to look at what was here you can just look at an old commit

src/cli/tests/test_parser.c
```diff
@@ -0,0 +1,78 @@
...
+    cr_assert_str_eq(words[2], "ME", "parse() did not create third token");
+    cr_assert_str_eq(words[3], "NOT", "parse() did not create fourth token");
+}
+
+//Tests the parsing of five words. (Having more words than 4 causes the parser to return NULL)
+Test(parse, more_words)
+{
+    char str[] = "LOOK AT ME NOT YEAH";
+    char **words = parse(str);
+    cr_assert_null(words, "parser() should return NULL if too many words");
```
SnowballAntrobus:
```suggestion
    cr_assert_null(words, "parser() should return NULL if more than 4 words");
```

src/cli/src/operations.c
```diff
@@ -183,13 +210,23 @@ char *inventory_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
     ITER_ALL_ITEMS_IN_INVENTORY(game->curr_player, t)
     {
         i++;
-        print_to_cli(ctx, i);
+
+        char str[10];
```
lydiafilipe: 10 here is also bit like a magic number, please see my other comment about these. An explanatory comment could also be a good idea

ecoble:
CLOSING STATEMENT: There are now tests for the lookup_t struct, cmd struct, and parser module. There has also been refactoring of the validate module into the operations module, and general bug fixes and clean up

lydiafilipe:
Issue Score: ✓+

Comments:
Great work!

PR #339, opened 2019-06-10
merge wdl/load_wdl into wdl/dev to prepare for PR to master

thoranderskow:
This PR contains the module load_wdl.h/load_wdl.c and test_wdl.c. This function is meant to load a wdl into all the game-state structs required for a game.


src/wdl/include/load_game.h
```diff
@@ -5,13 +5,10 @@
...
 
-#include <stdio.h>
-#include <stdlib.h>
-#include <stdbool.h>
+#include "wdl_common.h"
 #include "parse.h"
-#include "room.h"
 
-#define FILE_PATH "../examples/sample_yaml/connected_rooms.yaml"
+#define FILE_PATH "../examples/sample_yaml/simplified_wdl.yaml"
```
victor-a-a: This is okay for now.  However, in the future it might have to be changed to the more comprehensive testing file.

src/wdl/include/load_wdl.h
```diff
@@ -0,0 +1,22 @@
+#ifndef INCLUDE_LOAD_WDL_H
+#define INCLUDE_LOAD_WDL_H
+
+#include "load_item.h"
+#include "load_room.h"
+#include "load_game.h"
+#include "wdl_common.h"
+
+//file path that specifies which yaml file to use
+#define PATH_TO_YAML "../examples/sample_yaml/simplified_wdl.yaml"
```
victor-a-a: Is this a global variable?  If so, I don't think we should be using one.

src/wdl/src/load_wdl.c
```diff
@@ -0,0 +1,33 @@
...
+/*
+ * load_wdl, function that loads a wdl into all the game-state structs
+ *
+ * has no parameters, automatically load yaml found in FILE_PATH
+ *
+ *
+ */
+game_t *load_wdl()
+{
+  obj_t *big_document = get_doc_obj(PATH_TO_YAML);
```
victor-a-a: How will the user specify what file to load?

src/wdl/include/load_game.h
thoranderskow: In order to get the demo out as quickly as possible, I decided it was easiest for the other teams if they didn't have to worry about choosing a yaml file to load, thus we are using a macro. Obviously this will have to be changed in the future, but right now this seems to work.

src/wdl/include/load_wdl.h
thoranderskow: see above comment; once more this will likely be changed.

src/wdl/src/load_wdl.c
thoranderskow: see above comments; Likely the function will take in a file path at some point to specify the yaml file.

lydiafilipe:
Issue Score: ✓

Comments:
Nice work! However there is no closing statement.

PR #311, opened 2019-06-08
Examples and updated saving/loading 

jonahdf:
The long awaited PR to master. 
- 100% passing tests
- Fully updated game state structs
- Integrated CLI files
- Hash table integration
- Selective saving/loading (only mutable structs)

Warning: this is a LARGE pull request, since we could not figure out how to merge smaller changes at a time (nearly all our code is dependent on the rest). This might take a while to review. There are also indeed merge conflicts that we must work out, but I figure it is better to get this beast started now and catch the problems in the reviewing. I have made @dbarer the reviewer from another team because he has helped integrate some CLI changes and has probably gotten more familiar with our code.

Also worth noting that top-level makefile does not compile as of now, but checkpointing, CLI, and common all do.

PR #343, opened 2019-06-10
Checkpointing/make

jarvislam1999:
Fixed Makefile to compile for chiventure. Only solution we could have thought of. Typing make in chiventure compiles now. Yay?


jonahdf:
I resolved the merge conflicts with Ryan

mpcsh:
@jarvislam1999 can you please back out the style changes to CLI's code? We should avoid touching code that isn't ours, even if it does need a style update.

Gestalt-R:
Yeah, our code uses the allman style. Please don't change it without asking.

mpcsh:
@Gestalt-R the issue at hand was not the choice of Allman style but rather things like `while(condition)` instead of `while (condition)` (the former not being acceptable regardless of brace style). Regardless, the point was to not generate a diff unless necessary.

Gestalt-R:
Regardless, the style change was also incorporated into the request.

jonahdf:
@mpcsh I reverted the style commit. Are we good to merge now?

mpcsh:
@jonahdf it looks like the `master` merge you did here picked up commits that `dev` doesn't have - can you merge `master` into `dev` again?

mpcsh:
@jonahdf actually you'll have to merge `master` into both `dev` and this branch, `master` just saw yet another merge

jonahdf:
I tried merging master, it got quite complex, partially because of David's changes to dev and lots of merge conflicts that I had to resolve in both dev and make. I decided it would be simpler to just make the changes in dev, and include that in the #311 (and close this PR). Everything now compiles in dev!

PR #348, opened 2019-06-11
A FEW quick UI Fixes

omorkved:
I made a few **minor** fixes -- This is a small PR, let's get it into master!

1. Actual hotfix: `create_valid_map()` was sometimes returning the int value `FAILURE` instead of `NULL`, which created a warning in compiletime, so i fixed that 

2 - Added a call to `chiventure_ctx_free` in map_example.c

3 - fixed a comment in DFS.c file 

4 - Linking pieces together: As we've been waiting to do, finally added in call to `create_valid_map()` in `ui_ctx_init()` ! This means we get a chance to try and actually create...wait for it.... a _valid_ map. 
 
Currently in master, we just print a random four-room test map. Now in my update, crucially, if `create_valid_map `fails at any point, it just automatically returns NULL back to `ui_ctx_init()`, and `get_test_rooms()` is then still called upon. This means that:
a) we won't seg fault/throw an error if our coordinate-assigning fails
 b) while not ideal to print a test map in case of failure to assign unique coordinates, this is a big step up from what is currently in master (which is just printing the test map 100% of the time)

5 - changed an assert in `create_valid_map` to reutrn `NULL` to avoid future seg faults caused by 4 above (so if any other team tests with a NULL game_t, our function won't throw errors, it will just print a test map)


omorkved:
Recent merge to master just broke my build -- working on fix

omorkved:
@mpcsh Merged master fix, builds agian. Ready to merge!

mpcsh:
Issue Score: ✓+

Comments:
Great work!

PR #349, opened 2019-06-11
Added home screen animation

sebvil:
As the title indicates, I added a small animation to the home screen. Essentially the banner will flash for a couple seconds before the user can put any input.


src/ui/src/print_functions.c
```diff
@@ -28,23 +32,55 @@ void print_homescreen(window_t *win, const char *banner)
...
+
+        usleep(100 * 1000);
+
+        if (i > 4) {
+            usleep(500 * 1000);
+        }
+
+        int len = strlen(banner);
+        char temp[len];
+        strcpy(temp, banner);
```
patrickkoopmans: Because you have defined BANNER_LENGTH you can use strncpy() here (safe string copy) which will allow you to specify the max size of the banner to be copied. This might help in case the banner is changed from whatever generic design it is right now and becomes too long to fit the UI.

src/ui/src/print_functions.c
```diff
@@ -28,23 +32,55 @@ void print_homescreen(window_t *win, const char *banner)
...
+        int x = x_pos;
+        int y = y_pos;
+        wclear(win->w);
+        box(win->w, 0, 0);
+
+        wrefresh(win->w);
+
+        usleep(100 * 1000);
+
+        if (i > 4) {
```
patrickkoopmans: A small suggestion: I think that it is more intuitive to have an if, else control structure here because you are really choosing between two options depending on the iterator value. I think that this would also make clear what you mean by "the last couple a bit slower" in the comment above,

mpcsh:
@patrickkoopmans please do not be afraid to request changes, because it is now not clear if we should be waiting on these changes to be addressed before merging. Should I merge this PR now, or should @sebvil make the changes you noted?

src/ui/src/print_functions.c
sebvil: Actually, in BANNER_LENGTH is the width of the banner (I'm going to change it rn to BANNER_WIDTH), not the actual length of the banner string.

mpcsh:
Issue Score: ✓+

Comments:
Great work!

PR #351, opened 2019-06-11
Update Helptext

ecoble:
This is a small PR to update the help text. At the request of UI, we have include "MAP" and "SWITCH" in the help text under "SYSTEM ACTIONS," so the user knows these commands are available to them. The help text also now includes a list of all the available action commands, as listed in action management's valid_actions array.

The list of available action commands is not dynamically generated from action management's array, but supporting that in the future could be useful especially as the list of valid actions grows.


ecoble:
CLOSING STATEMENT: "Map" and "Switch" are now listed in the help text, as well as a list of all available actions from action management. In the future it would be nice if we could automatically generate this list of actions using action management's array.

lydiafilipe:
Issue Score: ✓+

Comments:
Nice work!

PR #352, opened 2019-06-11
Map Upgrades: Draws Player, Shows Current Location, Map Has Border.

Sshatzkin:
This PR involved the creation of 3 new functions and some altering of map_refresh.

Changes: 
1. draw_info is an internal function that draws the player's current coordinate to the top left corner of the map.
2. draw_border is an internal function that draws a border onto the map screen.
3. set_player_loc sets the values stored in the player_loc coordinate struct in the UI context. This stored coordinate is what draw_info uses when printing current location to the screen. This ensures that the player's position on the map screen always matches the displayed coordinates.
4. I added a few lines to refresh_map that allows for the drawing of the player in the middle of the screen.
5. Added more rooms to get_test_coord_hash to test more features and better demonstrate the map module.


src/ui/src/map.c
```diff
@@ -146,16 +146,71 @@ map_t *map_init()
...
 }
 
+/* See map.h for documentation */
 int map_set_displaywin(map_t *map, int ulx, int uly, int lrx, int lry)
 {
     map->ulx = ulx;
     map->uly = uly;
     map->lrx = lrx;
-    map->lry = lry;
+    map->lry = lry-1;
```
mpcsh: Spaces here

src/ui/examples/map_example.c
```diff
@@ -67,7 +43,7 @@ int main()
     /* This example tests without a loaded game stuct, so
      * the coord_hash is set manually.
      */
-    ctx->ui_ctx->coord_hash = coordmap;
+    //ctx->ui_ctx->coord_hash = coordmap;
```
mpcsh: Space after `//`

src/ui/src/coordinate.c
```diff
@@ -93,6 +93,11 @@ coord_record_t *get_test_coord_hash()
     room_t *room2 = room_new("Two", "", "");
     room_t *room3 = room_new("Three", "", "");
     room_t *room4 = room_new("Four", "", "");
+    room_t *room5 = room_new("Five", "", "");
```
mpcsh: What is this function doing in `coordinate.c`? Isn't this a test fixture?

mpcsh:
Also, it looks like the build is failing, which might be related to #354?

omorkved:
> Also, it looks like the build is failing, which might be related to #354?

Yeah sad, it JUST started failing in my PR too--was passing an hour ago. we're trying to figure out why the new merge to master no longer can find ui_ctx.h @mpcsh

src/ui/include/ui_ctx.h
```diff
@@ -9,6 +9,8 @@
 
 //Forward Declaration
 typedef struct map map_t;
+typedef struct coord coord_t;
```
omorkved: Why was this being forward declared? was there a makefile problem?

src/ui/examples/map_example.c
omorkved: Why is this commented out?

Sshatzkin: We actually don't need this because we now handle this in ui_ctx_init(). I actually just removed all these comments and the unnecessary code.

omorkved: It still builds?

src/ui/include/ui_ctx.h
Sshatzkin: Yes. I think it was caused by the fact that I added the function set_player_pos() to coordinate.h because it deals with the coordinate struct in the ui_ctx. It probably could've gone in a different file but it seemed like it fit there. This required coordinate.h to include ctx.h and created a little build problem that was quickly solved with forward declaration. It definitely made me a little nervous to be messing with all that again but everything builds and runs fine.

src/ui/src/coordinate.c
Sshatzkin: It's a "test" function but it simulates the output of DFS and is currently being used in ui_ctx init when there has been no game initialized. It allows us to draw a map even when there is no real game available and avoid the segfaults associated with not having any kind of hash map. 

If we should move it, where do you recommend moving it to?

omorkved:
> Looks good to me! Will you be removing all the samples rooms if we are able to load a real game?

Someday! They could be moved at some point to a different file but not deleted entirely -- Ideally they would be used only to generate our map_example executable (which displays an example of the map visual without need for loaded info from a sample WDL/game)

src/ui/src/coordinate.c
mpcsh: Hmm. When there is no game initialized, can we just initialize an empty map, to avoid confusion?

Sshatzkin:
@mpcsh  I've made the requested changes. I think it looks pretty good for re-review.

mpcsh:
@Sshatzkin there's just one unresolved comment above

mpcsh:
@Sshatzkin never mind - looks like I wasn't caught up on what's been happening with other teams

mpcsh:
Issue Score: ✓+

Comments:
Great work!

PR #358, opened 2019-06-11
Updated Makefiles to avoid generating .d files when running 'make clean'

borjasotomayor:
One of the pervasive issues with the Makefiles is that "make clean" actually results in invoking gcc to generate dependency files, instead of just calling "rm". This patch should address that issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,15 @@ language: c
 
 compiler: gcc
 
+before_script:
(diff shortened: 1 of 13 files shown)


mpcsh:
FYI @lydiafilipe @borjasotomayor the build looks to be stuck but if you click through to Travis, it's passed

PR #353, opened 2019-06-11
Created sample game struct using gamestate and actionmanagement functions

Tianlel:
We created a simple two-room (room1, room2) sample game struct with 1 player and 2 items (item_orb and item_table) and updated the Makefile for common.a.

room1 is to the left of room2. 
item_orb is placed in room1 and supports actions "TAKE", "DROP", and "PUT_ON".
item_table is placed in room2 and supports action "PUT_ON".

This game struct sample can be used by cli, action_management, and checkpointing for integration tests.

PR #335, opened 2019-06-10
Supporting action effects and conditions in item module

jessxia:
This pull request includes all code from issues to support the addition of actions into the item struct. 

Issue #215: Creating the structs for game_action_t, game_action_condition_t, and game_action_effect_t and functions for adding/getting/checking possible actions for an item.

Issue #315: Developing functions for the game_action_condition_t struct, including functions to create new conditions, add them to items, and check if conditions are met

Issue #316: Developing functions for the game_action_effect_t struct, including functions to create new effects, add them to items, and carry out effects

Issue #331: Adding a hash table of items into the game struct and adding functions to add and retrieve items from the hash table. 

We had to combine 331 in a PR with 316 and 315 since the latter two rely on the item hash table implemented in issue 331. 315 and 316 were created to split the work started in 215.

PR #359, opened 2019-06-11
Correcting inputs in add_action call in sample_game.c

jessxia:
This PR is to correct the add_action calls in the most recent action-management PR #353 (that I approved so this is my bad). Add_action, which is not yet in master, no longer takes in the action_type_t struct.

These changes should allow the #335 game-state PR to build.


jessxia:
Closing comments: Game-state PR #335 now builds (yay!). Sample_game.c is correct, with add_action not taking in an action_type struct.

lydiafilipe:
Issue Score: ✓+

Comments:
Nice work!

PR #363, opened 2019-06-12
FIX: Printing to CLI

ecoble:
This includes a fix for printing, previously trying to call print_to_cli more than once between input prompts would cause only the last output to show, because it was overwriting previous input. This PR fixes this by moving the cursor down after each output, instead of at the beginning of each input. This also includes a small change of printing action management's success strings in action operation functions, instead of a generic success message that was being used as a placeholder.


src/ui/src/print_functions.c
```diff
@@ -122,8 +121,8 @@ void print_to_cli(chiventure_ctx_t *ctx, char *str)
     char temp[len];
     strcpy(temp, str);
     char *tmp = strtok(temp, "\n");
-    
-    
+
+
```
mpcsh: Can we delete this extra newline

mpcsh:
LGTM but @ecoble please resolve the merge conflict!

ecoble:
CLOSING STATEMENT: This PR fixes an issue were multiple calls to print_to_cli would overwrite the previous one, so only the last call showed up. This has been fixed so all lines of text appear.

Note: This introduced a bug with displaying the help text, which is being fixed on PR #375

lydiafilipe:
Issue Score: ✓+

Comments:
Nice work!

Issue #276, opened 2019-06-04
Clean up dependencies in header files and sanitizes the makefiles


Tianlel:
Remove unnecessary includes and remove -Wl,--start-group/-Wl,--end-group  in tests/Makefile by correcting the linking order of the static libraries

Tianlel:
Closing statement:
I cleaned up the header file dependencies and corrected the linking order of static libraries in tests/Makefile

lydiafilipe:
Issue Score: ✓+

Comments:
Nice work!

PR #367, opened 2019-06-12
Implemented hist command 

Tianlel:
I re-implemented the history (HIST) command using the readline library. 
Now the command prints out the history of all user input. In the future, we'd like to only print out a list of valid user input.
I added a line `using_history();` to ui.c to enable the use of the readline history functions.

The changes in src/cli/tests/Makefile and includes are a result of #276, which has not been made into master yet.


mpcsh:
This, unfortunately, is just cutting it too close. We can only accept fixes at this point in time, not entirely new features.

PR #369, opened 2019-06-12
Emergency fix for name command

Gestalt-R:
The string I needed was freed for no particular reason. It should still be in use.


Gestalt-R:
@mpcsh

mpcsh:
Oops, that shouldn't have been an approval, my bad

Gestalt-R:
It gets freed with the rest of the stuff in the hashtable.

mpcsh:
Oh, then that should be an approval :thumbsup:

PR #374, opened 2019-06-12
fixed* memory leak

Gestalt-R:
Edit: Read to the very end, it gets better!
#369 
Arghh.. I'm tired and I let a leak slip through. While I do free the tables, it doesn't get rid of the rest of the command. As it is, it leaks a lot. This pull should fix it. The memory leak is exactly the size of all of the user input combined up to that point right now. 

This PR makes the only stored info the stuff we need for the new entries in the hashtable.

That minimizes leaking, but it's still there... technically. The memory is actively useful until the context struct is gotten rid of, which is good enough for the presentation. 

Really, this should be a backlog issue. In order to properly dispose of memory with the name command, I should have either rewritten the init_table or whatever it's called to use malloc, or add a linked list of pointers to be freed to the context struct. 

The best way to get rid of the leak entirely, if 0 prejudice is wanted, is this PR, but then deleting the entire name_operation, and any function call that references it. (Not too much to do.)

### DIFF
--- a/src/cli/src/cmd.c
+++ b/src/cli/src/cmd.c
@@ -10,7 +10,9 @@
 void add_entry(char *command_name, operation *associated_operation, action_type_t *action, lookup_t **table)
 {
     lookup_t *t = malloc(sizeof(lookup_t));
-    t->name = command_name;
+    char * newname = malloc(sizeof(char) * strlen(command_name));
+    strcpy(newname, command_name);
+    t->name = new_name;
     t->operation_type = associated_operation;
     t->action = action;
     HASH_ADD_KEYPTR(hh, *table, t->name, strlen(t->name), t);
@@ -127,6 +129,7 @@ int lookup_t_free(lookup_t **t)
     HASH_ITER(hh, *t, current_user, tmp)
     {
         HASH_DEL(*t, current_user);
+        free(current_user->name);
         free(current_user);
     }
     return SUCCESS;
--- a/src/cli/src/cmd.c
+++ b/src/cli/src/cmd.c
@@ -14,7 +14,7 @@
 void add_entry(char *command_name, operation *associated_operation, action_type_t *action, lookup_t **table)
 {
     lookup_t *t = malloc(sizeof(lookup_t));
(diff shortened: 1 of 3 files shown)


Gestalt-R:
I thoroughly apologize for rushing the pull request #369. You can also revert it if you don't want to bother with this. My guilty conscience wouldn't let me leave it alone after I realized Mark was right. I was too tired to realize I hadn't created the free function for the hashtable with everything in mind.

Gestalt-R:
Most likely, nobody will review this before the presentation. I hope nobody gets screwed over by my bad decision. @borjasotomayor @mpcsh @lydiafilipe

Gestalt-R:
Ignore what I said earlier. This resolves all the leaks.

Gestalt-R:
I said that assuming this commit here was pushed. Sorry.

Gestalt-R:
Like seriously, there are not even "necessary" leaks. They are all gone.

src/cli/src/cmd.c
```diff
@@ -14,7 +14,9 @@
 void add_entry(char *command_name, operation *associated_operation, action_type_t *action, lookup_t **table)
 {
     lookup_t *t = malloc(sizeof(lookup_t));
-    t->name = command_name;
+    char * newname = malloc(sizeof(char) * strlen(command_name));
```
mpcsh: Should be `char *newname`

src/ui/src/print_functions.c
```diff
@@ -126,6 +126,11 @@ void print_cli(chiventure_ctx_t *ctx, window_t *win)
         do_cmd(c, &quit, ctx);
     }
 
+
```
mpcsh: This extra newline should be deleted

Gestalt-R: Did so.

src/cli/src/cmd.c
Gestalt-R: Yup.

Gestalt-R:
This should not be put into master until after the presentation. I consider it very useful for anyone continuing to work on this project.

PR #375, opened 2019-06-12
HOTFIX: Fix helptext formatting

ecoble:
Fixed an issue where the help text was formatting incorrectly


ecoble:
CLOSING STATEMENT: The help text now prints correctly, there was a bug being caused because a call to move the cursor was inside of that while loop that handles large blocks of text, it has been moved outside of that loop

lydiafilipe:
Issue Score: ✓+

Comments:
Nice work!

PR #379, opened 2019-06-12
HOTFIX: Comparison bug in actionmanagement.c

SnowballAntrobus:
This is a quick bug fix for the actionmanagement.c In our last PR, I introduced a bug in our comparison tests. In this bug fix, I have fixed the bug and everything works again.


SnowballAntrobus:
Closing Statement:
I introduced a bug last PR by trying to negate ints like in Java Script. This has now been fixed we now use a normal equality comparison with the FAILURE macro.

Issue #381, opened 2019-06-23
Reorganize directory structure


borjasotomayor:
The original directory structure we chose, mostly motivated by having one Makefile per component, turned out to be a bit too convoluted. With the move to CMake (#380), we should reorganize the files into a simpler directory structure. I propose the following:

- `include/`: All shared header files, separated by component (e.g., `include/ui/`, `include/cli/`, etc.). This will allow include statements of this form:
  
  ```
  #include "cli/shell.h"
  ```
  
  This will prevent conflicts if two components have header files with the same name.

- `src/`: All C files and private header files, separated by component (e.g., `src/ui/`, `src/cli/`)

- `tests/`: All testing code.

- `docs/`: Documentation

- `games/`: Sample games (playable by end-users; games written purely for testing purposes should go in `tests/`)

borjasotomayor:
Closed by #397

Issue #380, opened 2019-06-23
Switch to CMake for building chiventure


borjasotomayor:
The Makefiles for the project have gotten pretty unwieldy, and don't lend themselves easily to checking software dependencies, generating installable packages, etc. We will switch to [CMake](https://cmake.org/), using a single top-level build file, instead of one per component. A good example to follow can be the [chitcp](https://github.com/uchicago-cs/chitcp) project, which conveniently also has a dependency on `protobuf-c`

borjasotomayor:
As it turns out, CMake recommends creating a CMakeLists.txt file per component as well, instead of a massive top-level one. However, unlike Makefiles, CMake handles all the dependency resolution in the backend, so the per-component files seem worthwhile for the sake of modularity.

borjasotomayor:
Finished migrating the main build to CMake (see commit 27674708). It should now be possible to build chiventure by running this:

    mkdir -p build
    cd build
    cmake ..
    make

The CI file has also been changed to reflect this.

Building examples and tests hasn't been migrated to CMake yet, but I think we should defer that until we've modified the directory structure.

borjasotomayor:
Closed by #397

PR #384, opened 2019-07-30
Migrate build to CMake (executable only)

borjasotomayor:
This PR migrates the build system from Makefile to CMake for the main chiventure executable only (see #380 for more details). The building of the examples and tests hasn't been migrated yet, but we should defer that until #381 is closed. Then, I'll submit another PR that builds the examples and tests with CMake.


borjasotomayor:
Ok, that required a few more fixes before it would build on Travis, but this should be ready for review now.

Issue #383, opened 2019-07-16
Create CLI load command


ecoble:
Create a general load command that allows the user to specify which WDL game file to load, if they didn't load a game when first starting the program. Several cli load commands were created for specific games for demo purposes, these will be removed.

Note: Ask Borja if the user should be able to switch and load a different WDL file if they have already loaded a WDL file

ecoble:
Created a load_wdl command to load a wdl file from the Chiventure cli if one was not specified when the game was started. An example of how to load a wdl file would be:

load_wdl ../test.yaml

If no file path is specified the cli returns an error, if an invalid file path is specified the game aborts

I also had to remove some print statements that were part of debugging when the game loaded, because they did not format properly in the chiventure CLI. I have just commented them out for now.

ecoble:
This issue was closed with PR #385

PR #385, opened 2019-09-06
Load a WDL file into the game

### LINKED ISSUE #382: Load Game from WDL file
Load a WDL game file from the command line when starting the game (Some of this work has already been done in the WDL branch). If the user does not specify a game file, Chiventure will still start but user will have to load a game from the Chiventure CLI

ecoble:
Closes #382 and #383 

This PR allows a game to be loaded from a WDL file when the game is started, or from the chiventure CLI if a WDL file wasn't specified upon starting. The first way to do this would be:

./chiventure /path/to/wdl/file

and the second way from the chiventure CLI would be

load_wdl /path/to/wdl/file

Currently the a game can be loaded and played but there are several problems that I think separate issues should be created for:

- The player cannot enter the final room of the game since this ends the game
- To avoid generic success and error messages, 'text_success' and 'text_fail' are required in the wdl file for each action on an item
- A player object is not loaded when the game is loaded, so actions like 'inventory' fail

Another thing to note is some wdl loading debugging messages caused issues on the chiventure CLI if the user loaded a game once chiventure was already running, for now I just commented out those print statements


ecoble:
Updated how actions are added so that success and failure text is optional again on an object's actions. If the success or failure text is not present, a generic message will be added, otherwise the one provided will be used.

ecoble:
A player, "player1", is now loaded to the game when it starts. Commands like "INV" now work, however the action "take" which should add items to the inventory, does not work.

ecoble:
Many of the sample yaml files were essentially duplicates with different names, so I removed them. As discussed the connections and moving between rooms works, the command is "go" not "walk". I have updated the error message to be "This action is not supported" instead of "You cannot perform this action", which I think may be more descriptive as to the error.

ecoble:
I ran all the files through astyle so the brackets should all be allman now, and fixed the issues in the header comments.

Issue #387, opened 2019-09-13
Create End of Game Behavior


ecoble:
Currently when the player tries to enter the final room, they are prevented from doing this because there is no code to end the game. For now, I will just allow the user to enter the final room then display a message telling them they have completed the game and may now quit.

ecoble:
Added functionality so that when the user enters the final room a generic success message appears. Currently this cannot be tested until PR #385 that allows for loading a game is merged

ecoble:
Closed with #403

PR #403, opened 2020-03-14
Create simple end of game 

ecoble:
Currently if a user tries to move into the final room to end the game, they get an error message. This PR allows for simple end of game behavior for now, the user moves into the final room and a message is printed telling them they have won. Additionally this fixes the "game fails type checking" bug. Closes issue #387


ecoble:
The fix to allow loading games has caused one of the tests (game type check) to fail, working on debugging this. There seems to be a conflict with the labelling of object types, specifically "0.start" vs just "start"

borjasotomayor:
The PR seems to add a whole bunch of duplicate files that all have " 2" added to the filename. Could you remove these from the PR? (I'm guessing they're backup files that inadvertently got added to a commit)

ecoble:
Duplicate files have been removed, game loading issue is now fixed as well as the test that was failing.

Issue #389, opened 2019-10-06
Fix 'Take' Command


ecoble:
The TAKE command should add the object to a player's inventory (if the object is able to be taken), but currently it does not. Fix this issue so that items that have successfully been taken can be seen in the inventory.

ecoble:
Currently there exists a function, "add_item_to_player", which adds an item to a player's inventory, but it is not in use. I am thinking of adding this function to update the inventory either in action management code or later in game state code, however both will require refactoring to have the necessary parameters for this function.

ecoble:
"TAKE" command now adds an item to the players inventory. However it is causing a seg fault, but this seg fault does not cause the game to crash

ecoble:
The seg fault is being caused by the call to HASH_ADD_KEYPTR, line 109 in player.c, which I am unsure how to fix. I have commented out the call to add_item_to_player for now, until the seg fault can be fixed

ecoble:
I've added the call to add_item_to_player() back in as we try to get game files working, but I am still struggling to debug the seg fault

borjasotomayor:
Following up on today's Zoom meeting (and so there's a written record), I believe the issue is that uthash does not allow an object to belong to multiple hash tables at the same time. So, before calling add_item_to_player, the item should be removed from the room. 

fyi, I also noticed there was an issue with delete_all_items that is likely unrelated to the segfault, but which needs to be fixed too. I've created a `fix-take` branch that fixes this issue. Please commit additional fixes for the take command into that branch.

ecoble:
Removing the item from the room's hash table before adding it to the inventory seems to have fixed the seg fault!

PR #410, opened 2020-04-03
Fix the TAKE command

ecoble:
This PR closes issue #389, making the TAKE command functional so users can take items from the room (provided that the item can be taken), add them to their inventory, and see the items in their inventory. This also fixes an issue with the `delete_all_items` function.


ecoble:
I am going to change the orb and distaff files back to .wdl extensions, to be consistent with Dawson's change in the other PR.

PR #417, opened 2020-04-08
Fixes memory leaks #374

Gestalt-R:
See https://github.com/uchicago-cs/chiventure/pull/374
The changes made roughly follow this form:
In the UI segment: previously, each string put in the input buffer for CLI was malloced but not freed. This was a conscious decision made unilaterally by me and mostly unnoticed by others last year, in order to reuse strings in the buffer as internal references.

(This was an awful idea, made late at night in confusion. I fixed it almost immediately after, but the request was denied due to it being less than 24 hours before final presentations.)

In the CLI segment:
The system adding entries for new commands to the lookup table is the only thing that needs to "remember" raw input, with the deletion of the history command. So I made it copy strings and free the new ones later. 

Lastly, I also left some constructive comments in case someone wants to add HIST back in. (If they want to, let me know if they need help!)

### DIFF
--- a/src/cli/src/cmd.c
+++ b/src/cli/src/cmd.c
@@ -12,7 +12,7 @@
 void add_entry(char *command_name, operation *associated_operation, action_type_t *action, lookup_t **table)
 {
     lookup_t *t = malloc(sizeof(lookup_t));
-    char *newname = malloc(sizeof(char) * strlen(command_name));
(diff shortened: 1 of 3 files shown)


Gestalt-R:
The last errors were freeing outside in instead of inside out, and an off-by-one error for strcpy

PR #420, opened 2020-04-26
Updated developer guidelines

### LINKED ISSUE #418: Update Developer Guidelines
The [developer guidelines](https://github.com/uchicago-cs/chiventure/blob/master/docs/developer.md) need to be updated for 2020. Some of the updates are pretty routine (like the new directory structure), but we need to decide on how we're going to update the branching strategy, which did not work that well last year (encouraging large PRs instead of small incremental PRs).

I believe at the end of last year's 220, we discussed switching to the following model:

- There are no per-component branches (like wdl/dev, cli/dev, etc. last year).
- "dev" takes on the role that "master" had last year: there is a high bar to merge into "dev", but teams now create branches straight out of "dev" (instead of spininng a branch out of their component-specific dev branch, which then needs to be merged into "master")
- "dev" is treated as a "pre-master" branch, where feature branches are integrated. Merging to "dev" requires a review by a senior developer.
- Merging from "dev" to "master" represents a "stable" release of chiventure. It is initiated by the product manager, and has no bearing on points, grades, etc.

In this model, there doesn't seem to be a way for having "non-master" PRs like we did last year, but my impression is that these were rarely used, and the goal is to instead have plenty of small PRs into "dev" (with approval from a senior developer), so those should still provide an opportunity to rake in points.

Also, to encourage more frequent PRs, do we want to require them to have two reviews (one student and one senior dev) instead of three (two students and one senior dev)? We can encourage them to request additional reviews when necessary (e.g., if you're writing code that affects other components), but only require two reviews for the code to be merged.

borjasotomayor:
Updated the developer guidelines to reflect the new directory structure, and the new branching model we will be using.

I have also added a `dev` branch and made it a protected branch. It looks like GitHub now includes an option to require that a branch be up to date with the branch you're merging with. I've enabled this since it seems like a sensible way of avoiding integration issues, but let me know if there's a reason why we may want to disable this.

This pull request closes #418.

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -4,10 +4,12 @@
 
 **i.e., where to put your files**
 
-The chiventure repository has three top-level directories:
+The chiventure repository has five top-level directories:
 
 - `src/`: Source code
+- `include/`: Shared include files
 - `docs/`: Documentation
+- `tests/`: Tests
 - `games/`: Sample games
 
 No other top-level directories can be created without prior discussion (ideally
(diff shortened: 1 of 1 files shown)


docs/developer.md
```diff
@@ -61,19 +74,32 @@ This component is responsible for any non-CLI elements of the user interface, su
...
+Your code must follow the [Style Guide](https://uchicago-cs.github.io/dev-guide/style_guide.html) in the [UChicago CS Developer Guide](https://uchicago-cs.github.io/dev-guide) 
 
 ## Git branches
 
 In general, we will be following Vincent Driessen's [git flow](https://nvie.com/posts/a-successful-git-branching-model/) model of branching. Please take a few minutes to read that article in full.
 
 There are a few modifications we will be making to Vincent's model:
 
-* As you read the article, whenever you see "merge to `master`", mentally replace that with "pull request to `master`".
-* We will not have release or hotfix branches.
```
dwahme, line 73: Are we going to keep this information in here, or not worry about it?

borjasotomayor: What are you referring to by "this information"?

dwahme: * As you read the article, whenever you see "merge to `master`", mentally replace that with "pull request to `master`".
* We will not have release or hotfix branches.

dwahme: The above ^^^

borjasotomayor: That's being removed from the developer guidelines. It's superseded by the fact that they won't be able to do pull requests to master, and that only the senior devs will be able to do release branches.

Issue #273, opened 2019-06-04
Support for Quit Command


ecoble:
As part of CLI/UI integration, we stopped supporting "QUIT" to quit from the game, and instead used UI's CTRL + D to quit. We can reintroduce "QUIT", this may require collaboration with UI since CLI is no longer in charge of the command prompt loop.

samdonio:
Just met with Elizabeth, it seems like the print_cli function is what controls what is entered into the cli, and this in turn is called by the window_print function. Right now both of these functions are void, but one way of adding quit could be to test if the user enters "quit" and return some value through both of them and then have this value checked in the game loop where it is checked if the user pressed "CTRL+D"

mberamendiconde:
^^^ I was also in this meeting with Elizabeth, and we bounced off approaches like Sam mentioned above

samdonio:
Return values were added to several functions in order to determine if the user inputs QUIT, however now chiventure doesn't quit at all. After using gdb, the return values seem correct, so QUIT not working doesn't immediately make sense, but I will look into possibly using the getstr() function and if it the user types QUIT then end_win is called.

samdonio:
Created the cli/quit-working-solution branch that now has a working solution to adding the quit command to chiventure. An out parameter was added to many of the functions that interact with the cli in the ui module and if the term "QUIT" is inputted by the user the pointer 'quitval' has value 0, and thus ends the ui loop effectively ending chiventure.

mberamendiconde:
Success! This issue is now closed. The user can type "QUIT" into the command line rather than only using "Ctrl + D" to quit the game and exit the terminal.

ecoble:
Issue Score: ✔️++

Comments:
Great job implementing this backlog issue! I know working with the UI code can be tricky, nice work!

PR #590, opened 2020-05-14
Added a full-game test to the CLI

borjasotomayor:
This PR ultimately adds a full-game test to the CLI (where a game is created in-memory from scratch, so we can test CLI functions on it). However, this required some non-trivial re-factoring, because there was no way to test the CLI functions without starting the UI (which prevents the tests from running, since they get stuck in a UI loop). The specific changes are:

- `do_cmd` has been altered to take a callback function (with callback arguments) as parameters. `print_cli` now calls `do_cmd` with a callback function that calls `print_to_cli`.
- The `quit` output parameter of `do_cmd` has been replaced by a return value.
- There was another call to `do_cmd` from `testshell.c` that has also been updated.

With these changes, the full-game test is able to pass a callback function that simply checks whether the output string from the CLI matches the expected one.

The tests all pass, and I've checked that the CLI still works from the UI (running the sample games)

### DIFF
--- a/include/cli/cmd.h
+++ b/include/cli/cmd.h
@@ -217,17 +217,45 @@ cmd *cmd_from_string(char *s, chiventure_ctx_t *ctx);
 cmd *cmd_from_tokens(char **ts, lookup_t **table);
 
 
+#define CLI_CMD_SUCCESS (0)
+#define CLI_CMD_SUCCESS_NOOUTPUT (1)
+#define CLI_CMD_SUCCESS_QUIT (2)
+#define CLI_CMD_CALLBACK_ERROR (3)
+
+/*
+ * Callback function type for the CLI.
+ *
+ * Functions of this type can be passed to do_cmd,
+ * which will call the function if the command is run
+ * successfully.
+ *
+ * Parameters:
+ *  - ctx: pointer to chiventure context struct
+ *  - outstring: The string to be printed as a result of running this command
+ *  - args: Additional arguments
+ *
+ * Returns:
+ *  - CLI_CMD_SUCCESS: Success
+ *  - CLI_CMD_CALLBACK_ERROR: Callback function was unable to print the string
+ */
+typedef int cli_callback(chiventure_ctx_t *ctx, char *outstring, void *args);
+
 /*
  * Executes the given command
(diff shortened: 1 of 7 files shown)


src/cli/src/cmd.c
```diff
@@ -255,16 +255,20 @@ void do_cmd(cmd *c,int *quit, chiventure_ctx_t *ctx)
      */
     if (strcmp(cmd_name_tos(c),"QUIT")==0)
```
ecoble, line 256: Not something for this PR but typing "quit" doesn't actually work for Chiventure (it should work for the shell), we should probably implement that

borjasotomayor: ha, looks like you had actually already created an issue for this last year: https://github.com/uchicago-cs/chiventure/issues/273 :-)

include/cli/cmd.h
```diff
@@ -217,17 +217,45 @@ cmd *cmd_from_string(char *s, chiventure_ctx_t *ctx);
...
  *
  * Returns:
- * - nothing -> output handled elsewhere
+ * - CLI_CMD_SUCCESS: Success, and the command produced some output
+ * - CLI_CMD_SUCCESS_NOOUTPUT: Success, but the command did not produce any output
+ * - CLI_CMD_SUCCESS_QUIT: Success, but the command requires that chiventure quit
+ * - CLI_CMD_CALLBACK_ERROR: Callback function reported an error
  */
-void do_cmd(cmd *c,int *quit, chiventure_ctx_t *ctx);
+int do_cmd(cmd *c, cli_callback callback_func, void *callback_args, chiventure_ctx_t *ctx);
```
dwahme, line 258: Minor question- I've generally seen the convention of having the ctx struct be the first argument to functions. Is this something we want to worry about in this PR or later?

borjasotomayor: My preference is for it to come first, but the "convention" in the CLI code seems to be to put it at the end, so I went with that. However, I do think we need to normalize this across the entire codebase.

dwahme: Ok, I think this could be something for a team to address later then

PR #593, opened 2020-05-15
Battle State Module | starting point

alex-sheen:
I set up the basic folder layouts according to the developer guidelines. I then created the following files:
battle_structs.h
battle_state.h
battle_state.c
test_battle_state.c
Within these files, I implemented the battle and character structs with their corresponding new/init/free functions. I left stubs for many modules, including things like moves/stats/items.


src/battle/src/battle_state.c
```diff
@@ -0,0 +1,113 @@
...
+/* See battle_state.h */
+int free_character(character_t *c)
+{
+    assert(c != NULL);
+
+    free(c->name);
+    free(c->stats);
+    free(c->moves);
+    free(c->items);
+    free(c->next);
```
celiaanderson, line 58: Maybe we could have separate functions for freeing one particular character and then for freeing a list of characters? Just for some flexibility later, like if we want to free an enemy character that gets defeated while other enemies are left

src/battle/src/battle_state.c
```diff
@@ -0,0 +1,113 @@
...
+    c->stats = stats;
+    c->moves = moves;
+    c->items = items;
+    c->next = next;
+
+    return SUCCESS;
+}
+
+/* See battle_state.h */
+int free_character(character_t *c)
```
celiaanderson, line 50: Maybe we could have separate functions for freeing one particular character and then for freeing a list of characters? Just for some flexibility later, like if we want to free an enemy character that gets defeated while other enemies are left

alex-sheen: Oh that's a good point, I'll make a second function right now

src/battle/src/battle_state.c
```diff
@@ -0,0 +1,113 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <assert.h>
+#include "battle/battle_state.h"
+
+
+/* See battle_state.h */
+character_t *character_new(char *name, char_t ct, stat_t *stats,
```
celiaanderson, line 9: looks like the char_t name "ct" doesn't match the "type" name used in the header and also in the init function

include/battle/battle_state.h
```diff
@@ -0,0 +1,124 @@
...
+ * - move_t: a pointer to moves module (stub)
+ * - item_t: a pointer to items module (stub)
+ * - next: a pointer to a potential next character in a list
+ * - SUCCESS for successful init
+ * - FAILURE for unsuccessful init
+ */
+int character_init(character_t *character, char *name, char_t type,
+     stat_t *stats, move_t *moves, item_t *items, character_t *next);
+
+/* Frees a character struct from memory */
```
celiaanderson, line 122: missing parameter and return value info

include/battle/battle_state.h
```diff
@@ -0,0 +1,124 @@
...
+/* Creates a new character  struct
+ * Parameters:
+ * - character: a pointer to character in memory
+ * - name: name string
+ * - char_type: char_t indicating character type
+ * - stat_t: a pointer to stats module (stub)
+ * - move_t: a pointer to moves module (stub)
+ * - item_t: a pointer to items module (stub)
+ * - next: a pointer to a potential next character in a list
+ * - SUCCESS for successful init
```
celiaanderson, line 116: just minor thing - missing something like "return:" to separate parameters from return values

include/battle/battle_state.h
```diff
@@ -0,0 +1,124 @@
...
+
+/* Initialize a character  struct
+ * Parameters:
+ * - name: name string
+ * - char_type: char_t indicating character type
+ * - stat_t: pointer to stats module (stub)
+ * - move_t: pointer to moves module (stub)
+ * - item_t: pointer to items module (stub)
+ * - next: a pointer to a potential next character in a list
+ * Return: a pointer to the new character
```
celiaanderson, line 102: for character_new and character_init comments: parameter names aren't consistent with function parameter names

include/battle/battle_state.h
```diff
@@ -0,0 +1,124 @@
...
+ * next: allows for character lists using utlist.h
+ */
+typedef struct _character_t {
+    char *name;
+    char_t type;
+    class_t class;
+    stat_t *stats;
+    move_t *moves;
+    item_t *items;
+    struct _character_t *next;
```
celiaanderson, line 50: If we keep track of a list of characters this way, we may want to have a function for updating the enemy pointer within the battle if we add in an enemy during the battle? - since once we add in another enemy to the front of this list, the pointer to the enemy struct that would be currently in the battle struct may not be able to find the new enemy. Though for now it's probably not a big deal if we're just going to be starting out with single enemies

include/battle/battle_state.h
alex-sheen: Gotcha, the parameter names and function parameters names didn't match for the battle ones too.

include/battle/battle_state.h
alex-sheen: Yeah, we should write that down for later. The function would have to be able to loop through the list. I'm now thinking that we might want to change the character struct to have a previous value, since utlists has a built in double linked list. But we can always add that in a future issue.

src/battle/src/battle_state.c
```diff
@@ -0,0 +1,141 @@
...
+
+    return SUCCESS;
+}
+
+/* battle_state.h */
+int free_character(character_t *c)
+{
+	assert(c != NULL);
+
+	if(c->next != NULL)
```
dwahme, line 54: It's a better idea to simply free a single character here instead of trying to free the entire list. Instead, calling `free` on a single character should just free associated memory and remove itself from the list if it hasn't been already.

utlist provides a macro for removing from the list called `LL_DELETE` (or `DL_DELETE` for doubly linked lists)

src/battle/src/battle_state.c
```diff
@@ -0,0 +1,141 @@
...
+    if(c != NULL)
+	{
+    	free(c->name);
+    	free(c->stats);
+    	free(c->moves);
+    	free(c->items);
+    	int rc = free_characters(c->next);
+    	free(c);
+	}
+    return SUCCESS;
```
dwahme, line 89: You should use `LL_FOREACH_SAFE` here to iterate over the linked list (or `DL_FOREACH_SAFE` for doubly linked lists). It'll look something like

```
character_t *elem, *tmp;

DL_FOREACH_SAFE(c, elt, tmp) {
    DL_DELETE(c, elem);
    free(elem);
}
```

src/battle/src/battle_state.c
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/* See battle_state.h */
+battle_t *battle_new(character_t *p, character_t *ene,
+    environment_t env, char_t ct)
+{
+    battle_t *b;
+    int rc;
+    b = malloc(sizeof(battle_t));
```
dwahme, line 98: Use `calloc` instead of `malloc`

src/battle/src/battle_state.c
```diff
@@ -0,0 +1,141 @@
...
+    c->stats = stats;
+    c->moves = moves;
+    c->items = items;
+    c->next = next;
+
+    return SUCCESS;
+}
+
+/* battle_state.h */
+int free_character(character_t *c)
```
dwahme, line 50: This should be named `character_free()` instead to follow convention

src/battle/src/battle_state.c
```diff
@@ -0,0 +1,141 @@
...
+		free(c->stats);
+		free(c->moves);
+		free(c->items);
+		free(c);
+		return SUCCESS;
+	}
+}
+
+/* See battle_state.h */
+int free_characters(character_t *c)
```
dwahme, line 78: This should be named `character_free_all()` instead to follow convention

src/battle/src/battle_state.c
```diff
@@ -0,0 +1,141 @@
...
+    b->player = p;
+    b->enemy = ene;
+    b->environment = env;
+    b->turn = ct;
+
+    return SUCCESS;
+}
+
+/* See battle_state.h */
+int free_battle(battle_t *b)
```
dwahme, line 131: This should be `battle_free()` to follow convention

tests/CMakeLists.txt
```diff
@@ -3,4 +3,5 @@ add_subdirectory(checkpointing)
 add_subdirectory(cli)
 add_subdirectory(game-state)
 add_subdirectory(ui)
-add_subdirectory(wdl)
\ No newline at end of file
+add_subdirectory(wdl)
+add_subdirectory(battle)
```
dwahme, line 7: Please add this in alphabetical order

tests/battle/test_battle_state.c
```diff
@@ -0,0 +1,116 @@
...
+    rc = character_init(&c, "character_init_Name",PLAYER, NULL, NULL, NULL, NULL);
+
+    cr_assert_eq(rc, SUCCESS, "character_unit() failed");
+
+    cr_assert_str_eq(c.name, "character_init_Name", "character_new() didn't set name");
+    cr_assert_eq(c.type, PLAYER, "character_new() didn't set type");
+}
+
+/* Tests free_character() */
+Test(battle_state, free_character)
```
dwahme, line 35: Change this test name to `character_free`

tests/battle/test_battle_state.c
```diff
@@ -0,0 +1,116 @@
...
+
+    cr_assert_not_null(c, "character_new() failed");
+
+    rc = free_character(c);
+
+    cr_assert_eq(rc, SUCCESS, "character_free() failed");
+}
+
+/* Tests free_characters() */
+Test(battle_state, free_characters)
```
dwahme, line 50: Change this to `character_free_all` instead

tests/battle/test_battle_state.c
```diff
@@ -0,0 +1,116 @@
...
+{
+    character_t *c1;
+	character_t *c2;
+    int rc;
+
+    c2 = character_new("free_characters_Name2",PLAYER, NULL, NULL, NULL, NULL);
+    c1 = character_new("free_characters_Name1",PLAYER, NULL, NULL, NULL, c2);
```
dwahme, line 57: Once you update `character_new` to not immediately pass in `next`, this will need to be something of the form

```
c1 = character_new("free_characters_Name1",PLAYER, NULL, NULL, NULL, NULL);
c2 = character_new("free_characters_Name2",PLAYER, NULL, NULL, NULL, NULL);
DL_APPEND(c1, c2);
```

tests/battle/test_battle_state.c
```diff
@@ -0,0 +1,116 @@
...
+    cr_assert_eq(rc, SUCCESS, "character_unit() failed");
+
+    cr_assert_not_null(b.player, "battle_new() didn't set player");
+    cr_assert_not_null(b.enemy, "battle_new() didn't set enemy");
+    cr_assert_eq(b.environment, ENV_SNOW, "character_new() didn't set environment");
+    cr_assert_eq(b.turn, ENEMY, "character_new() didn't set turn");
+}
+
+/* Tests free_battle */
+Test(battle_state, free_battle)
```
dwahme, line 102: Change this test name to `battle_free` instead

include/battle/battle_state.h
```diff
@@ -0,0 +1,140 @@
...
+ * ENV_SNOW: snow
+ * ENV_WATER: water
+ * (temporary values, to be developed at a later time)
+ */
+typedef enum _environment_t {
+     ENV_GRASS = 0,
+     ENV_DESERT = 1,
+     ENV_SNOW = 2,
+     ENV_WATER = 3
+} environment_t;
```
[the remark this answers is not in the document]
alex-sheen: I'm going to add an ENV_NONE. I added those environment variables as placeholders and wasn't thinking too hard about how it would work, but having a default environment that is explicitly the default sounds like good idea.

include/battle/battle_state.h
```diff
@@ -0,0 +1,140 @@
...
+
+/* A character type enum that contains the following:
+ * PLAYER: player
+ * ENEMY: enemy/enemies
+ * (temporary values, to be developed at a later time)
+ */
+typedef enum char_t {
```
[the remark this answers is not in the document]
alex-sheen: That makes more sense. I had it that way because I was thinking they both share a "player" and "enemy" label, but it's probably more confusing than helpful.

include/battle/battle_state.h
```diff
@@ -0,0 +1,140 @@
...
+ */
+typedef struct _character_t {
+    char *name;
+    char_t type;
+    class_t class;
+    stat_t *stats;
+    move_t *moves;
+    item_t *items;
+    struct _character_t *next;
+} character_t;
```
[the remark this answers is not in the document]
alex-sheen: Oh yes, that sounds a lot better. I chose character because I couldn't think of a better word.

include/battle/battle_state.h
```diff
@@ -0,0 +1,140 @@
...
+ * - SUCCESS for successful init
+ * - FAILURE for unsuccessful init
+ */
+int battle_init(battle_t *b, combatant_t *player, combatant_t *enemy,
+    environment_t env, turn_t turn);
+
+/* Frees a battle struct from memory */
+int battle_free(battle_t *b);
+
+/* Initialize a combatantstruct
```
dwahme, line 96: Typo- combatantstruct

include/battle/battle_state.h
```diff
@@ -0,0 +1,140 @@
...
+ * - move_t: pointer to moves module (stub)
+ * - item_t: pointer to items module (stub)
+ * returns: a pointer to the new character
+ */
+combatant_t *combatant_new(char *name, bool is_friendly, stat_t *stats,
+    move_t *moves, item_t *items);
+
+/* Creates a new combatant struct
+ * Parameters:
+ * - c: a pointer to combatantin memory
```
dwahme, line 110: Typo- combatantin

include/battle/battle_state.h
```diff
@@ -0,0 +1,140 @@
...
+/* Frees a battle struct from memory */
+int battle_free(battle_t *b);
+
+/* Initialize a combatantstruct
+ * Parameters:
+ * - name: name string
+ * - is_friendly: bool indicating character type
+ * - stat_t: pointer to stats module (stub)
+ * - move_t: pointer to moves module (stub)
+ * - item_t: pointer to items module (stub)
```
dwahme, line 102: These should say
- pointer to the stats of the combatant
- pointer to the linked list of moves for the combatant
- pointer to the linked list of items for the combatant

include/battle/battle_state.h
```diff
@@ -0,0 +1,140 @@
...
+    move_t *moves, item_t *items);
+
+/* Creates a new combatant struct
+ * Parameters:
+ * - c: a pointer to combatantin memory
+ * - name: name string
+ * - is_friendly: bool indicating character type
+ * - stat_t: a pointer to stats module (stub)
+ * - move_t: a pointer to moves module (stub)
+ * - item_t: a pointer to items module (stub)
```
dwahme, line 115: These should say
- pointer to the stats of the combatant
- pointer to the linked list of moves for the combatant
- pointer to the linked list of items for the combatant

alex-sheen:
To summarize, this pull request implemented the initial battle_state, battle_structs, and test_battle_state code. This includes the battle and combatant structs and their associated functions and struct definitions and tests. It is worth highlighting that the battle_state declares a combatant_t type which will store the information for players and enemies within the battle.

dwahme:
Issue Score: ✔️+

Comments:

The summary for this issue is too terse. A good rule of thumb is that you should be writing at least one full paragraph in your summary (see https://uchicago-cs.github.io/cmsc22000/project/policies/#a-descriptive-summary for examples). A quick explanation of each module would have been helpful for reviewers.

Issue #561, opened 2020-05-12
Add proper CMake file to skill trees skeleton code then make pull request


jackandthebean:
# Add proper CMake file to skill trees skeleton code then make pull request
### What
Our skeleton code consists of two files: `skilltrees.h` and `skilltrees.c`, which implement the structs that will hold chiventure-supported skills, as well as their associated new, init, and free functions. The skeleton code also implements basic functions that manipulate skill lists—adding/removing skills—and retrieve skill types that already exist within a given skill list. These files need a `CMakeLists.txt` to compile and progress into the PR phase.

### Why
Our skeleton code is ready for our first PR, #601.

jackandthebean:
3a175ce

jackandthebean:
I'm closing this issue, as the CMake files were added and our PR has been reviewed. I've now made the suggested changes to the PR (77e54f6) and look forward to making more extensive tests!

jackandthebean:
I am closing this issue, as our skill trees skeleton code was successfully merged into `dev`! Besides making @firatciftci 's suggested changes, I reworked all the code to match the style of all our assignments more closely, increased the flexibility of the skill "effect" function pointers, and added a `skill_execute()` function which simply executes the function pointed to by a given `skill_t` struct. The flexibility of the skill effect functions was one of our team's most contentious issues, the solution for which I documented in #559. With this skeleton code merged into `dev`, hopefully the process of merging our more advanced functions will be more seamless; we look forward to making those future PRs!

firatciftci:
Issue Score: ✔️++

Comments:
Amazing work!

PR #601, opened 2020-05-18
Skill Trees: Skeleton Code

### LINKED ISSUE #527: Begin implementation phase of skill trees.
After designing what our data strutures would look like, skilltrees needs to begin writing the code to hold skills. At the end of this sprint, the code will be saved onto a new branch of chiventure called 'skilltrees'. The end goal should be to fully write necessary structs for holding and creating skills, as well as designing structs for the skills themselves, and additional functions that allow for basic manipulation of skills like freeing, creating, testing if a skill is in a list, etc.

jackandthebean:
# Skill Trees: Skeleton Code
### Following up on initial implementation in Sprint 2
We devoted the first two sprints to the design of the `x/rpg-skilltrees` module. Towards the end of the second sprint, @VannStrasen began the initial write-up of our implementation (#527), drawing from the data structures and function headers we developed during our team design discussions. He closed #527 with the following:
> I initially wrote this code in a branch called `skilltrees`, which I realized was less descriptive than necessary, so I shifted the code over to `skilltrees/skeleton_code`. Because I also did not use the correct format for `.c` and `.h` files (they were in the wrong folder), the `Makefile` was not formatted properly. While it was compiling correctly on my `Makefile`, I am not sure if it will compile correctly in the chiventure `Makefile`, because I haven't been able to work on that yet. As Sprint 2 is ending now, I'm not issuing a pull request until the Makefile is finished—that'll be work for the beginning of the next sprint.

Now I have added a properly formatted `CMake` file (#561), updated the top-level chiventure `CMake` file, and our skeleton code builds a static library without error, so I am creating this pull request to incorporate the foundational skill trees code into the `dev` branch.
### Specific files to be added
Specifically, our skeleton code consists of two files: `skilltrees.h` and `skilltrees.c`, which implement the structs that will hold chiventure-supported skills, as well as their associated new, init, and free functions. The skeleton code also implements basic functions that manipulate skill lists—adding/removing skills—and retrieve skill types that already exist within a given skill list. We are currently developing the more advanced skill tree functions and look forward to incorporating those as well, following the approval of this foundational pull request.


include/skilltrees/skilltrees.h
```diff
@@ -0,0 +1,90 @@
...
+// Player’s skill, unique to each character:
+struct skill_t {
+    char* name;
+    unsigned int cur_level;
+    unsigned int experience;
+    char* description;
+    int (*execute_skill)(int param1, int param2);
+};
+
+typedef struct skill_t skill_t;
```
firatciftci, line 11: Minor comment on style: You can have this written in a simpler manner:
```c
typedef struct skill {
.
.
.
} skill_t;
```

include/skilltrees/skilltrees.h
```diff
@@ -0,0 +1,90 @@
...
+
+// Struct which holds a player's every skill, differentiated by passive and
+// active.
+// When not holding a skill, each space in the passive or active skill lists is
+// defined as NULL.
```
firatciftci, line 16: Comment on style: You should use multi-line comments (`/* ... */`) instead of regular comments (`// ...`) for function headers. So this would look like the following:
```c
/* struct_name
 *
 * active_skills: [description]
 * cur_num_active: [description]
 * ...
 */
```

The same goes for function header comments, where instead of struct field descriptions, you should list the input parameters and also the return value.

include/skilltrees/skilltrees.h
```diff
@@ -0,0 +1,90 @@
...
+struct allskills_t {
+    skill_t** active_skills;
+    unsigned int cur_num_active;
+    unsigned int max_num_active;
+    skill_t** passive_skills;
+    unsigned int cur_num_passive;
+    unsigned int max_num_passive;
+};
+
+typedef struct allskills_t allskills_t;
```
firatciftci, line 26: The name `allskills_t` seems a bit vague to me; without looking at the source code, it's easy to mistake it with a struct for all of the skills in the game. To make it a bit clearer, I think a name such as `player_skills_t` would be better.

include/skilltrees/skilltrees.h
```diff
@@ -0,0 +1,90 @@
+
+// Player’s skill, unique to each character:
+struct skill_t {
+    char* name;
+    unsigned int cur_level;
+    unsigned int experience;
+    char* description;
+    int (*execute_skill)(int param1, int param2);
+};
```
firatciftci, line 9: You might want to add an integer field `id` to keep track of the skills, instead of relying on the `name` field. This way, you would be able to have two skills with the same name (assuming they have different IDs). This would also let you create new `skill_t` structs with just an integer value, rather than dealing with string manipulation from the get-go.

include/skilltrees/skilltrees.h
```diff
@@ -0,0 +1,90 @@
...
+                   int (*given_skill)(int, int));
+
+// Frees a skill.
+// Returns an error code.
+int free_skill(skill_t* skill);
+
+// Creates a new skill list, initially mallocing space for 5 active and 5
+// passive skills.
+// Returns the malloc'd list.
+allskills_t* create_skill_list();
```
firatciftci, line 40: You might want to take in the max space for active and passive skills as function parameters, instead of hardcoding it to 5 for each by default.

jackandthebean:
@firatciftci thank you for the feedback! I made all the changes you suggested, reworked all the code to match the style of all our assignments more closely, increased the flexibility of the skill "effect" function pointers, and added a `skill_execute()` function which simply executes the function pointed to by a given `skill_t` struct.

jackandthebean:
I will put my closing statement for this closed PR at the end of #561.

firatciftci:
Issue Score: ✔️++

Comments:
Great work on this PR!

PR #610, opened 2020-05-19
Rpg npc: Npc Room Management Basic Features

philipchenaux:
This pull request concerns the basics of the npc room_management feature (i.e. npcs spawning in rooms). This request includes the addition of a npcs_in_room struct as well as the basic functions including new, init, free, as well as counting the number of npcs in a room, and adding an npc to a room. We have taken the existing code from the npcs team and made our own addition to their modules, therefore we have made changes to:
```
npc.h
npc.c
test_npc.c
```

We have not changed any of the npc team's code, we have only added our own code to their files, the particular functions we have added are the following:
```
npcs_in_room_init
npcs_in_room_new
npcs_in_room_free
get_num_of_npcs
add_npc_to_room
```
In addition we have added the following struct:
```
npcs_in_room_t
```
There may appear to be a lot of revisions, but most of it was copied code from the npcs branch that they are currently in the midst of a pull request (so it is most likely the most recent code).


include/game-state/npc.h
```diff
@@ -0,0 +1,200 @@
...
+* used to point to the npc_t structs in the traditional sense,
+* and those which are used to hash npc_t structs with the
+* UTHASH macros as specified in src/common/include */
+ typedef struct npc npc_hash_t;
+
+/* Struct for addings npcs to rooms */
+typedef struct npcs_in_room {
+    /* hh is used for hashtable, as provided in uthash.h */
+    UT_hash_handle hh;
+    char *room_id;
```
avianurag, line 29: Why is the room_id stored as a string? Doesn't it make more sense to store these simply as integers, and allow the future developers the option to add a more versatile naming process for the rooms? I think this might be a minor information leak, and having this as an int would be more efficient. @firatciftci does this make sense?

src/game-state/src/npc.c
```diff
@@ -0,0 +1,175 @@
...
+
+    return SUCCESS;
+}
+
+/* See npc.h */
+int npc_room_init(npcs_in_room_t *npcs_in_room, char* room_id)
+{
+    assert(npcs_in_room != NULL);
+
+    strncpy(npcs_in_room->room_id,room_id,strlen(room_id)+1);
```
avianurag, line 21: This use of strncpy is what made me wonder about the use of a string to store the room_id. Just seems a little unnecessary, but of course, I may be wrong, or this may be something decided by openworld!

src/game-state/src/npc.c
```diff
@@ -0,0 +1,175 @@
...
+
+    return npc;
+}
+
+/* See npc.h */
+ npcs_in_room_t *npcs_in_room_new(char* room_id)
+ {
+    npcs_in_room_t *npcs_in_room;
+    npcs_in_room = malloc(sizeof(npcs_in_room_t));
+    memset(npcs_in_room, 0, sizeof(npcs_in_room_t));
```
avianurag, line 51: Please teach me about this function @philipchenaux @shayna-k :)

tests/game-state/test_npc.c
```diff
@@ -0,0 +1,171 @@
...
+	npc_t *npc; 
+	npc = npc_new("npc_22", 20, NULL);
+
+	cr_assert_not_null(npc, "npc_new() failed");
+	
+	cr_assert_eq(npc->npc_id, "npc_new didn't set npc_id"); 
+	cr_assert_eq(npc->health, 20, "npc_new() didn't set health"); 
+}
+
+
```
avianurag, line 19: This is just a style thing, but I think you are supposed to write a short comment describing what exactly this function is testing.

tests/game-state/test_npc.c
```diff
@@ -0,0 +1,171 @@
...
+	//NULL case 
+	item_t* itm = item_new("itm", "short desc", "longer description"); 
+	npc_t* npc = npc_new("npc_22", 20, NULL); 
+	int ret1 = add_item_to_npc(npc, itm); 
+	cr_assert_eq(ret1, SUCCESS, "add_item_to_npc() failed"); 
+
+	//non-NULL Success Case 
+	npc_t* npc2 = npc_new("npc_32", 40, NULL);
+	item_t* itm1 = item_new("itm1", "short desc 1", "longer description 1");
+    item_t* itm2 = item_new("itm2", "short desc 2", "longer description 2");
```
avianurag, line 111: Another extremely minute and stylistic thing, but the indentation is slightly off here.

include/game-state/npc.h
philipchenaux: That makes sense, but it was just to keep it the same as what the open-world team has used for room_ids, i think it is because sometimes they may want to identify the rooms as "dungeon_5" for example or smth.

src/game-state/src/npc.c
philipchenaux: It was more of a convention, but I guess it doesn't matter too much. @firatciftci  do you think this should be changed?

tests/game-state/test_npc.c
philipchenaux: Nice catch Avi !!!

tests/game-state/test_npc.c
philipchenaux: Good point thanks Avi !!!

shayna-k: Alright yup thanks for the note!

include/game-state/npc.h
firatciftci: @avianurag's point is important here, I think forcing room IDs to be strings by default is a bit inefficient. My suggestion would be to have the `room_id` field be numeric, with an optional field in the struct named `room_name` that is of type `char *`. This way, keeping track of the rooms and their names would be more manageable, instead of having to parse strings such as `"dungeon_5"`.

include/game-state/npc.h
```diff
@@ -0,0 +1,200 @@
...
+/*
+ * Allocates a new npcs_in_room struct in the heap
+ *
+ * Parameters:
+ *  room_id: the unique id of the room
+ *
+ * Returns:
+ *  pointer to allocated npcs_in_room struct
+ */
+ npcs_in_room_t *npcs_in_room_new(char* room_id);
```
firatciftci, line 91: By making the `room_id` field numeric, the `new` and `init` functions would take the respective numeric type parameter instead of `char *`.

include/game-state/npc.h
```diff
@@ -0,0 +1,200 @@
...
+/*
+ * Returns the number of npcs in a room
+ *
+ * Parameters:
+ *  npcs_in_room: the struct holding the npcs in the room
+ *
+ * Returns:
+ *  int, the number of npcs in the room
+ */
+int get_num_of_npcs(npcs_in_room_t *npcs_in_room);
```
firatciftci, line 139: Minor stylistic comment: The name of this function doesn't connote that it's for NPCs in a room, so having a more descriptive function name would be better. Maybe go with something like `npcs_in_room_get_number`?

src/game-state/src/npc.c
firatciftci: My comments above should clear this issue up; `strncpy()` should still be used if the function is given a string for the `room_name`, but for the `room_id`, a simple integer assignment should be enough.

include/game-state/npc.h
philipchenaux: Yep thanks, we changed the name of the function!

include/game-state/npc.h
philipchenaux: Yes we  passed in the relevant types into each function given  the change in our struct.

include/game-state/npc.h
philipchenaux: Okay we have separated it into:
```C
int room_id;
char *room_name;
```
in our `npcs_in_room_t` struct.

philipchenaux: Hi, sorry Firat I was just taking at your suggestion again, and I do understand the inherent inefficiency, but it appears that already appears that the room struct  is defined as a `char *room_id`. In addition, it appears that other teams have followed this convention by also including the char* as a form of ID, such as  `char *npc_id`. I would like to change it to an int, but this would involve changing the definitions of these other structs.

firatciftci: I see your point about the room struct; in that case, we can leave that be for now, as modifying the room struct itself would be a much larger task that is not needed to be tackled at the moment.

firatciftci:
Issue Score: ✔️++

Comments:
Great work!

PR #625, opened 2020-05-22
Battle/classes moves

Purrsephone:
Hi fellow devs. 

I would like to request that my branch be merged into dev. I have implemented a mock player class system for use in testing battle code. This includes hard coding 4 mock classes. There are init and new functions, as well as one make_test_class function for each class. Each has a test associated with it. 

Thanks.


src/battle/src/battle_classes.c
```diff
@@ -0,0 +1,118 @@
...
+
+    if(rc != 0)
+    {
+        fprintf(stderr,"Could not initialize class\n");
+        return NULL;
+    }
+
+    return class;
+}
+
```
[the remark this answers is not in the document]
Purrsephone: Ah, these aren't actually test functions. They create structs for use in testing. My bad!

include/battle/battle_classes.h
```diff
@@ -0,0 +1,90 @@
...
+ * - info: a short description of the class 
+ * - st: the stat that the class receives a bonus for 
+ * - bonus: the amount added to the designated stat 
+ * 
+ * Returns:
+ * - 0 for success, 1 if an error occurs 
+ */  
+int init_class(class_t *class, class_type_t cl, char* info, 
+	       possible_stats_t st, double bonus);
+
```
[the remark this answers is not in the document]
Purrsephone: Wasn't planning to do a free_class func since we won't actually be using these functions, but happy to do so if Dawson also thinks I should. Good catch though!

include/battle/battle_classes.h
```diff
@@ -0,0 +1,90 @@
...
+ * cleric 
+ * paladin 
+ * wizard 
+ */
+typedef enum class_type
+{
+    bard,
+    cleric,
+    paladin,
+    wizard
```
dwahme, line 20: rename these to 
```
CLASS_BARD, CLASS_CLERIC, CLASS_PALADIN, CLASS_WIZARD
```
so that we don't get variable name overlap- it's C convention to do constants in all caps

include/battle/battle_classes.h
```diff
@@ -0,0 +1,90 @@
...
+ * strength (str)
+ * dexterity (dex)
+ * charisma (chrsma) 
+ */
+typedef enum possible_stats
+{
+    spd,
+    str,
+    dex,
+    chrsma
```
dwahme, line 38: Make these
```
STAT_SPEED, STAT_STRENGTH, STAT_DEXTERITY, STAT_CHARISMA
```
save the small names for temporary variables in functions, not global constants

include/battle/battle_classes.h
```diff
@@ -0,0 +1,90 @@
...
+ * a short description of the class
+ * which stat the class gets a bonus for
+ * how much that bonus is
+ */
+typedef struct class
+{
+    class_type_t cl;
+    char *info;
+    possible_stats_t st;
+    int bonus;
```
dwahme, line 53: make these
```
    class_type_t class_type;
    char *info;
    possible_stats_t stats;
    int bonus;
```
so that they're easier to understand when we're in the code

include/battle/battle_classes.h
```diff
@@ -0,0 +1,90 @@
...
+ * - info: a short description of the class 
+ * - st: the stat that the class receives a bonus for 
+ * - bonus: the amount added to the designated stat 
+ * 
+ * Returns:
+ * -A pointer to the class, or NULL if a class
+ *  cannot be allocated 
+ */ 
+class_t *new_class(class_type_t cl, char* info, 
+		   possible_stats_t st, double bonus);
```
dwahme, line 71: Make these better variable names like above

include/battle/battle_classes.h
```diff
@@ -0,0 +1,90 @@
...
+ * - cl: a class type
+ * - info: a short description of the class 
+ * - st: the stat that the class receives a bonus for 
+ * - bonus: the amount added to the designated stat 
+ * 
+ * Returns:
+ * - 0 for success, 1 if an error occurs 
+ */  
+int init_class(class_t *class, class_type_t cl, char* info, 
+	       possible_stats_t st, double bonus);
```
dwahme, line 88: Make these better variable names like above

include/battle/battle_classes.h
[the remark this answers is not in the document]
dwahme: If you're convinced that we'd never use `class_free()`, then since it's a mock implementation I'll say we won't need it. I find this scenario unlikely but I'll defer to your judgement here

src/battle/src/battle_classes.c
```diff
@@ -0,0 +1,118 @@
...
+/** See battle_classes.h */
+int init_class(class_t *class, class_type_t cl, char* info, possible_stats_t st, double bonus)
+{
+    assert(class != NULL);
+    assert(info != NULL);
+
+    class->cl = cl;
+
+    class->info = (char*) malloc(sizeof(char) * 100);
```
dwahme, line 17: Always `calloc()` instead of `malloc()`. Also, it's better to `#DEFINE MAX_CLASS_INFO_LEN (100)` in the header of this file, and then `calloc(MAX_CLASS_INFO_LEN + 1, sizeof(char))` instead so that we don't just have random numbers floating around with no meaning

src/battle/src/battle_classes.c
```diff
@@ -0,0 +1,118 @@
...
+    strcpy(class->info, info);
```
dwahme, line 18: Most C string functions (`strcpy`, `strcat`, etc.) have an equivalent "safe" version (`strncpy`, `strncat`, etc.) where you can specify the maximum number of characters to copy/concatenate/etc. Use these instead to protect against accidental (or malicious) buffer overflows

You should use `strncpy()` instead of `strcpy()` (the length is `MAX_CLASS_INFO_LEN`

tests/battle/test_battle_classes.c
```diff
@@ -0,0 +1,108 @@
+#include <criterion/criterion.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include "../../src/battle/src/battle_classes.c"
+#include "../../include/battle/battle_classes.h"
```
dwahme, line 6: You shouldn't ever need to include a .c file, and the other include should just be `#include "battle/battle_classes.h"`

include/battle/battle_classes.h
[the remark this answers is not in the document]
Purrsephone: You right, you right. Made a free func and a test.

dwahme:
Issue Score: ✔️++

Comments:
Great job! Well done with a small, simple PR.

dwahme:
Issue Score: ✔️

Comments:

The title of an issue has to convey what the issue will be producing. In the future, try using simple titles with a "Verb-Object" structure that clearly states what you will be doing in the issue (e.g., “Design and implement a ‘circle’ module”, “Research libraries for drawing simple geometric shapes”, etc.)

The summary for this issue is too terse. A good rule of thumb is that you should be writing at least one full paragraph in your summary (see https://uchicago-cs.github.io/cmsc22000/project/policies/#a-descriptive-summary for examples). If this issue was generated from a Scrum board card, please remember that the summary you write in the Sprint meeting is meant to be a quick draft; it needs to be fleshed out when you convert the card to an issue.

PR #635, opened 2020-05-24
Implementation of the AI Module

nmartinez233:
The AI module allows the player to battle with a computer, with the computer returning moves based on the game's difficulty (either returning the move dealing the least damage, most damage, or a random move). The primary function in the AI module is give_move(), which calls other functions as necessary to return the move based on difficulty. In order to test these functions for accuracy, I also created a sandbox enemy and player, along with new_move() and new_item() which allocates and initializes moves and items respectively.


tests/battle/test_AI.c
```diff
@@ -0,0 +1,331 @@
...
+}
+
+/* Creates example hardcoded moves for the enemy*/
+move_t* create_enemy_moves()
+{
+    move_t *head, *earthquake, *poke, *rock_throw;
+    head = NULL;
+    earthquake = create_move(1, NULL, true, 100, 20);
+	poke = create_move(2, NULL, true, 1, 1);
+	rock_throw = create_move(3, NULL, true, 300, 20);
```
Isaaq-Khader, line 116: For the attack stat, the pokemon formula usually has damage values that range from 0 - 150. Having a move that does 300 damage is a lot! Also, moves usually don't have a defensive value unless they are meant to up the stats of the user. I would maybe put these as 0. 
My recommendation for the moves would be this:
`
earthquake = create_move(1, NULL, true, 100, 0);
poke = create_move(2, NULL, true, 40, 0);
rock_throw = create_move(3, NULL, true, 90, 0);`

tests/battle/test_AI.c
```diff
@@ -0,0 +1,331 @@
...
+}
+
+/* Creates example hardcoded stats for the player*/
+move_t* create_player_moves()
+{
+    move_t *head, *fire_blast, *punch, *burn;
+    head = NULL;
+    fire_blast = create_move(4, NULL, true, 100, 50);
+	punch = create_move(5, NULL, true, 20, 20);
+	burn = create_move(6, NULL, true, 1, 2);
```
Isaaq-Khader, line 130: Same thing as I mentioned above. Burn would be a move that would probably add a status effect which we can have for now but we would maybe want to figure out how exactly status effects are going to work.

tests/battle/test_AI.c
nmartinez233: The large difference between numbers was just for testing purposes and I forgot to change them to normal values - just fixed it!

tests/battle/test_AI.c
nmartinez233: Makes sense - I edited the move name to blaze_kick

include/battle/battle_structs.h
```diff
@@ -23,6 +23,7 @@ typedef struct item {
 
 /* moves stub */
 typedef struct move {
+    int ID;
```
Isaaq-Khader, line 26: Make ID to id so it's not confused with being a definition like #define MAX_NAME_LEN(50)

src/battle/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 add_library(battle
+            src/AI.c
```
Isaaq-Khader, line 2: src/battle_ai.c

src/battle/src/AI.c
```diff
@@ -0,0 +1,132 @@
+#include "battle/AI.h"
+#include <math.h>
+
+/* see AI.h */
```
Isaaq-Khader, line 4: change comments like these to become battle_ai.h

tests/battle/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 set(TEST_EXE test-battle)
 
 add_executable(${TEST_EXE}
+        test_AI.c
```
Isaaq-Khader, line 4: test_battle_ai.c

tests/battle/test_AI.c
```diff
@@ -0,0 +1,330 @@
+#include <criterion/criterion.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+#include "battle/AI.h"
```
Isaaq-Khader, line 5: #include "battle/battle_ai.h"

tests/battle/test_AI.c
```diff
@@ -0,0 +1,330 @@
...
+
+/* Creates + initializes a move*/
+move_t *create_move(int ID, item_t* item, bool attack, int damage, int defense)
```
Isaaq-Khader, line 8: int ID -> int id

tests/battle/test_AI.c
```diff
@@ -0,0 +1,330 @@
...
+ {
+     move_t* move = (move_t*) calloc(1, sizeof(move_t));
+
+     move->ID = ID;
```
Isaaq-Khader, line 12: move->id = id;

tests/battle/test_AI.c
```diff
@@ -0,0 +1,330 @@
...
+void check_give_move(int difficulty, combatant_t* player, combatant_t* enemy, move_t* expected)
+{
+    move_t *actual_move, *expected_move;
+    actual_move = give_move(difficulty, player, enemy);
+
+    cr_assert_eq(actual_move->ID, expected->ID, "give_move did not return expected move!");
+}
+
+/* Ensures give_move returns the easiest move when difficulty is 1*/
+Test(AI, give_move_easy)
```
Isaaq-Khader, line 196: change tests to be like Test(test_ai, ...)

nmartinez233:
> Just one small thing I realized. I think you should change the file names from AI to battle_ai so that it's categorized with us. Just don't forget to change it in the required spots. I'll made comments where it should be changed. Sorry to be picky but I think these changes will help :)

Ahh! Thanks for catching this, the changes make complete sense I completely overlooked it!

dwahme:
Before I don an in-depth review of the code, I want to discuss high-level module design.

Honestly, I don't think that providing functionality for finding the weakest possible move is going to be something that's going to be useful for game authors- generally, the (simplest) "easy" system for other games is just pulling a random move.

I think that we can simplify the module by providing classifying AI functions by their behavior- right now, this would look like "random" behavior (pick a random move) and "best" behavior (pick the move that does the most damage. We can then use that as our framework for providing an AI system.

Some advantages of doing it this way:
- Instead of limiting ourselves to easy, medium, and hard, we can open ourselves up to different and more complex AI implementations in the future (creating more avenues for work both in the next sprint and future iterations of chiventure)
- Our implementation of "best" behavior (named "hard") right now is actually technically "greedy" behavior- it picks the move that does the most damage at that point in time, but in the future there could be moves that allow stat changes to actually kill the opponent faster
- It gives the game author more transparency when choosing which AI to use- instead of having to pick between easy, medium, and hard (and look up what those actually mean for chiventure), they can more easily see "random", "greedy", "best", "semi-best" (50% chance to be random, 50% chance to be best), and more AI strategies

With that in mind, I propose the following changes:
- remove all the easy AI stuff (side effect or making the PR smaller which is never bad)
- rename the medium AI stuff to random
- rename the hard AI stuff to either greedy or best

@nmartinez233 before we make any changes, I'd like to hear your thoughts on the above

nmartinez233:
> First off, good job so far on this- it's good work. Before I do an in-depth technical review of the code, I want to discuss high-level module design for a bit.
> 
> I think that we can simplify and future-proof the module by providing classifying AI functions by their behavior- right now, this would look like "random" behavior (pick a random move) and "best" behavior (pick the move that does the most damage). We can then use that as our framework for providing an AI system.
> 
> I don't think that providing functionality for finding the weakest possible move is going to be something that's going to be super useful for game authors- generally, the (simplest) "easy" system for other games is just pulling a random move.
> 
> Some advantages of doing it this way:
> 
> * Instead of limiting ourselves to easy, medium, and hard, we can open ourselves up to different and more complex AI implementations in the future (creating more avenues for work both in the next sprint and future iterations of chiventure)
> * Our implementation of "best" behavior (named "hard") right now is actually technically "greedy" behavior- it picks the move that does the most damage at that point in time, but in the future there could be moves that allow stat changes to actually kill the opponent faster
> * It gives the game author more transparency when choosing which AI to use- instead of having to pick between easy, medium, and hard (and look up what those actually mean for chiventure), they can more easily see "random", "greedy", "best", "semi-best" (50% chance to be random, 50% chance to be best), and more AI strategies. Note that some of those, I'm just spitballing on- we don't actually need to implement them yet
> * We'll have more freedom to implement more AI strategies in the future
> 
> With that in mind, I propose the following changes:
> 
> * remove all the easy AI stuff (side effect or making the PR smaller which is never bad)
> * rename the medium AI stuff to random
> * rename the hard AI stuff to either greedy or best
> 
> @nmartinez233 before we make any changes, I'd like to hear your thoughts on the above

You do make some valid points; one of the main uncertainties regarding the AI module was how it was exactly going to integrate with chiventure since I didn't even know if it has a difficulty setting.  I think when we were designing this a couple of weeks back I didn't completely think this part of it through. If I'm understanding this correctly, then we would no longer have the give_move() function which takes in a difficulty and instead we'd only have the functions that find the "best" or "random" moves (with more to add later) and those functions would be called directly when a move is needed?

dwahme:
I think that `give_move()` can stay- that should take in an enum containing `BATTLE_AI_BEST` and `BATTLE_AI_RANDOM` (to be expanded with more options later), and call the appropriate functions (so, very similar to how it is now).

Regardless, this function will need to exist somewhere- I think the most obvious option is this AI module, but I could see an argument for putting it in battle_logic or battle_flow instead

nmartinez233:
Sounds good - should I go ahead and make the changes now?

dwahme:
Yes, please do. I'm going to rerequest @Isaaq-Khader on the PR as well, although there shouldn't be tons of changes, mostly deletion and renaming

nmartinez233:
The changes to the PR are up!

tests/battle/test_battle_ai.c
```diff
@@ -0,0 +1,259 @@
...
+item_t* create_player_items()
+{
+    item_t *head, *dagger, *tea_leaves, *medicine;
+    head = NULL;
+    dagger = create_item(1, 1, 20, "A hearty dagger sure to take your breath away... for good",
+	true, 20, 5, 0);
+	tea_leaves = create_item(2, 1, 1, "Make yourself a warm cup of tea to heal your wounds!", true,
+	0, 0, 10);
+	medicine = create_item(3, 1, 1, "A first aid kit, straight from your doctor!", true, 0, 0, 30);
+	DL_APPEND(head, dagger);
```
dwahme, line 85: There are tab characters here. Replace these with spaces

tests/battle/test_battle_ai.c
```diff
@@ -0,0 +1,259 @@
...
+item_t* create_enemy_items()
+{
+    item_t *head, *mace, *diamond_sword, *force_shield;
+    head = NULL;
+    mace = create_item(4, 1, 20, "Temporary blindness leaves you quite vulnerable...", true, 
+        0, -30, 0);
+	diamond_sword = create_item(5, 1, 50, "Brings quick death to those who dare battle you...", 
+        true, 20, 0, 0);
+	force_shield = create_item(6, 1, 30, "Rest comfortably as this shield protects you for 1 move", 
+        true, 0, 30, 5);
```
dwahme, line 101: There are tab characters here. Replace these with spaces

tests/battle/test_battle_ai.c
```diff
@@ -0,0 +1,259 @@
...
+}
+
+/* Creates example hardcoded moves for the enemy*/
+move_t* create_enemy_moves()
+{
+    move_t *head, *earthquake, *poke, *rock_throw;
+    head = NULL;
+    earthquake = create_move(1, NULL, true, 100, 0);
+	poke = create_move(2, NULL, true, 40, 0);
+	rock_throw = create_move(3, NULL, true, 90, 0);
```
dwahme, line 115: There are tab characters here. Replace these with spaces

tests/battle/test_battle_ai.c
```diff
@@ -0,0 +1,259 @@
...
+}
+
+/* Creates example hardcoded stats for the player*/
+move_t* create_player_moves()
+{
+    move_t *head, *fire_blast, *punch, *blaze_kick;
+    head = NULL;
+    fire_blast = create_move(4, NULL, true, 100, 0);
+	punch = create_move(5, NULL, true, 20, 0);
+	blaze_kick = create_move(6, NULL, true, 60, 0);
```
dwahme, line 129: There are tab characters here. Replace these with spaces

src/battle/src/battle_ai.c
```diff
@@ -0,0 +1,95 @@
...
+    {
+        if (random_move->next == NULL)
+        {
+            return random_move;
+        }
+        random_move = random_move->next;
+    }
+
+    return random_move;
+}
```
dwahme, line 49: A more straightforward way to accomplish this would be `DL_FOREACH`, there's a bit of funky sidestepping we're doing with adding 1 to random, and it would be safer to default to NULL if no move is found. Something like this would be better
```
move_t* find_random(combatant_t* player, combatant_t* enemy)
{
    int count;
    move_t *tmp;
    move_t *random_move = NULL;

    DL_COUNT(enemy->moves, tmp, count);
    int idx = rand() % count;

    int i = 0;
    DL_FOREACH(enemy->moves, tmp)
    {
        if (i == idx)
        {
            random_move = tmp;
        }

        i++;
    }

    return random_move;
}
```

src/battle/src/battle_ai.c
```diff
@@ -0,0 +1,113 @@
...
+    for (i = 0; i < random - 1; i++)
+    {
+        if (random_move->next == NULL)
+        {
+            return random_move;
+        }
+        random_move = random_move->next;
+    }
+
+    return random_move;*/
```
dwahme, line 66: PR's should never have commented out code

dwahme:
Issue Score: ✔️+

Comments:

PR #647, opened 2020-05-25
Adding Logic to Battle

Isaaq-Khader:
This pull request adds components to the battle team that allows for game rules to be in place. By merging this, it will add the following functions check_target, battle_over, goes_first, find_item, consume_item, player_use_item, and award_xp. These allow for functionality in the battle. This does not add any new structs but some of these functions have code values for the future if we want to give a custom message according to what occurred. For example, in player_use_item, it returns a value depending on anything that can occur like an empty inventory or not having any more of that desired item.


include/battle/battle_logic.h
```diff
@@ -0,0 +1,89 @@
...
+Function detemines what player goes first
+ *
+ *  Parameters: 
+ *  - difficulty: difficulty level of the game
+ *  Returns:
+ *  -1 - invalid input
+ *  0 - user goes first
+ *  1 - enemy goes first    
+ */
+int goes_first(double p_speed, double e_speed);
```
celiaanderson, line 44: parameters in header comment doesn't match

include/battle/battle_logic.h
```diff
@@ -0,0 +1,89 @@
...
+
+/* 
+ * Finds the desired item to be used
+ * 	Parameters:
+ *   inventory - an player or enemy's inventory
+ * 	 ID - the ID number of the desired item
+ * 	Returns:
+ * 	 temp->item - pointer to the desired item 
+ */
+item_t *find_item(item_t *inventory, int ID);
```
celiaanderson, line 54: the return value "temp->item" in the header comment seems to refer to local variable names in the implementation; i think removing "temp->item" could just make this more comment clear

celiaanderson, line 54: according to the style guide, it looks like the variable name "ID" should be lowercase unless it's a constant like MAX_NAME_LEN

include/battle/battle_logic.h
```diff
@@ -0,0 +1,89 @@
...
+ *  Parameters: 
+ *   c - combatant information
+ *   inv - inventory of the player
+ *   ID - the ID of the item
+ *  Returns:
+ *   0 - successfully used item
+ *   1 - there is nothing in the inventory
+ *   2 - item does not exist or the player does not have any more
+ */
+int player_use_item(combatant_t *c, item_t *inv, int ID);
```
celiaanderson, line 78: lowercase for ID

src/battle/src/battle_logic.c
```diff
@@ -0,0 +1,107 @@
...
+        return 1;
+    }
+    else
+    {
+        return -1;
+    }
+}
+
+/* see logic.h */
+item_t *find_item(item_t *inventory, int ID)
```
celiaanderson, line 57: lowercase for ID

src/battle/src/battle_logic.c
```diff
@@ -0,0 +1,107 @@
...
+int consume_item(combatant_t *c, item_t *item)
+{
+    c->stats->hp += item->hp;
+    c->stats->strength += item->attack;
+    c->stats->defense += item->defense;
+    return 0;
+}
+
+/* see logic.h */
+int player_use_item(combatant_t *c, item_t *inv, int ID)
```
celiaanderson, line 82: lowercase for ID

include/battle/battle_logic.h
```diff
@@ -0,0 +1,89 @@
+#ifndef BATTLE_LOGIC_H
+#define BATTLE_LOGIC_H
+
+#include <stdlib.h>
+#include <stdio.h>
+#include "battle/battle_structs.h"
+#include "battle/battle_state.h"
```
celiaanderson, line 7: from the comments on my last PR, the include statements should be in alphabetical order (but with the standard libraries first) so something like
`#include <stdio.h>`
`#include <stdlib.h`
`#include "battle/battle_state.h"`
`#include "battle/battle_structs.h"`

src/battle/src/battle_logic.c
```diff
@@ -0,0 +1,107 @@
+#include "battle/battle_logic.h"
+#include "common/utlist.h"
+
+/* check logic.h */
```
celiaanderson, line 4: change the function comments in this file to say "battle_logic.h" instead of "logic.h"

include/battle/battle_logic.h
celiaanderson: also, it may be better if this function returned turn_t from battle_structs.h (the enum with the options PLAYER and ENEMY) - it may make the implementation in the c file more readable even though it looks like the actual return values of the function match the values of PLAYER (0) and ENEMY (1)

tests/battle/test_battle_logic.c
```diff
@@ -0,0 +1,362 @@
+#include <criterion/criterion.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "battle/battle_state.h"
+#include "battle/battle_structs.h"
+#include "battle/battle_logic.h"
+
```
celiaanderson, line 9: alphabetical order

tests/battle/test_battle_logic.c
```diff
@@ -0,0 +1,362 @@
...
+Test(logic, target_exists)
```
celiaanderson, line 10: change logic to battle_logic for all the tests in the file

celiaanderson: add comment for test here

tests/battle/test_battle_logic.c
```diff
@@ -0,0 +1,362 @@
...
+
+    bool res = check_target(b, "Orc John");
+    printf("      CHECK_TARGET RETURNED: %d\n", res);
+
+    cr_assert_eq(res, true, "check_target() failed!");
+
+    battle_free(b);
+}
+
+Test(logic, target_does_not_exist)
```
celiaanderson, line 38: add comment for test

tests/battle/test_battle_logic.c
```diff
@@ -0,0 +1,362 @@
...
+
+    int res = goes_first(p->stats->speed, e->stats->speed);
+
+    cr_assert_eq(res, 0, "goes_first() failed!");
+
+    combatant_free(p);
+    combatant_free(e);
+}
+
+Test(logic, find_item)
```
celiaanderson, line 210: add comments for rest of tests in this file

include/battle/battle_logic.h
Isaaq-Khader: Ah that one slipped past me I'll change the header now thank you!
I agree with that, I'll change it so that it returns the turn struct we have in place.

include/battle/battle_logic.h
Isaaq-Khader: Got it! I'm removing the temp->item now from the comments

Isaaq-Khader: Ah good catch. I'm changing that now!

tests/battle/test_battle_logic.c
Isaaq-Khader: Oh jeez I thought I had it all in alphabetical order. I might need to look back at the alphabet for a bit. Thanks for the catch!

Isaaq-Khader:
I made the changes Celia mentioned and everything runs as intended. The commits give detail on what exactly I changed to address these changes.

Isaaq-Khader:
I just completed the merge! Everything passes according to Travis!

include/battle/battle_logic.h
```diff
@@ -0,0 +1,87 @@
...
+ * Parameters:
+ * - p: information about the player
+ * - e: information about the enemy
+ * Returns:
+ * - 0 for game still in play
+ * - 1 for game over due to enemy
+ * - 2 for game over due to player
+ * - -1 for an error with not passing conditions
+ */
+int battle_over(combatant_t *p, combatant_t *e);
```
dwahme, line 32: It would be better to define an enum with constants `BATTLE_ERROR`, `BATTLE_IN_PROGRESS`, `BATTLE_VICTOR_PLAYER`, and `BATTLE_VICTOR_ENEMY` or something. Using int codes means you have magic numbers that float around where the meaning isn't immediately clear

dwahme: Also, from the way this function is named/documented, I would expect it to take in a `battle_t*` instead of 2 combatants

include/battle/battle_logic.h
```diff
@@ -0,0 +1,87 @@
...
+
+/* 
+ * Finds the desired item to be used
+ * 	Parameters:
+ *   inventory - an player or enemy's inventory
+ * 	 id - the id number of the desired item
+ * 	Returns:
+ * 	 returns a pointer to the desired item 
+ */
+item_t *find_item(item_t *inventory, int id);
```
dwahme, line 52: I'm pretty sure game_state's `item_t` uses a `char*` as the item id. While I don't think we don't need to make this change to the mock struct, we should keep this in mind. Also, when we switch it over, we'll have lookup through the `item_t` hashtable given to us with 
```
HASH_FIND_STR(item_hash_table, item_name, dummy_item_pointer)
```

include/battle/battle_logic.h
```diff
@@ -0,0 +1,87 @@
...
+ *  Parameters: 
+ *   c - combatant information
+ *   inv - inventory of the player
+ *   id - the number id of the item
+ *  Returns:
+ *   0 - successfully used item
+ *   1 - there is nothing in the inventory
+ *   2 - item does not exist or the player does not have any more
+ */
+int player_use_item(combatant_t *c, item_t *inv, int id);
```
dwahme, line 76: You generally want to stay away from int codes. Also 1 and 2 seem like the same thing; it might be better to just use `SUCCESS` or `FAILURE` here

include/battle/battle_logic.h
```diff
@@ -0,0 +1,87 @@
...
+
+/* 
+ * Function detemines if the player goes first
+ *
+ *  Parameters: 
+ *  - b: information about the battle
+ *  Returns:
+ *  PLAYER or ENEMY which shows who goes first
+ */
+turn_t goes_first(battle_t *b);
```
dwahme, line 42: Wouldn't it be better just to return a pointer to the fastest combatant, or even better have this return a linked list of combatants sorted by speed? What are your thoughts on this?

(FYI- utlist provides a sorting macro `DL_SORT` that would be useful for the second idea)

include/battle/battle_logic.h
dwahme: Also, wouldn't it make more sense to have `combatant_t` have an inventory associated with it (reducing the number of arguments in this function)?

src/battle/src/battle_state.c
```diff
@@ -50,10 +51,20 @@ int combatant_init(combatant_t *c, char *name, bool is_friendly, stat_t *stats,
 /* battle_state.h */
 int combatant_free(combatant_t *c)
 {
-	assert(c != NULL);
+	if (c == NULL)
```
dwahme, line 54: No tabs

include/battle/battle_logic.h
Isaaq-Khader: That sounds like a really great idea! I'm starting to realize my code is based on a 1v1 situation which I have to make a few adjustments here. I like your second idea of a linked list based on speed. How can I go about implementing this?

include/battle/battle_logic.h
Isaaq-Khader: This would make sense. I'll get on changing these.

include/battle/battle_logic.h
Isaaq-Khader: Oh got it so then should I just change it from `int id` -> `char* id`?

Isaaq-Khader: Or is there more to this function that should be changed?

include/battle/battle_logic.h
Isaaq-Khader: Got it! This makes sense, I'm making an enum with the constants you mentioned and am rewriting battle_over to meet that

src/battle/src/battle_state.c
Isaaq-Khader: Sorry, I'm not too sure about the issue. Does the if statement have incorrect indentation?

include/battle/battle_logic.h
dwahme: This is more of a heads up than anything else; I don't think a specific change for this PR is needed, but keep this in mind as we integrate in sprint 4

include/battle/battle_logic.h
dwahme: You'll need to use `DL_SORT`. Check out the utlist documentation first- https://troydhanson.github.io/uthash/utlist.html

Relevant stuff
> DL_SORT(head,cmp);
> cmp: pointer to comparison function which accepts two arguments-- these are pointers to two element structures to be compared. The comparison function must return an int that is negative, zero, or positive, which specifies whether the first item should sort before, equal to, or after the second item, respectively. (In other words, the same convention that is used by strcmp). Note that under Visual Studio 2008 you may need to declare the two arguments as void * and then cast them back to their actual types.

The cmp function would basically just need to return the difference of the speed of the 2 combatant_t's. I can provide technical support as necessary if you're having trouble implementing it- ping me on slack if you need me; using function pointers in sorting functions is doable the first time you do it but can be a headache if you don't know what's up

src/battle/src/battle_state.c
dwahme: Oh yeah, to be more clear, there's a tab character there, and it should be 4 spaces instead

include/battle/battle_logic.h
Isaaq-Khader: Gotcha! Sounds good!

src/battle/src/battle_state.c
Isaaq-Khader: Done!

include/battle/battle_logic.h
Isaaq-Khader: I am making this into an issue for the future or something I can get back to. I'll write up a card for this in the backlog but we'll keep it as it is for now. I guess we can see this function as if one of the enemies is faster than the enemy, then all of the enemy's buddies can attack with him.

src/battle/src/battle_logic.c
```diff
@@ -0,0 +1,103 @@
...
+    combatant_t *temp;
+    DL_FOREACH(b->enemy, temp)
+    {
+        if (strncmp(temp->name, target, MAX_NAME_LEN) == 0)
+        {
+            return true;
+        }
+    }
+    return false;
+}
```
dwahme, line 16: Since we don't have a function for selecting a target, it would be just as effective to have this return the pointer to the combatant instead of a boolean (and then NULL stands for false, and non-NULL stands for true

tests/battle/test_battle_logic.c
```diff
@@ -0,0 +1,476 @@
...
+	stat_t *estats1 = calloc(1, sizeof(stat_t));
+    estats1->hp = 12;
+    stat_t *estats2 = calloc(1, sizeof(stat_t));
+    estats2->hp = 13;
+
+    combatant_t *p = combatant_new("Player", true, pstats, NULL, NULL);
+    combatant_t *e = combatant_new("Enemy", false, estats, NULL, NULL);
+    combatant_t *e1 = combatant_new("Enemy", false, estats1, NULL, NULL);
+    combatant_t *e2 = combatant_new("Enemy", false, estats2, NULL, NULL);
+   	cr_assert_not_null(p, "combatant_new() failed");
```
dwahme, line 165: There's a tab character here

include/battle/battle_logic.h
```diff
@@ -0,0 +1,87 @@
...
+
+/* 
+ * Finds the desired item to be used
+ * 	Parameters:
+ *   inventory - an player or enemy's inventory
+ * 	 id - the id number of the desired item
```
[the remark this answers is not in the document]
Isaaq-Khader: Quick question for the future, how do you tell when there are tabs instead of spaces?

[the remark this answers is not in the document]
dwahme: Because the indentation was a little wonky. If you try to highlight it, you'll see that the highlight jumps all at once, instead of space by space

Isaaq-Khader:
This PR focused on bringing in functions that allow a game to run such as who goes first and using an item as the player or enemy sees fit. It brings logic to the game that allows the battle to continuously run.

dwahme:
Issue Score: ✔️++

Comments:
Great job! Well done specifying exactly what this PR does and doesn't affect

Issue #609, opened 2020-05-18
Design/Implement a global data structure for game stats


carlosazpurua1004:
Aftering meeting with Professor Borja, he explained that some stats might contain global properties (like global max values), so it would be a good idea to implement a global table of all stats in a game written in chiventure. 
This table would keep track of such global properties for every stat, and whether or not a stat is in the game at all, since in our current implementation of stats, different players/npcs can have different stats.

Issue will be resolved by a pull request with the data structure written, and possibly some stubbed functions if time permits.

richardxd:
Updates on stats_global struct 
e292496
eee554c

richardxd:
The bugs have been removed after I figured out how to make Chiventure. The stubbed out functions in stats.h corresponds to stats.c right now ,and they are ready to be implemented.

carlosazpurua1004:
The struct and stubs of functions have been merged in https://github.com/uchicago-cs/chiventure/pull/653

ecoble:
Issue Score: ✔️++
Comments:
Great work, especially with your summary!

Issue #591, opened 2020-05-15
Implement and write tests for add_stat and display_stat


sumedhsarwate:
The add_stat and display_stat functions have already been stubbed, and I will implement the functions such that add_stat adds a stat to the player's stats hash table and display_stat returns a list of the stats the player has. I will also write tests for these functions.

sumedhsarwate:
As per our team meeting, add_stat_player (renamed to avoid confusion with the global add_stat function) will now take in a pointer to a global stats struct as a parameter instead of the string name of the stat. The player stats struct is also going to be updated so that it will contain a pointer to a global stat instead of a string for its name.

sumedhsarwate:
Now that our previous pull request has been approved, I've created a new branch to implement the add and display functions.

richardxd:
I have created a new rpg-stat branch that includes the test files. 23dca317a29370767069421dc03fc6b30aae4a01
This could result in a merge conflict if the branch you created also contains test folder/files. Please check this out before you started working on the test!

sumedhsarwate:
@richardxd and I fixed the conflict between the branches so that we have the same folder/files for the tests

sumedhsarwate:
In order for the tests to pass and to follow the UTHash documentation, I had to change the parameter of add_stat_player to take in a pointer to a pointer for the hash table. The updated header is:

int add_stat_player(stats_hash_t **sh, stats_t *s)

sumedhsarwate:
Closing Statement: This issue is being closed by a pull request that includes the implementation and tests for both add_stat and display_stat. We've updated the stats struct so that the key is a string, and the implementation for add_stat has been updated accordingly (the changes can be seen in commit 2ac017a).

ecoble:
Issue Score: ✔️++
Comments:
Great work, especially with the frequent updates!

PR #653, opened 2020-05-26
RPG-playerstats: global stats and add/display functions

sumedhsarwate:
We've implemented the global stats struct and the add_stat_player and display_stat functions and tests. Addresses issues #591 and #609.


carlosazpurua1004:
If you look on the travis output
https://travis-ci.com/github/uchicago-cs/chiventure/builds/168145141
You can see that they didn't actually run the rpg-stats tests (I think).

sumedhsarwate:
@carlosazpurua1004 the rpg-stat tests are a part of the game-state tests (the tests for add and display are 2 out of the 101 tests that were run)

include/game-state/stats.h
```diff
@@ -3,10 +3,27 @@
...
+ /* This struct represents the global table that keeps track of all stats available.
+  * It contains:
+  *      the name of the stat,
+  *      which is also the key to the hashtable
+  *
+  *      the maximal value a stat could have
+  * */
+typedef struct stats_global{
+    char *name;
+    double max;
```
ecoble, line 16: If the global stats has a max, should player stat also have a max? What is the difference between these two?

include/game-state/stats.h
```diff
@@ -169,14 +215,25 @@ int add_stat_player(stats_hash_t *sh, stats_t *s);
...
+ *  SUCCESS on success, FAILURE if an error occurs.
+ */
+int free_stats(stats_t *stat);
+
+/*
+ * Frees a global stat
  *
  * Parameters: 
- * sh: pointer to the stats hash table to be freed
+ * gsh: pointer to the global statto be freed
```
ecoble, line 232: The parameter is called stat, also need a space between "stat" and "to"

src/game-state/src/stats.c
```diff
@@ -43,17 +43,42 @@ double get_stat_mod(stats_hash_t *sh, char *stat)
...
+    return SUCCESS;
 }
 
 /* See stats.h */
 char* display_stats(stats_hash_t *s)
 {
-    printf("display_stats: function not yet implemented\n");
-    return "0"; // still needs to be implemented
+    stats_t *stat;
+    int size = 2 + (20 * HASH_COUNT(s));
```
ecoble, line 65: Create macros for 2 and 20, no magic numbers

include/game-state/stats.h
```diff
@@ -66,30 +83,59 @@ typedef struct effects{
...
+ *
+ * Parameters:
+ * name: the unique string ID to be given to the stat
+ * max: maximal value this stat could have
+ * 
+ * Returns:
+ *  Pointer to allocated global stats struct
+ */
+
+stats_global_t* stats_global_new(char *name, double max);
```
ecoble, line 126: move *

tests/game-state/test_stats.c
```diff
@@ -0,0 +1,70 @@
...
+#include <stdbool.h>
+#include <stdio.h>
+#include "game-state/stats.h"
+
+/* Checks that add_stat_player correctly adds a new stat 
+   to a hash table*/
+Test (stats, add_stat_player)
+{
+    stats_hash_t *sh = NULL;
+    stats_global_t g;
```
ecoble, line 12: Once stats_global_new gets implemented, you should change this test to create a global stat using that function

include/game-state/stats.h
sumedhsarwate: I believe the global stats max is the max value a stat could ever have, while the player stat max is the current max for the player/npc and it could be dependent on things like class or level, but it'll never be greater than the global max.

sumedhsarwate:
@ecoble @carlosazpurua1004 I just realized that the header comment for display_stats says the function should return the names of the stats along with their values/modifiers, but I currently have it as only returning a list of their names. Should I change the function so it returns a string in a format like "stat name: value, modifier\n" or leave it as is?

ecoble:
> @ecoble @carlosazpurua1004 I just realized that the header comment for display_stats says the function should return the names of the stats along with their values/modifiers, but I currently have it as only returning a list of their names. Should I change the function so it returns a string in a format like "stat name: value, modifier\n" or leave it as is?

It's up to you, I think leaving it is fine since I'm not sure how much this function will end up being used anyways

include/game-state/stats.h
```diff
@@ -3,21 +3,38 @@
...
 
+// GLOBAL STATS STRUCT DEFINITION ----------------------------------------------------
+ /* This struct represents the global table that keeps track of all stats available.
+  * It contains:
+  *      the name of the stat,
+  *      which is also the key to the hashtable
+  *
+  *      the global maximal value a stat could have
+  * */
+typedef struct stats_global{
```
ecoble, line 14: Space before brace

include/game-state/stats.h
```diff
@@ -33,7 +50,7 @@ typedef struct stats stats_hash_t;
  * 
  *      the modifier of the effect on that stat 
  * 
- *      the duration of the effect, an umber 
+ *      the duration of the effect, an number
```
ecoble, line 53: nit: a number

ecoble: Although really it would be better to say "an int"

include/game-state/stats.h
```diff
@@ -66,30 +83,59 @@ typedef struct effects{
...
+ */
+
+stats_global_t *stats_global_new(char *name, double max);
 
 /*
  * Allocates a new stat
  *
  * Parameters:
- * stats_name: the unique string ID to be given to the stat
+ * stats_name: the pointer to the global stat struct.
```
ecoble, line 132: The parameter is named "stat"

include/game-state/stats.h
```diff
@@ -155,7 +201,7 @@ double get_stat_mod(stats_hash_t *sh, char *stat);
  * Returns:
  *  SUCCESS on success, FAILURE if an error occurs.
  */
-int add_stat_player(stats_hash_t *sh, stats_t *s);
+int add_stat_player(stats_hash_t **sh, stats_t *s);
 
 /*
  * Print the stats in a hashtable and their values/modifiers
```
ecoble, line 207: As you commented, this can be changed to just print the stat names

include/game-state/stats.h
```diff
@@ -169,14 +215,25 @@ int add_stat_player(stats_hash_t *sh, stats_t *s);
 char *display_stats(stats_hash_t *sh);
 
 /*
- * Frees a stats hash table
+ * Frees a stat
+ *
+ * Parameters: 
+ * sh: pointer to the stat to be freed
```
ecoble, line 221: The parameter is named stat

include/game-state/stats.h
```diff
@@ -169,14 +215,25 @@ int add_stat_player(stats_hash_t *sh, stats_t *s);
...
  *
  * Parameters: 
- * sh: pointer to the stats hash table to be freed
+ * stat: pointer to the global stat to be freed
  * 
  * Returns:
  *  SUCCESS on success, FAILURE if an error occurs.
  */
-int free_stats(stats_hash_t *sh);
+int free_stats_global(stats_global_t* stat);
```
ecoble, line 237: move * to be stats_global_t *stat

src/game-state/src/stats.c
```diff
@@ -1,14 +1,16 @@
 #include "game-state/stats.h"
+#define MIN_STRING_LENGTH 2
+#define MAX_NAME_LENGTH 20
```
ecoble, line 3: Consider upping this to 40 or 50, just to have a buffer

sumedhsarwate:
I'm going to keep the implementation for display_stat as it is, so I've updated the header comment for it and I've addressed the other comments as well.

include/game-state/stats.h
```diff
@@ -169,14 +215,25 @@ int add_stat_player(stats_hash_t *sh, stats_t *s);
...
  *
  * Parameters: 
- * sh: pointer to the stats hash table to be freed
+ * stat: pointer to the global stat to be freed
  * 
  * Returns:
  *  SUCCESS on success, FAILURE if an error occurs.
  */
-int free_stats(stats_hash_t *sh);
+int free_stats_global(stats_global_t *stat);
```
carlosazpurua1004, line 237: Did we ever reach concesus on whether the free functions should free the whole table or just individual stats? I'm still unsure of a situation where you want to free an individual stat, and not the whole table, so we might as well do that?

src/game-state/src/stats.c
```diff
@@ -1,14 +1,16 @@
...
-int stats_init(stats_t *s, char *stats_name, double init)
+int stats_init(stats_t *s, stats_global_t *stats_name, double init)
 {
     printf("stats_init: function not yet implemented\n");
     return 0; // still needs to be implemented
 }
 
 /* See stats.h */
-stats_t *stats_new(char *stats_name, double init)
+stats_t *stats_new(stats_global_t *stats_name, double init)
```
carlosazpurua1004, line 13: I made other changes to the headers of these functions while writing them, I think it would be best to just undo changes here to avoid merge conflicts since these changes since they don't affect the rest of the code.

include/game-state/stats.h
```diff
@@ -66,30 +83,59 @@ typedef struct effects{
...
  *  s: A stats struct Must point to already allocated memory.
- *  stats_name: the unique string ID of the stat
+ *  stat: the pointer to the global stat struct
  *  init: starting value
  *
  * Returns:
  *  SUCCESS on success, FAILURE if an error occurs.
  */
-int stats_init(stats_t *s, char *stats_name, double init);
+int stats_init(stats_t *s, stats_global_t *stat, double init);
```
carlosazpurua1004, line 113: See later comment about header.

include/game-state/stats.h
```diff
@@ -66,30 +83,59 @@ typedef struct effects{
...
  * Parameters:
- * stats_name: the unique string ID to be given to the stat
+ * stat: the pointer to the global stat struct.
  * init: starting value
  * 
  * Returns:
  *  Pointer to allocated stats struct
  */
-stats_t *stats_new(char *stats_name, double init);
+stats_t *stats_new(stats_global_t *stat, double init);
```
carlosazpurua1004, line 138: See later comment about header.

src/game-state/src/stats.c
```diff
@@ -43,17 +43,42 @@ double get_stat_mod(stats_hash_t *sh, char *stat)
...
 
 /* See stats.h */
-int add_stat_player(stats_hash_t *sh, stats_t *s)
+int add_stat_player(stats_hash_t **sh, stats_t *s)
 {
-    printf("add_stat: function not yet implemented\n");
-    return 0; // still needs to be implemented
+    stats_t *check;
+    
+    HASH_FIND(hh, *sh, s->global, sizeof(stats_global_t), check);
```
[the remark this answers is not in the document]
carlosazpurua1004: nvm

include/game-state/stats.h
sumedhsarwate: I'll change it to freeing the entire table for now, but it might be good to add a function for freeing the individual stat later

sumedhsarwate:
I've updated the stats struct so that we now have a string key and I've updated the add_stat functions and the tests accordingly.

sumedhsarwate:
Summary: The global stats struct was added to dev along with the add_stat and display_stat functions and their tests. The stats struct has also been updated so that its key is a string instead of a pointer to a global stats struct.

ecoble:
Issue Score: ✔️++
Comments:
Great work, especially with the discussions on this PR!

PR #656, opened 2020-05-26
Adding Stats and effects to NPC's

carlosazpurua1004:
- added stat and effect as field to NPC struct
- initialize them as NULL in NPC_init
- removed health and xp fields and associated functions

- changed name of function "add_stat_player" to the more appropriate "add_stat"

### DIFF
--- a/include/game-state/npc.h
+++ b/include/game-state/npc.h
@@ -11,8 +11,8 @@ typedef struct npc
 {
     /* hh is used for hashtable, as provided in uthash.h */
     UT_hash_handle hh;
-    char *npc_id;
-    int health;
+    stats_t *stats; 
+    stat_effect_t *effects;
     convo_t *dialogue; // placeholder for incoming dialogue module
     item_hash_t *inventory;
 } npc_t;
@@ -36,11 +36,10 @@ typedef struct npcs_in_room {
 typedef struct npcs_in_room npcs_in_room_hash_t;
 
 /*
- * Initializes an npc with given health.
+ * Initializes an npc (currently with an empty stats/effect table)
  *
  * Parameters:
  *  npc: an npc; must point to already allocated memory
- *  health: the starting health of the npc
  *  npc_id: string referring to npc id; passed implicitly
             from npc_new 
  *  dialogue: pointer to a convo struct for the npc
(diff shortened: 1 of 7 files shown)


carlosazpurua1004:
Apperantly NPC code I changed wasn't supposed to be in Dev, the actual NPC code is elsewhere, so i'll have to find and change that code.

include/game-state/npc.h
```diff
@@ -11,8 +11,8 @@ typedef struct npc
 {
     /* hh is used for hashtable, as provided in uthash.h */
     UT_hash_handle hh;
-    char *npc_id;
-    int health;
+    stats_t *stats;
```
ecoble, line 14: As already noted in the comment, these changes should go in the NPC module

src/npc/src/npc.c
```diff
@@ -1,26 +1,27 @@
...
     npc->inventory = NULL;
+    npc->stats = NULL;
+    npc->effects = NULL;
 
     return SUCCESS;
 }
 
 /* See npc.h */
-npc_t* npc_new(char *npc_id, int health)
+npc_t* npc_new(char *npc_id)
```
sumedhsarwate, line 17: I think the * needs to be moved

include/npc/npc.h
```diff
@@ -3,14 +3,16 @@
...
 //TODO-#include "dialogue.h"
 
 /* A non-playable character in game */
 typedef struct npc {
     /* hh is used for hashtable, as provided in uthash.h */
     UT_hash_handle hh;
     char *npc_id;
-    int health;
+    stats_t *stats;
+    stat_effect_t *effects;
```
sumedhsarwate, line 15: These are hash tables right? Would it be helpful to change the type to stats_hash_t and effects_hash_t?

src/npc/src/npc.c
carlosazpurua1004: The return one or the parameter?

sumedhsarwate: the return one

carlosazpurua1004:
Going to make these changes on a different new branch to avoid merge conflicts etc.

PR #659, opened 2020-05-26
Custom actions/branch

emmamedeiros:
This branch contains the implementation for the branch block module, which includes new, init, and free functions. This is the final module for the custom-actions structs implementation.


tests/custom-actions/test_branch_block.c
```diff
@@ -0,0 +1,723 @@
+#include <criterion/criterion.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "branch_block.h"
+#include "game-state/item.h"
+#include "ast_block.h"
+
```
MaxineK36, line 7: These tests could use some quick block comments splitting up the sections and walking through what's going on (or, optionally, you could create a local helper to build your attributes)

There's no need to have tests for when num_controls and num_conditionals are >1, since your code at the moment doesn't do anything except transfer that information directly into the struct, so behavior is necessarily the same for 1 conditional as for 5.

emmamedeiros:
The branch block module has now been implemented, tested, and merged into dev which concludes the basic implementation of the block struct functions for custom-actions. Thus, this pull request will be closed and the custom-actions/branch branch will be deleted.

MaxineK36:
Issue Score: ✔️+

Comments:
Good work! One thing to improve for future PRs: The summary for this issue is too terse, and is difficult to understand without the context of the related PRs

PR #681, opened 2020-05-27
Implementation of autogenerate module; generates 1 or multiple rooms if the current room is a "dead end"

bryan22lee:
This PR is for the [autogenerate.h](https://github.com/uchicago-cs/chiventure/blob/openworld/autogen/include/openworld/autogenerate.h) module, which can be found in the [openworld/autogen](https://github.com/uchicago-cs/chiventure/tree/openworld/autogen) branch. Tests are written in [test_autogenerate.c](https://github.com/uchicago-cs/chiventure/blob/openworld/autogen/tests/openworld/test_autogenerate.c). This module has two main goals and relies on [gen_structs.h](https://github.com/uchicago-cs/chiventure/blob/dev/include/openworld/gen_structs.h) context struct: (1) to create a base for single-room autogeneration and (2) to extend the single-room autogeneration model to create simple multi-room generation function that generates _around_ the current room of the game. The two points above hold _only if_ the current room of the game is a "dead end", i.e. there are no outward paths out of the current room. The interface for this module has four functions (found in [autogenerate.h](https://github.com/uchicago-cs/chiventure/blob/openworld/autogen/include/openworld/autogenerate.h)). (1) any_paths() is used to determine "dead ends", (2) roomspec_to_room() does the work of generating a room based on given roomspec_t information, (3) and room_generate() determines if the current room is a dead end and generates/adds a room and a path; it connects a newly generated (singular) room to the current room of the game. The above three functions are responsible for single-room autogeneration. Finally, (4) multi_room_generate extends the room_generate to iterate through the entire speclist of the given gencontext_t struct to generate rooms _around_ the current room of the game. The intention is to allow for further exploration when a "dead end" is reached, and the hope is to scale this for openworld generation in the future. After this PR, following goals will be to merge this work to depend further on [default_rooms.h](https://github.com/uchicago-cs/chiventure/blob/dev/include/openworld/default_rooms.h) and [default_items.h](https://github.com/uchicago-cs/chiventure/blob/dev/include/openworld/default_items.h) for streamlined room generation that can actually be demonstrated.


src/openworld/src/autogenerate.c
```diff
@@ -0,0 +1,81 @@
...
+bool any_paths(room_t *r)
+{
+    assert(r != NULL && "any_paths: Given room is NULL");
+    return (r->paths != NULL);
+}
+
+/* See autogenerate.h */
+room_t* roomspec_to_room(game_t *game, roomspec_t *roomspec)
+{
+    room_t *res = room_new(roomspec->room_name, roomspec->short_desc, roomspec->long_desc);
```
navila-luna, line 28: It really depends on who merges first, but room spec_to_room would later include a hard coded mock NPC such as get_hostile_npcs().

bryan22lee: That makes sense; as of now, this module does not interact with the hard-coded structures, but I was planning to add dependence on default_rooms.h and default_items.h after this PR. I'll make a note for get_hostile_npcs() as well.

src/openworld/src/autogenerate.c
```diff
@@ -0,0 +1,81 @@
...
+
+    while (context->speclist != NULL) {
+        room_generate(game, context);
+
+        // Go to next roomspec in context->speclist
+        context->speclist = context->speclist->next;
+    }
+
+    // Restore the original head of context->speclist
+    context->speclist = speclist_head;
```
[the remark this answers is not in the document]
bryan22lee: Got it, that simplifies these lines a lot! I replaced that segment with the DL_FOREACH() usage

src/openworld/src/autogenerate.c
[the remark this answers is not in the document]
bryan22lee: To keep room_id field unique per room, I have added an extra parameter to roomspec_to_room(), room_generate(), and multi_room_generate() that allows the user of the function to input a unique room_id string for the to-be-generated room.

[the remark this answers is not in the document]
bryan22lee: Does this make sense? I was thinking that, in that case, room_id's are unique per room when generating a room, and roomspec->room_name is used as the key for hash table lookup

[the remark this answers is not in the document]
bryan22lee: Updated so a unique parameter char *room_id must also be given when generating a room

src/openworld/src/autogenerate.c
```diff
@@ -0,0 +1,81 @@
...
+{
+    assert(r != NULL && "any_paths: Given room is NULL");
+    return (r->paths != NULL);
+}
+
+/* See autogenerate.h */
+room_t* roomspec_to_room(game_t *game, roomspec_t *roomspec)
+{
+    room_t *res = room_new(roomspec->room_name, roomspec->short_desc, roomspec->long_desc);
+    res->items = roomspec->items;
```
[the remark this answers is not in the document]
bryan22lee: Thanks for the help. Used HASH_ITER() to iterate through the hashed items of the given roomspec, creating a deep copy of them for the items field (type item_hash_t*) of the generated room

src/openworld/src/autogenerate.c
```diff
@@ -0,0 +1,80 @@
...
+    if (!any_paths(game->curr_room)) 
+    {
+        // Adds one generated room from the head of the speclist only
+        room_t *new_room = roomspec_to_room(game, context->speclist->spec, room_id);
+
+        // Add addRoom to gameNew
+        assert(SUCCESS == add_room_to_game(game, new_room));
+        
+        // Add path from the current room to addRoom
+        path_t* path_to_room = path_new(new_room, context->speclist->spec->room_name);
```
dwahme, line 53: The second argument to path new should be a direction- "NORTH", "SOUTH", "EAST", or "WEST"

src/openworld/src/autogenerate.c
```diff
@@ -0,0 +1,80 @@
...
+        assert(SUCCESS == add_path_to_room(game->curr_room, path_to_room));
```
dwahme, line 54: When you create a path to a room, it only goes one way. You need to create a path going in the opposite direction as well here

src/openworld/src/autogenerate.c
bryan22lee: This was resolved by creating a 2D array buffer to hold each possible direction ("NORTH", "SOUTH", "EAST", "WEST") on the stack, and randomly getting one of these elements for each created path. This way, directions are also kept random for path generating.

src/openworld/src/autogenerate.c
bryan22lee: I created and added an opposite path as well, that leads _to_ the room that the previously created path leads _from_. I made sure it was properly added to the current room of the game (opposite of the previous path), and that it's direction ("NORTH", "EAST", "SOUTH", "WEST") was flipped.

src/openworld/src/autogenerate.c
```diff
@@ -0,0 +1,103 @@
...
+            random_index++;
+        }
+        else
+        {
+            random_index--;
+        }
+
+        // Add the opposite path to the game
+        path_t* path_to_room2 = path_new(game->curr_room, directions[random_index]);
+        assert(SUCCESS == add_path_to_room(new_room, path_to_room));
```
dwahme, line 77: A more straightforward way to approach this is to reorder the directions such that we can use modulus to wrap around the directions
```
        char directions[3][6];
        strncpy(directions[0], "NORTH", 6);
        strncpy(directions[2], "EAST", 5);
        strncpy(directions[1], "SOUTH", 6);
        strncpy(directions[3], "WEST", 5);

        // Random direction for the path
        unsigned int forwards = rand() % 4;
        path_t* path_to_room = path_new(new_room, directions[forwards]);
        assert(SUCCESS == add_path_to_room(game->curr_room, path_to_room));

        // Add the opposite path to the game
        unsigned int backwards = (forwards + 2) % 4;
        path_t* path_to_room2 = path_new(game->curr_room, directions[backwards]);
        assert(SUCCESS == add_path_to_room(new_room, path_to_room));
```

src/openworld/src/autogenerate.c
```diff
@@ -0,0 +1,103 @@
...
+
+        return SUCCESS; /* room added */
+    }
+
+    return FAILURE; /* room not added */
+}
```
dwahme, line 83: Additionally, I'm not convinced that the approach of always just picking one random direction is the best approach here- if a path in that direction exists, what happens? Or if this function is supposed to only work on rooms without paths, then it's pretty limiting; we can generalize this function pretty easily to build a path for any room that doesn't already have 4 paths leading out, or fail otherwise

There's a few different ways to solve this problem, but the simplest is to pick an initial direction, and then just bump up the index of the direction we choose by one if a path in that direction exists already (and use a for/while loop to make sure that we don't keep going forever). It would look something like the following
```
/* See autogenerate.h */
int room_generate(game_t *game, gencontext_t *context, char *room_id)
{
    char directions[3][6];
    strncpy(directions[0], "NORTH", 6);
    strncpy(directions[2], "EAST", 5);
    strncpy(directions[1], "SOUTH", 6);
    strncpy(directions[3], "WEST", 5);

    // Pick a random starting direction for the path
    unsigned int initial_dir = rand() % 4;

    for (int bump = 0; bump < 4; bump++)
    {
        int forwards = (initial_dir + bump) % 4;
               
        if (path_exists_in_dir(room, directions[forwards])) // TODO- implement this
        {
            continue; // skip attempting to build a path in this direction
        }
        // else, build a path

        // Adds one generated room from the head of the speclist only
        room_t *new_room = roomspec_to_room(game, context->speclist->spec, room_id);
        assert(SUCCESS == add_room_to_game(game, new_room));

        // Build the path to the new room
        path_t* path_to_room = path_new(new_room, directions[forwards]);
        assert(SUCCESS == add_path_to_room(game->curr_room, path_to_room));

        // Add the opposite path to the game
        unsigned int backwards = (forwards + 2) % 4;
        path_t* path_to_room2 = path_new(game->curr_room, directions[backwards]);
        assert(SUCCESS == add_path_to_room(new_room, path_to_room));

        return SUCCESS; // we made a new room
    }

    return FAILURE; // no rooms were made
}
```

src/openworld/src/autogenerate.c
```diff
@@ -0,0 +1,103 @@
...
+
+    item_hash_t *current, *tmp;
+    HASH_ITER(hh, roomspec->items, current, tmp) {
+        assert(SUCCESS == copy_item_to_hash(&res->items, roomspec->items, current->item_id));
+    }
+
+    res->paths = NULL;
+
+    return res;
+}
```
dwahme, line 38: I don't see the `game_t *game` param used here- do we need it?

include/game-state/room.h
```diff
@@ -214,6 +214,18 @@ item_t* get_item_in_room(room_t* room, char* item_id);
...
+ *
+ * Parameters:
+ *  paths - of type path_hash_t*
+ *
+ * Returns:
+ * - 0 if success
+ * - 1 if failure
+ */
+int delete_all_paths(path_hash_t *paths);
+
```
dwahme, line 228: Also, we should remove this edit from the game-state module

src/openworld/src/autogenerate.c
bryan22lee: Got it, that's definitely a much more straightforward way to get directions and their opposites!

src/openworld/src/autogenerate.c
bryan22lee: Oops I was using it before I changed the way it works. Will get rid of that parameter.

include/game-state/room.h
bryan22lee: This has been removed

src/openworld/src/autogenerate.c
bryan22lee: It didn't occur to me before but it definitely makes sense to account for existing paths so there are no overlapping directions. Also updated roomspec_t struct in gen_structs module to have an additional num_built field that keeps track of the number of times the roomspec is used to generate a room. This value is appended to the room_name to create a unique room_id to pass into roomspec_to_room

dwahme:
Issue Score: ✔️+

Comments:

PR #698, opened 2020-05-28
Added example CLI program

borjasotomayor:
This pull request adds an example program that is equivalent to running the main chiventure executable, except:

- The game is created in memory
- The CLI is monkeypatched to accept two additional commands: "glitter" and "taste"

The game looks like this:

```
   Welcome to Chiventure!

  > look
   Verily, this is the first room.
  > look rock
   You were hoping this was The Rock but, alas, it is just a plain and ordinary rock
  > glitter
   You throw some glitter. For a fleeting moment, the room feels a bit more fabulous. The glitter seems to have no effect beyond
that.
  > glitter rock
   I do not know what you mean.
  > taste rock
   It has gravel-ey bouquet.
  > taste
   You must identify an object to act on
```

I've also added a top-level "examples" target that can be used to build these (and future) examples.

Finally, this PR removes an old CLI example that is no longer relevant.


borjasotomayor:
That's odd, I don't get that message (I get the expected `It has gravel-ey bouquet.` message). What is the exact sequence of commands you entered that led to that happening?

borjasotomayor:
Nevermind, it looks like I had not re-built after merging in the latest changes from dev, and now it does happen. I'll look into why it's doing that.

borjasotomayor:
So it looks like what's happening is that, if you don't define end conditions for the game, then action management interprets the lack of end conditions as meaning "all the end conditions have been met and the game is over". This is what happens at the end of processing an action:

https://github.com/uchicago-cs/chiventure/blob/706ee3789199a1afca7b44e4b7f506e91e75fbc5/src/action_management/src/actionmanagement.c#L118-L129

And this is where the end conditions are checked:

https://github.com/uchicago-cs/chiventure/blob/706ee3789199a1afca7b44e4b7f506e91e75fbc5/src/game-state/src/game.c#L126-L131

It seems that, if there are no end conditions, it would make more sense for that to be interpreted as "the game never ends", which seems more flexible than requiring that a game author always specify and end room or a set of end conditions.

@MaxineK36: Could you check with Action Management to see whether we can just change "true" to "false" above? Doing so only makes one test fail (and it's the one that checks the return value when the end conditions are empty)

borjasotomayor:
@MaxineK36 talked with Action Management about this on their Slack channel, and they confirmed that this is something they can fix, but it makes more sense for that to be done as a separate pull request once this example makes it in to dev. 

@ecoble + @firatciftci: please go ahead and review the PR under the assumption that the end of game issue will be fixed separately.

borjasotomayor:
The example program has been added, but does not yet function exactly as expected. Once issue #708 is fixed, running `taste rock` should produce the expected result.

PR #721, opened 2020-05-30
Fixing Game State Header File Structure

dwahme:
The game-state module used to have 4 header files in the `/src/game-state/src` directory that exposed useful functions for other parts of chiventure:
- `/src/game-state/src/common-game-action.h`
- `/src/game-state/src/common-item.h`
- `/src/game-state/src/common-path.h`
- `/src/game-state/src/common-room.h`

This change moves the functions from the `/src/game-state/src/common-*.h` files into the appropriate `/incude/game-state/*.h` files, since there wasn't a clear dividing line between them before. This also means that the functions are now in the standard include path for chiventure.


borjasotomayor:
LGTM :shipit: 

I've also removed the `target_include_directories` commands in various CMake files that included the private game state headers.

Issue #669, opened 2020-05-27
Create an executable demonstrating functionality of playerclass code


simrenk:
First, we will attempt to integrate with battle-systems; we will ask them to provide functionality for loading the skills from playerclass (our class struct) into battle. 
If this cannot be integrated prior to Saturday at the latest, we will create a more general main.c demonstrating our class functionality.

simrenk:
**5/26** Reached out to battle-systems, explained the function we need them to implement, asked if they are interested in/able to implement the function. Awaiting further communication.

simrenk:
**5/28** @ethanwtai and I met with @Purrsephone from battle-systems. **Takeaways:**

-  Battle-systems has submitted a PR for integrating our class struct
-  Action Management should actually likely be the group that should load our skills
-  Battle-systems will then need to interface with Action Management to retrieve our skills so we 
    can use them in battle
- @Purrsephone will be meeting with @dwahme to figure out details and will update us on 
      feasibility. @ethanwtai and I are also currently in communication with @dwahme to figure out if 
      integrating with battle-systems is feasible from a timing perspective

ethanwtai:
Since we have decided to create a main executable which will showcase our code, we aim to have code to display a few features of our code: 
- basic information of the class
- proof-of-concept associated skills with each class (#725 )
- class item restrictions

ethanwtai:
In #769, we were able to create an executable that demonstrates basic information of the class, proof-of-concept associated skills with each class, and class item restrictions, as well as how they would work out in a battle scenario. Unfortunately, we were not able to fully integrate with battle systems or chiventure, as we created a separate executable from both.

dwahme:
Issue Score: ✔️++

Comments:
Great job! Well done specifying atomic items that are left to do.

PR #725, opened 2020-05-31
Creating an example class/skill list

ethanwtai:
Instead of integrated custom action blocks, I've created a separate struct to contain a list of classes alongside a specific hard-coded skill for each to help integrate with battle-systems. To help create an executable demonstrating the playerclass features (#669), class_list.c provides a hard-coded way of using sample class-based skills.


src/playerclass/examples/class-list.h
```diff
@@ -0,0 +1,27 @@
...
+#include "common/utlist.h"
+#include "playerclass/class.h"
+#include "playerclass/class_structs.h"
+
+typedef struct move_list {
+    class_t* c;
+    char* spell;
+    int damage;
+    int id;
+    struct move_list *next;
```
simrenk, line 16: might make it easier for battle-systems to navigate through this list if you make it doubly linked and add a *prev pointer as well but definitely not strictly necessary. would need to use DL_PREPEND instead of LL if you decide to do this!

src/playerclass/examples/class-list.c
```diff
@@ -0,0 +1,67 @@
...
+        fprintf(stderr, "Could not allocate memory for move_list()");
+        exit(1);
+    }
+    
+    wizardlist->c = wizard;
+    wizardlist->spell = "Fireball";
+    wizardlist->damage = 10;
+    wizardlist->id = 1;
+    
+    LL_PREPEND(wizardlist, bardlist);
```
simrenk, line 45: I'm not sure that LL_PREPEND and next are working the way they're supposed to. From [searching up documentation](https://doc.riot-os.org/group__sys__ut.html#ga851fb4b583b0dc000097eeee6dd551d4), I found that the second element is added to the head of the first. So bardlist is added prior to wizardlist. But if wizardlist is the last elem in the linkedlist, then its next should be NULL. 

**Easy fixes: 1. swapping bardlist and wizardlist in LL_PREPEND or 2. keeping LL_PREPEND as is, not initializing bardlist's next and initializing wizardlist's next as NULL** 

Here's the actual code for [LL_PREPEND ](https://libcoap.net/doc/reference/4.2.1/utlist_8h.html#a851fb4b583b0dc000097eeee6dd551d4) as well in case it helps (since LL_PREPEND just calls LL_PREPEND2 look at that).

src/playerclass/examples/class-list.c
```diff
@@ -0,0 +1,67 @@
...
+        fprintf(stderr, "Could not allocate memory for move_list()");
+        exit(1);
+    }
+    
+    knightlist->c = knight;
+    knightlist->spell = "Sword Slash";
+    knightlist->damage = 9;
+    knightlist->id = 2;
+
+    LL_PREPEND(bardlist, knightlist);
```
simrenk, line 64: If you end up swapping bardlist and wizardlist in LL_PREPEND above, this should actually be LL_PREPEND(wizardlist, bardlist).

If you end up making wizardlist the last elem by making it's next NULL and keeping LL_PREPEND the same, this LL_PREPEND will stay the same as well

src/playerclass/examples/class-list.c
```diff
@@ -0,0 +1,67 @@
...
+ * See class-list.h for more information.
+ */
+#include "class-list.h"
+
+/* see class-list.h */
+move_list_t* move_list() {
+
+    // Bard
+
+    class_t *bard = class_new("Bard", "Song-singer", "Song-singer and tale-teller", NULL, NULL, NULL, NULL, NULL);
```
simrenk, line 12: Also, not sure about the answer myself but, do these classes need to be initialized?

src/playerclass/examples/class-list.h
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+#include <stdlib.h>
+#include <stdio.h>
+#include "common/utlist.h"
+#include "playerclass/class.h"
+#include "playerclass/class_structs.h"
+
+typedef struct move_list {
+    class_t* c;
```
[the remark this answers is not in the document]
ethanwtai: I called it c since class was recognized as the struct. Do you still want me to change it to class?

src/playerclass/examples/class-list.c
ethanwtai: Since I don't think we're showing any of the other classes' attributes, I didn't initialize them.

src/playerclass/examples/class-list.h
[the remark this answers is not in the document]
simrenk: Okay sounds good!

src/playerclass/examples/class-list.c
```diff
@@ -0,0 +1,68 @@
...
+        fprintf(stderr, "Could not allocate memory for move_list()");
+        exit(1);
+    }
+    
+    bardlist->c = bard;
+    bardlist->spell = "Diss Track";
+    bardlist->damage = 8;
+    bardlist->id = 0;
+    bardlist->next = NULL;
+    bardlist->prev = NULL;
```
dwahme, line 27: Can we get a simple new function for this?

src/playerclass/examples/class-list.c
```diff
@@ -0,0 +1,68 @@
...
+
+    if (wizardlist == NULL){
+        fprintf(stderr, "Could not allocate memory for move_list()");
+        exit(1);
+    }
+    
+    wizardlist->c = wizard;
+    wizardlist->spell = "Fireball";
+    wizardlist->damage = 10;
+    wizardlist->id = 1;
```
dwahme, line 44: Can we replace this with a call to the new function

src/playerclass/examples/class-list.c
```diff
@@ -0,0 +1,68 @@
...
+
+    if (knightlist == NULL){
+        fprintf(stderr, "Could not allocate memory for move_list()");
+        exit(1);
+    }
+    
+    knightlist->c = knight;
+    knightlist->spell = "Sword Slash";
+    knightlist->damage = 9;
+    knightlist->id = 2;
```
dwahme, line 63: Can we replace this with a call to the new function

src/playerclass/examples/class-list.h
```diff
@@ -0,0 +1,28 @@
+/*
```
dwahme, line 1: This needs `#ifndef` guards

src/playerclass/examples/class-list.h
```diff
@@ -0,0 +1,28 @@
...
+
+/* Returns a hard-coded linked list with the class, spell name, damage, 
+ * and id number of the spell. 
+ * 
+ * Parameters:
+ * - none
+ * 
+ * Returns:
+ * - move_list_t* linked list with the class, spell, damage, and id number
+ */
```
dwahme, line 28: There's no function header here

dwahme, line 28: Also, don't forget the closing `#endif` guard

src/playerclass/examples/class-list.c
```diff
@@ -0,0 +1,51 @@
...
+    DL_PREPEND(bardlist, wizardlist);
+    
+    // Knight
+
+    class_t *knight = class_new("Knight", "Brave", "Brave and shiny", NULL, NULL, NULL, NULL, NULL, NULL);
+    move_list_t* knightlist = new_move_list(knight, "Sword Slash", 9, 2);
+
+    DL_PREPEND(wizardlist, knightlist);
+
+    return knightlist;
```
dwahme, line 50: The way you're appending these lists is a little unorthodox (and seems like it might've been a workaround for an underlying utlist issue). It should look a bit more like the following
```
move_list_t *head = NULL;
DL_APPEND(head, bardlist);
DL_APPEND(head, wizardlist);
DL_APPEND(head, knightlist);
return head;
```

dwahme:
Issue Score: ✔️+

Comments:

PR #697, opened 2020-05-28
Skill Trees: Skill TREE Implementation

jackandthebean:
# Skill Trees: Skill TREE Implementation
### Following up on skeleton code PR from Sprint 3
Our first PR – the `rpg-skilltrees` skeleton code (#601) – implemented individual skills, player skill inventories and their associated functions. However, actual skill **trees** and associated skill nodes were not implemented. The main object of this pull request is to incorporate the first of the skill **trees** implementation. The rest of the skill trees code is set for a subsequent PR, as a significant number of changes have been made to our code since the skeleton code PR.
### Specific changes since the last PR
Beyond the addition of `skilltree.h` and `skilltree.c` files that contain the start of the new skill trees implementation, our previous code has been refactored.
[We do not want to keep the skeleton code files, as that code has been relocated as follows.]
* `stdst.h` and `stdst.c` now contain all data structures and common helper functions, including previously defined skill and inventory structs, as well as the new skill node and skill tree structs (NOTE: some structs, though essentially the same as in the first PR, were renamed to avoid naming conflicts with other teams)
* `skill.h` and `skill.c` contain all the previously written functions having to do with individual skills
* `inventory.h` and `inventory.c` contain all the previously written functions having to do with skill inventories, with the addition of a `skill_levels_update()` function to update levels of all skills in an inventory
* All unit test files were removed for this PR, as @michaelleestuy and @steeltrapp will be updating our unit tests to be more robust and reflect all minor changes in previously written code (e.g. struct renames, etc.). These newer unit tests will come in a subsequent PR

PR #728, opened 2020-06-01
Skill Trees: Additional 'Skill Tree' Implementation

VannStrasen:
**Following up on [our previous PR](https://github.com/uchicago-cs/chiventure/pull/697)**
Our most recent PR – the initial skill tree code creation (#697) – implemented a major part of our skill tree functionality. However, it did not fully implement all functions because doing so would be too much for one PR, so we split up the task across different sprints. 

**Specific changes since the last PR**
* A handful of functions inside the skilltree.c file were implemented.
* Some small changes were made to the return types of functions. 
* Header comments in skilltree.h were changed when this implementation phase discovered more return types than previously thought necessary.

**Notes to PR reviewers**
* While this looks like there is a lot of code added to dev, a large portion of this code should instead be reviewed in PR #697. (Refer to the specific changes for what was actually edited)
* These skills have yet to be tested, and that will be done by @steeltrapp and @michaelleestuy later in the sprint.


VannStrasen:
Fixed the warnings generated from the merge into dev!

src/skilltrees/skilltree.c
```diff
@@ -102,48 +102,199 @@ int skill_tree_free(skill_tree_t* tree) {
...
+        return NULL;
+    } else {
+        return acquired;
+    }
 }
 
 /* See skilltree.h */
 skill_t** skill_prereqs_missing(skill_tree_t* tree,
                                skill_inventory_t* inventory, sid_t sid,
                                int* nmissing) {
```
firatciftci, line 233: To preserve the same style in variable naming convention, `nmissing` should be changed to something like `num_missing`.

src/skilltrees/skilltree.c
```diff
@@ -102,48 +102,199 @@ int skill_tree_free(skill_tree_t* tree) {
...
-    /* TO DO */
-    return NULL;
+    assert(tree != NULL && inventory != NULL);
+
+    unsigned int nprereqs;
```
firatciftci, line 236: To preserve the same style in variable naming convention, `nprereqs` should be changed to something like `num_prereqs`.

src/skilltrees/skilltree.c
```diff
@@ -102,48 +102,199 @@ int skill_tree_free(skill_tree_t* tree) {
...
 }
 
 /* See skilltree.h */
 int inventory_skill_acquire(skill_tree_t* tree, skill_inventory_t* inventory,
                            skill_t* skill) {
-    /* TO DO */
-    return 0;
+    assert(tree != NULL && inventory != NULL && skill != NULL);
+
+    unsigned int nmissing;
```
firatciftci, line 285: To preserve the same style in variable naming convention, `nmissing` should be changed to something like `num_missing`.

src/skilltrees/skilltree.c
VannStrasen: Fixed in most recent commit.

src/skilltrees/skilltree.c
VannStrasen: Fixed in most recent commit.

src/skilltrees/skilltree.c
VannStrasen: Fixed in most recent commit.

VannStrasen:
Good catch — I fixed those and found a couple more, so I edited those as well.

firatciftci:
Issue Score: ✔️+

Comments:

PR #492, opened 2020-05-04
Action management/walk synonym

isaacfink21:
Resolves #476, reintroducing the "WALK" synonym for "GO" action.

PR #736, opened 2020-06-03
Implementation for action synonyms

isaacfink21:
A previous pull request (#492), reintroduced legality for the WALK, PICKUP, USE, and EAT and DRINK action synonyms. However, pre-existing functionality was only available for WALK. This PR implements functionality for the rest of the synonyms by rerouting them to refer to their "parent" actions. This way, for example, "EAT" is identical to "CONSUME" for the purpose of altering the game state, but these two terms are still distinguished in CLI such that they could theoretically return different messages in the future (e.g. "You cannot eat the wand" instead of "You cannot consume the wand").


src/action_management/src/actionmanagement.c
```diff
@@ -82,6 +82,16 @@ int do_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *i, char **ret_
         return WRONG_KIND;
     }
 
+    // use representative c_name for action synonyms
```
MaxineK36, line 85: This should be a block comment (`/* use representative c_name for action synonyms */`)

isaacfink21:
PR Approved! Action synonyms `WALK`, `PICKUP`, `EAT`, `DRINK`, and `USE` are now fully implemented. These synonyms will be recognized separately in CLI, but upon being processed in action management, will be converted to their appropriate parent action such that they are functionally identical for the purposes of altering the game state.

MaxineK36:
Issue Score: ✔️+

Comments:
Good work overall! One thing for future PRs: when you create a PR, you are requesting a review of the code in the branch you want to merge. As such, you should not push any commits to the branch right after creating the PR; you should only do so in response to a code review. Make sure that, when you create a PR, the code is ready for review and passes all the tests.

PR #742, opened 2020-06-03
Removed certain dependencies in chiventure_ctx_t

borjasotomayor:
This pull request refactors out a couple of dependencies in chiventure_ctx_t that have proven to be problematic. Most notably:

- The UI context is no longer initialized in `chiventure_ctx_init` but, rather, inside `start_ui` (we could also move this initialization code to `chiventure.c`, but I'm not sure that's necessary given that the only situation where we need an initialized UI context is if we're going to run the UI). This means that the `ncurses.h` header file is only included in the UI code (where previously it ended up being included indirectly from `ctx.h`, which was causing issues for the graphics team)
- `chiventure_ctx_init` and `chiventure_ctx_new` now take a `game_t` as a parameter instead of a filepath. This provides more flexibility when creating a context object, where a game may be created in memory (instead of being loaded from a file). Previously, we had to pass a NULL `filepath`, and then replace the default game with the in-memory game (see `extra-command.c` for an example of this)

Unrelated to this refactoring, the changes brought to light a small issue in the game-state tests, where unallocated memory was being passed to a function that expected that memory to be allocated (and was not previously segfaulting by chance).

### DIFF
--- a/include/common/ctx.h
+++ b/include/common/ctx.h
@@ -28,7 +28,9 @@ typedef struct chiventure_ctx {
  * Allocates and initializes a new chiventure context.
  *
  * Parameters:
- *  - game: The game to be run in this context (can be NULL)
+ *  - game: The game to be run in this context.
+ *          If NULL, an empty game is created that simply
+ *          warns the user they need to load a game.
  *
  * Returns:
  *  - A pointer to the context, or NULL if it cannot be allocated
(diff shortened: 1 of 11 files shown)


include/common/ctx.h
```diff
@@ -30,25 +28,25 @@ typedef struct chiventure_ctx {
  * Allocates and initializes a new chiventure context.
  *
  * Parameters:
- *  - filepath: path to yaml file to load game
+ *  - game: The game to be run in this context (can be NULL)
```
MaxineK36, line 31: Might be helpful to have a note here about the behavior of the function when game is NULL

borjasotomayor: Fixed!

PR #743, opened 2020-06-04
Battle systems message strings for printing during a battle 

celiaanderson:
This pull request adds functions for returning strings storing messages to be printed about the current battle. This includes the functions print_start_battle(), print_battle_move(), print_battle_winner(), and print_hp(). The first three functions allocate space (BUFFER_SIZE) for a string and then store a message in that string that describe a) the start of the battle, b) a player or enemy move, and c) a player or enemy win; these functions then return that string. The last function, print_hp(), adds on a message about the current HP of the player and the enemies to an already allocated string, so information about HP can be easily added onto any message describing the battle, or printed on its own if given a blank string. 

These functions are in battle_print.c/.h and are tested in test_battle_print.c. They are meant to be used in battle flow later, so that these strings can then be printed to describe the battle to the player.


src/battle/src/battle_print.c
```diff
@@ -0,0 +1,92 @@
...
+    DL_FOREACH(enemies, enemy_elt)
+    {
+        char* name = enemy_elt->name;
+        int enemy_hp = enemy_elt->stats->hp;
+
+        slen += snprintf(string + slen, BUFFER_SIZE, "-- %s's HP: %d\n", name, enemy_hp);
+    }
+
+    return SUCCESS;
+}
```
Isaaq-Khader, line 47: Before this print statement, I think it might be best to have it so that it also includes "Enemy HP" since in the future there might be companions/allies to help battle and it could differentiate who is battling.

Isaaq-Khader: The way I see this looking is like this:

-- Your HP: 89
Enemy HP:
-- Bob's HP: 64
-- Annie's HP: 75

or something that looks prettier

celiaanderson: That's a good idea, thanks!

celiaanderson:
Added an "ENEMY HP" header write before printing out all the enemy HPs, based off the suggestion. Tests are passing!

src/battle/src/battle_print.c
```diff
@@ -0,0 +1,93 @@
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "battle/battle_print.h"
+
+#define BUFFER_SIZE (200) // copied from actionmanagement.c, changed to 200
```
dwahme, line 7: You should include the actionmanagement file instead

src/battle/src/battle_print.c
```diff
@@ -0,0 +1,93 @@
...
+
+/* see battle_print.h */
+char *print_start_battle(battle_t *b)
+{
+    char *enemy_name = b->enemy->name;
+    int player_hp = b->player->stats->hp;
+    int enemy_hp = b->enemy->stats->hp;
+
+    char *string = calloc(1, BUFFER_SIZE);
```
dwahme, line 16: Your calloc should be `calloc(BUFFER_SIZE + 1, sizeof(char))` (amount of items, size of an item). While `sizeof(char) == 1`, it's much clearer to the reader why we're callocing this size. Also, the +1 is for the null terminator

src/battle/src/battle_print.c
```diff
@@ -0,0 +1,93 @@
...
+    int rc = print_hp(b, string);
+    assert(rc == SUCCESS);
+
+    return string;
+}
+
+/* see battle_print.h */
+int print_hp(battle_t* b, char* string)
+{
+    int slen = strlen(string);
```
dwahme, line 31: Most C string functions (strcpy, strcat, etc.) have an equivalent "safe" version (strncpy, strncat, etc.) where you can specify the maximum number of characters to copy/concatenate/etc. Use these instead to protect against accidental (or malicious) buffer overflows

You should be using `strnlen()` here

src/battle/src/battle_print.c
```diff
@@ -0,0 +1,93 @@
...
+    int rc = print_hp(b, string);
+    assert(rc == SUCCESS);
+
+    return string;
+}
+
+/* see battle_print.h */
+char *print_battle_winner(battle_status_t status, int xp)
+{
+    char *string = calloc(1, BUFFER_SIZE);
```
dwahme, line 82: Your calloc should be `calloc(BUFFER_SIZE + 1, sizeof(char))` (amount of items, size of an item). While `sizeof(char) == 1`, it's much clearer to the reader why we're callocing this size. Also, the +1 is for the null terminator

src/battle/src/battle_print.c
```diff
@@ -0,0 +1,93 @@
...
+
+    slen += snprintf(string + slen, BUFFER_SIZE, "ENEMY HP\n");
+    combatant_t *enemies = b->enemy;
+    combatant_t *enemy_elt;
+    DL_FOREACH(enemies, enemy_elt)
+    {
+        char* name = enemy_elt->name;
+        int enemy_hp = enemy_elt->stats->hp;
+
+        slen += snprintf(string + slen, BUFFER_SIZE, "-- %s's HP: %d\n", name, enemy_hp);
```
dwahme, line 44: Don't ever do pointer arithmetic if you can help it- in this case, we actually are allowing the possibility of a buffer overflow. You should be using `strncat()` here instead (or, first a call to `snprintf()` to build the second half of the mesage, and then `strncat()` to append it

tests/battle/test_battle_print.c
```diff
@@ -0,0 +1,204 @@
...
+    player_t *ctx_player = new_ctx_player("player_name", NULL, player_stats, NULL, NULL);
+    npc_enemy_t *npc_enemy = make_npc_enemy("Bob", NULL, enemy_stats, NULL, NULL);
+    environment_t env = ENV_DESERT;
+    battle_t *b = set_battle(ctx_player, npc_enemy, env);
+    cr_assert_not_null(b, "set_battle() failed");
+    b->player->stats->hp = 89;
+    b->enemy->stats->hp = 64;
+
+    /* Set up string to store message in */
+    char* string = calloc(1, BUFFER_SIZE);
```
dwahme, line 55: Your calloc should be `calloc(BUFFER_SIZE + 1, sizeof(char))` (amount of items, size of an item). While `sizeof(char) == 1`, it's much clearer to the reader why we're callocing this size. Also, the +1 is for the null terminator

tests/battle/test_battle_print.c
```diff
@@ -0,0 +1,204 @@
...
+    cr_assert_not_null(e2, "make_npc_enemy() failed");
+    environment_t env = ENV_DESERT;
+    battle_t *b = set_battle(ctx_player, head, env);
+    cr_assert_not_null(b, "set_battle() failed");
+    b->player->stats->hp = 89;
+    b->enemy->stats->hp = 64;
+    b->enemy->next->stats->hp = 75;
+
+    /* Set up string to store message in */
+    char* string = calloc(1, BUFFER_SIZE);
```
dwahme, line 93: Your calloc should be `calloc(BUFFER_SIZE + 1, sizeof(char))` (amount of items, size of an item). While `sizeof(char) == 1`, it's much clearer to the reader why we're callocing this size. Also, the +1 is for the null terminator

src/battle/src/battle_print.c
celiaanderson: I originally didn't include it since the action management files also include game state, so if I included it, structs like item_t would clash between game state and our battle stubs. Should I just go into the other battle systems files and rename item_t? I think player_t and game_t in battle flow structs would also have to be renamed

dwahme: Let's just move this definition to the .h file then and call it `BATTLE_BUFFER_SIZE` to avoid name conflicts. Since this is going to change a few things, we can update the value to pull from actionmanagement in a later PR that deals with integration

src/battle/src/battle_print.c
```diff
@@ -0,0 +1,104 @@
...
+
+    char temp[BUFFER_SIZE + 1];
+
+    n = snprintf(temp, BUFFER_SIZE, "-- Your HP: %d\n", player_hp);
+    strncat(string, temp, BUFFER_SIZE - slen);
+    slen += n;
+
+    snprintf(temp, BUFFER_SIZE, "ENEMY HP\n");
+    strncat(string, temp, BUFFER_SIZE - slen);
+    slen += n;
```
dwahme, line 44: Does `n` need to be redefined in the second `snprintf` here?

celiaanderson:
Summary: As part of issue #743, pull request added new files battle_print.c/.h and test_battle_print.c that include new functions for returning newly calloced strings storing a message that describes a point in the current battle. The fundamental goal and design of these functions are the same as in the initial description. The changes during this pull request were (a) to add a line "ENEMY HP\n" in print_hp to indicate that enemy HPs are about to printed out, (b) fix the calloc calls, (c) switch from strlen to strnlen, (d) use strncat for adding extra lines of message in print_hp instead of doing pointer arithmetic, and (e) change BUFFER_SIZE to BATTLE_BUFFER_SIZE.

dwahme:
Issue Score: ✔️++

Comments:
Great job! Well done having a very self-contained PR.

PR #745, opened 2020-06-04
Battle/class moves

Purrsephone:
Hello friends, 

This code takes a linked list of class specific skills from player class and adds them to the move list for use in the executable. I also made init, new, and free functions for a move_t. I have to include a .c file right now, which I will fix soon.


src/battle/src/battle_move_maker.c
```diff
@@ -0,0 +1,32 @@
+#include <criterion/criterion.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include <string.h>
+#include "battle/battle_move_maker.h"
+#include "../../playerclass/examples/class-list.c"
```
ethanwtai, line 7: I think you should include the header file not the .c file

include/battle/battle_move_maker.h
```diff
@@ -0,0 +1,22 @@
...
+#include "battle_moves.h"
+
+ /* Packages class-specific move from player class's linked list 
+ into battle move struct, adds this move to the input player's move list 
+ * Parameters:
+ * - player: a pointer to the player we want to get class-specific moves for 
+ * Returns:
+ * - player: the input player with correct move added to move list 
+ */
+player_t *add_class_move(player_t *player);
```
dwahme, line 19: Can we call this `build_moves()` or something along those lines instead? The name of this function kinda threw me off

src/battle/src/battle_move_maker.c
```diff
@@ -0,0 +1,32 @@
...
+{
+    move_list_t *moves = move_list();
+
+    move_list_t *tmp;
+
+    player_t *ret_player;
+
+    move_t *ret_move;
+
+    char *player_class_name = player->class->name;
```
dwahme, line 20: We can remove the blank lines here- it's spaced out a bit too much

src/battle/src/battle_move_maker.c
```diff
@@ -0,0 +1,32 @@
...
+
+    LL_FOREACH(moves, tmp)
```
dwahme, line 22: This should be `DL_FOREACH()`

src/battle/src/battle_move_maker.c
```diff
@@ -0,0 +1,32 @@
...
+    {
+        if(!(strcmp(player_class_name, tmp->c->name)))
+        {
+            ret_move = move_new(tmp->spell, tmp->id, NULL, true, tmp->damage, 0);
+            DL_PREPEND(player->moves, ret_move);
```
dwahme, line 27: Why are we prepending instead of appending? (I've only seen people do this when to get around a utlist crash due to misusing the linked lists)

src/battle/src/battle_move_maker.c
```diff
@@ -0,0 +1,32 @@
...
+            return player;
```
dwahme, line 28: Why do we have an early return here? This means only one move would get added

src/battle/src/battle_move_maker.c
```diff
@@ -0,0 +1,32 @@
...
+
+    player_t *ret_player;
+
+    move_t *ret_move;
+
+    char *player_class_name = player->class->name;
+
+    LL_FOREACH(moves, tmp)
+    {
+        if(!(strcmp(player_class_name, tmp->c->name)))
```
dwahme, line 24: Most C string functions (strcpy, strcat, etc.) have an equivalent "safe" version (strncpy, strncat, etc.) where you can specify the maximum number of characters to copy/concatenate/etc. Use these instead to protect against accidental (or malicious) buffer overflows

You should be using `strncmp()` here

src/battle/src/battle_move_maker.c
Purrsephone: There is only 1 hard-coded move per class.

dwahme:
Issue Score: ✔️

Comments:

The title of an issue has to convey what the issue will be producing. In the future, try using simple titles with a "Verb-Object" structure that clearly states what you will be doing in the issue (e.g., “Design and implement a ‘circle’ module”, “Research libraries for drawing simple geometric shapes”, etc.)

PR #746, opened 2020-06-04
Skilltrees/skill tree tests: updated unit tests for skill tree functions. 

michaelleestuy:
This PR contains the unit tests for all of the functions supported by the skill tree module at the moment. Currently, some of these tests fail for the present implementations, because some of these functions are incomplete, and there are some bugs.


steeltrapp:
Moved in-progress tests to separate folder, separate from cmake.

steeltrapp:
Moved in-progress tests out of separate folder, and placed them back into original files, marked for criterion to skip them.

firatciftci:
Issue Score: ✔️+

Comments:
Good work! The only thing I'll point out is that, when you create a PR, you are requesting a review of the code in the branch you want to merge. As such, you should not push any commits to the branch right after creating the PR; you should only do so in response to a code review. Make sure that, when you create a PR, the code is ready for review and passes all the tests.

PR #747, opened 2020-06-04
Graphics/UI: Map  5-Room Example using Chiventure Structures

damonhines:
This is an example program that creates a 5 room map and shows the current location of a player. This program was created with Chiventure room structs and game struct. This example should display a character located in the second room.

Commands to run this program within the chiventure directory:
`cd build `
`make `
`make examples`
`src/ui/examples/./map_raylib_example`

Screen Shot 2020-06-04 at 1 37 24 PM


src/ui/examples/map_raylib_example.c
```diff
@@ -0,0 +1,87 @@
...
+
+    while (!WindowShouldClose())   
+    {
+        BeginDrawing();
+
+            ClearBackground(BLACK);
+            
+            if (game->curr_room != NULL)
+            {
+                DrawRectangle(screenWidth/4*2 - 90, 170, 180, 130, RED);
```
ecoble, line 56: All of these numbers should be variables or macros so they aren't "magic numbers"

ecoble:
Issue Score: ✔️+
Comments:
Great job on this PR, especially with the summary! Like issues, PRs should also have closing statements.

PR #749, opened 2020-06-04
Custom Scripts Custom Type Struct Being implemented for npcs-rooms

suday18:
This PR implements the custom type module and specifically the object_t struct to be used within the npcs-rooms struct. This flexible data structure that can hold either a fundamental data type, such as bool, char, int, etc., or a Lua directory which will be responsible for returning the data. IN our case, we applied this struct to modify the room_id variable which was previously a char*.

Changes:

Updates to the npcs-rooms.c and npcs-rooms.h files in the game-state directory. 
Modifications to the room_id variable inside these two files and the existing npcs-room struct.


include/custom-scripts/custom_type.h
```diff
@@ -1,143 +1,162 @@
...
+ * A lua directory string must start with LUA
+ * Parameters:
+ * - s: string to be investigated
+ * Returns:
+ * - boolean value (true = lua directory)
+ */
+bool string_is_lua(char* s);
+
+/**
+ * extra_lua(), given input is a Lua directory, extracts the Lua directory
```
firatciftci, line 154: Small typo here: this should be `extract_lua` instead of `extra_lua`.

src/npc/src/rooms-npc.c
```diff
@@ -43,9 +49,10 @@ int npc_mov_init(npc_mov_t *npc_mov, char* npc_id, npc_mov_enum_t mov_type,
     return SUCCESS;
 }
 
-
+/*Changes to Room_ID: We are using this struct to modify the room_id 
+which will enable custom scripts and lua files to be loaded*/ 
 /* See rooms-npc.h */
-npcs_in_room_t *npcs_in_room_new(char* room_id)
+npcs_in_room_t *npcs_in_room_new(object_t* room_id)
```
hugo-chan, line 55: `npcs_in_room->room_id == 0` 
This line of code may not work as intended since you've changed the `room_id` field of the `npcs_in_room_t` struct to be of the type `object_t` and I don't think you can make you can make a comparison between object_t and 0.

src/npc/src/rooms-npc.c
```diff
@@ -100,7 +108,7 @@ int npcs_in_room_free(npcs_in_room_t *npcs_in_room)
 {
     assert(npcs_in_room != NULL);
 
-    free(npcs_in_room->room_id);
```
hugo-chan, line 103: I don't think we have implemented a free function for `object_t` (which we should have), so this line of code may not work as intended.

hugo-chan, line 103: I don't think we have implemented a free function for `object_t` (which we should have), so this line of code may not work as intended.

src/npc/src/rooms-npc.c
```diff
@@ -114,8 +122,8 @@ int npc_mov_free(npc_mov_t *npc_mov) {
     assert(npc_mov != NULL);
 
     free(npc_mov->npc_mov_type.npc_mov_indefinite);
-    free(npc_mov->track);
```
hugo-chan, line 117: Same concern here. Free function for `object_t`, if I recall correctly, is not in the module yet.

src/npc/src/rooms-npc.c
suday18: Yes, that's why I commented that line out in the new version.

src/npc/src/rooms-npc.c
suday18: Yes, that's why I commented that line out in the new version.

src/npc/src/rooms-npc.c
suday18: Hi Hugo, yea that make sense. I'm not sure how I could actually create a comparison such that that line will work? Any suggestions?

suday18:
Also, I'm not sure why the Travis build is failing. I looked inside the following CMakelists.txt file: https://github.com/uchicago-cs/chiventure/blob/custom-scripts/npc/CMakeLists.txt. It seems to have the Lua library linked along with the custom-scripts/src directory

src/npc/src/rooms-npc.c
firatciftci: @suday18 You should mark this as a TODO comment above the `free()` line and explain why it's commented out at the moment, so that other students in the future understand that this is the case. (This applies to other commented out `free()` calls as well.)

src/npc/src/rooms-npc.c
firatciftci: This is a valid concern, @suday18 don't forget that you need to conclude this by 4pm tomorrow, so you might want to try out something on your own to remedy this, in case @hugo-chan doesn't see this in time.

firatciftci:
@suday18 Please get this PR and also the other PR regarding the npc module integration to a state in which the Travis builds are successfully passing. Ideally, we should review the PR once everything is ready to merge and the code is stable. Let me know if you have any issues with this step, but also keep in mind that we are working with a time limit as well.

src/npc/src/rooms-npc.c
hugo-chan: I am not sure what the code is supposed to do and I think consulting the NPC team would be a good idea. But perhaps comparing it with NULL instead will work?

src/npc/src/rooms-npc.c
suday18: Have added the comments and marked the following as TODO as per this commit: https://github.com/uchicago-cs/chiventure/pull/749/commits/c5186e9d32dc12bcb52c45b52636b18511b598c1

src/npc/src/rooms-npc.c
suday18: I think I have also fixed this issue as per this commit: https://github.com/uchicago-cs/chiventure/pull/749/commits/c5186e9d32dc12bcb52c45b52636b18511b598c1

src/npc/src/rooms-npc.c
suday18: Have added the comments and marked the following as TODO as per this commit: 

https://github.com/uchicago-cs/chiventure/pull/749/commits/c5186e9d32dc12bcb52c45b52636b18511b598c1

suday18:
Closing Comment: Despite making a number of changes, we've not been able to get the Travis build to work. There were a lot more dependencies that we did not initially anticipate. As such, we have decided to close this pull request, since it is very close to the deadline and it seems unlikely that this could get to work with the rest of the chiventure code.

PR #750, opened 2020-06-04
Custom Scripts Custom Type Struct Being implemented for npcs 

suday18:
This PR implements the custom type module and specifically the object_t struct to be used within the npc struct. This flexible data structure that can hold either a fundamental data type, such as bool, char, int, etc., or a Lua directory which will be responsible for returning the data. IN our case, we applied this struct to modify the room_id variable which was previously a char*.

Changes:

Updates to the npc.c and npc.h files in the npc directory.
Modifications to the npc_id variable inside these two files and the existing npc struct.

### DIFF
--- a/include/npc/npc.h
+++ b/include/npc/npc.h
@@ -5,12 +5,20 @@
 #include "game-state/item.h"
 //TODO-#include "dialogue.h"
 #include "playerclass/class_structs.h"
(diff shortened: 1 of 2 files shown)


suday18:
This is being closed as we have created another PR which has the same changes but we have simply just pulled from dev instead.

PR #751, opened 2020-06-05
Custom actions/custom action t

emmamedeiros:
The custom action struct module has been created in files custom_action.h, custom_action.c and test_custom_action.c. Makefiles have also been updated accordingly. The module outlines the custom action struct, along with implementation and testing for new, init, and free functions for the struct.


emmamedeiros:
@annabelleperng thank you for your comments!! I'm working on figuring out if the assert statements should be changed, but in the meantime I think given the structure of the code it would be easier to have a "None" string for any field that doesn't have a value for a custom action. I made a note of that in the .h code, but it seems like the only other place that would be important is more in regard to the reading/structure of the files, which it seems like we won't get to at this point, so we can leave that in our documentation.

emmamedeiros:
This pull request was approved, and thus the custom-actions/custom_action_t branch will now be deleted.

MaxineK36:
Issue Score: ✔️+

Comments:
Good work overall! One thing for future PRs: when you create a PR, you are requesting a review of the code in the branch you want to merge. As such, you should not push any commits to the branch right after creating the PR; you should only do so in response to a code review. Make sure that, when you create a PR, the code is ready for review and passes all the tests.

Issue #739, opened 2020-06-03
Adding a linked list for arguments to be passed to Lua in object_t


hugo-chan:
With our current implementation of `object_t`, we recognized that our custom scripts are limited to scripts that take in no arguments. For example, a script that is supported by our current implementation is a random number generator script (the script does not take in any arguments from C and simply returns a number generated within the Lua script). This limitation significantly diminishes the usefulness of Lua scripts in chiventure because we would like our custom scripts to be dynamic. The custom scripts should be able to return different values based on the game state (i.e. some aspect of conditional flow). To achieve this, we would require a way for C to communicate with Lua, which this issue aims to create by passing arguments.

Specifically, we will modify the `object_t` struct to have a linked list of arguments that will be pushed to the virtual stack when calling the function inside the Lua script. The arguments inside the linked list can be any type specified in the union (this enables modification in the future).

There are several steps we foresee:
1. Need to create the struct representing the nodes of the linked list (this will require functions similar to the ones written for object_t)
2. Need to write an argument add function that will append add an argument (node) to the linked list
3. Need to modify current object_t functions and tests to accommodate for this design change
4. Need to significantly modify the getter functions. The arguments in the linked list of the object_t need to be pushed to the virtual stack one-by-one
5. Write test Lua scripts (that take in arguments unlike before) and unit tests for our functions

We will know this issue is complete when these steps are completed and we successfully submit a PR.

We will be working on the `custom-scripts/argument-list` branch.

suday18:
We've added  a number of additional tests for the arg_t struct, specifically with respect to the following functions:` arg_t_new, arg_t_bool, arg_t_char, arg_t_int, arg_t_str, arg_t_add`. 

In addition, we have further tested and all these tests are passing. We're currently working on adding a few more Lua tests that will check if the correct arguments are passed to Lua and that the script is making the expected changes to those arguments.

hugo-chan:
We have achieved what we described in the description and have written tests for our code. The current supported types for the arguments are `bool`, `char`, `int`, and `string`, but more can be added in the future to make `object_t` even more dynamic.

We submitted a pull request here #752.

firatciftci:
Issue Score: ✔️++

Comments:
Amazing work!

PR #752, opened 2020-06-05
Support for a linked list of arguments to be passed in Lua in object_t

hugo-chan:
The motivation behind this PR is described in detail in issue #739.

In summary, we have added support for a linked list of arguments in our `object_t` struct in our `custom_type` module, enabling Lua custom scripts in chiventure to be much more dynamic and customizable. For example, these Lua scripts can take in arguments related to the game state and return different values based on these arguments. 

Currently, the only supported types for these arguments are `bool`, `char`, `int`, and `string` (the type of the argument does not need to be the same as the type of the data in the `object_t`), but more can be added in the future.

Changes:
- Addition of a linked list field within `object_t`
- Functions operating on this linked list
    - Creation of a new node
    - Initialization of a new node (`bool`, `char`, `int`, `string`)
    - Addition of a node to the linked list
- Modified getter functions so that arguments are passed to the Lua script through the virtual stack
- Created tests for the linked list functions (pass)
- Created test Lua scripts and `object_t`s with argument lists to test the correctness of the new getter functions (pass)


include/custom-scripts/custom_type.h
```diff
@@ -22,6 +22,82 @@ typedef enum data_type
...
+        bool b;
+        char c;
+        int  i;
+        char *s;
+        char *lua;
+    } data;
+
+    // Pointer to the next argument
+    struct arg_t *next;
+} arg_t;
```
firatciftci, line 44: Having `arg_t` a doubly-linked list instead of a single linked list would make more sense, as it's easier to work with and usually more efficient. (It also ties in better with other link list usage cases in chiventure)

include/custom-scripts/custom_type.h
```diff
@@ -22,6 +22,82 @@ typedef enum data_type
...
+
+/**
+ * arg_t_add() adds an arg_t struct to the end of the arg_t linked list
+ * Parameters:
+ * - head of the arg_t linked list
+ * - arg_t struct to be added
+ * Returns:
+ * - head of the arg_t linked list
+ */
+arg_t *arg_t_add(arg_t *head, arg_t *add);
```
firatciftci, line 97: A comment from Dawson about the `arg_t` function set:

It would be better if you provide `obj_add_arg_<type>(object_t obj, <type> val)` style functions instead of (or in addition to) the existing `arg_t_<type>()` functions since that's how the module is gonna be used anyway. Having to type both `arg_t_<type>()` and `arg_t_add()` frankly just makes it extra work.

Then, you don't have to update the `obj_t_<type>()` functions at all. The module would be used more like the following:
```c
obj_t new_obj = obj_t_int(0, "filepath here");
int rc = obj_add_arg_bool(new_obj, true);
rc = obj_add_arg_lua(new_obj, "other filepath here");
.... // and you can check the ret val of rc for success/failure
```

include/custom-scripts/custom_type.h
hugo-chan: Just changed it to doubly linked list. Modified `arg_t_new()` and `arg_t_add()` accordingly and added a new test.

include/custom-scripts/custom_type.h
hugo-chan: Just implemented the suggestion. 

Arguments to be added are now called using `obj_add_arg_<type>(object_t obj, <type> val)` to make the module more user-friendly. Simplified module accordingly so that all the `arg_t` functions that won't be used by the user are now moved from the `.h` file and tucked away in the `.c` file.

firatciftci:
Issue Score: ✔️++

Comments:
Great work!

PR #759, opened 2020-06-05
Fixed a small bug in autogenerate.c for backwards paths

bryan22lee:
Discovered a small bug in [autogenerate.c](https://github.com/uchicago-cs/chiventure/blob/dev/src/openworld/src/autogenerate.c) that created, but did not add, a backwards path for room_generate(). This bug _must_ be addressed to be able to return to the last room whenever a new room is generated. Fixed this bug and added tests for it so that it is not missed again in the future.


dwahme:
Issue Score: ✔️+

Comments:

PR #773, opened 2020-06-07
Putting class (and so stats/effects) into Player and NPC

carlosazpurua1004:
New PR implementing what was discussed in meeting with NPC and Class team.
Gives Player and NPC structs class fields, which already contains stat/effect hashtables


carlosazpurua1004:
Note I asked the NPC team to review this a while back, but they never responded, though its only changes we agreed upon and pretty small PR

include/game-state/player.h
```diff
@@ -8,7 +8,7 @@
 
 #include "game_state_common.h"
 #include "item.h"
-#include "stats.h"
```
sumedhsarwate, line 11: I think you should leave this in because the `add_item_to_player` function will require stat functions.

include/game-state/player.h
```diff
@@ -19,8 +19,7 @@ typedef struct player {
     int level;
     int health;
     int xp;
-    stats_hash_t *player_stats;
-    effects_hash_t *player_effects;
+    class_t player_class;
```
sumedhsarwate, line 22: I think this should be a pointer.

include/npc/npc.h
```diff
@@ -23,6 +23,9 @@ typedef struct npc {
     /* long description of the npc, <301 chars */
     char *long_desc;
 
+    /* the npcs class */
+    class_t npc_class;
```
sumedhsarwate, line 27: Also should be a pointer.

include/npc/npc.h
```diff
@@ -23,6 +23,9 @@ typedef struct npc {
...
+
     int health;
```
sumedhsarwate, line 29: Not sure if this is your job, but should this be removed?

carlosazpurua1004: It should, but I don't want to modify NPC's code too much at this point in case they need it for their monkey patch etc, would have to also modify some function headers and calls

include/game-state/player.h
```diff
@@ -19,8 +20,7 @@ typedef struct player {
     int level;
     int health;
     int xp;
-    stats_hash_t *player_stats;
-    effects_hash_t *player_effects;
+    class_t *player_class;
```
ecoble, line 23: Be aware of what changes you might have to make to existing code / PRs because of this

ecoble:
Issue Score: ✔️+
Comments:
Great work here thinking of what dependencies you have! Like issues, PRs should have closing statements too.

PR #775, opened 2020-06-07
Enhancement of NPCs Moving Between Rooms Feature

philipchenaux:
This is an enhancement on our previous branch rpg-npc/npc_room_movement in which we have added the functionality of npcs with a definite movement type to move between rooms, as well as improving the testing across the npc_room module. We have amended the following files
```
include/npc/room-npc.h

src/npc/src/room-npc.c

tests/npc/test_rooms_npc.c
```
In particular we have only added the following function with tests:
```C
int move_npc_def(npc_mov_t *npc_mov)
```
In addition to slightly changing the following function as well:
```C
int reverse_path(npc_mov_t *npc_mov) 
```


anshikab:
I feel like there are multiple data structures handling functionality that could be streamlined (there is a union and an enum for movement type, in addition to a movement struct). This is not something that needs to be addressed right now, but maybe in the future.

anshikab:
In the move_npc_def function, if current_room->next is NULL (i.e the NPC has reached the end of its path), I feel like calling reverse_path limits the opportunity to add additional rooms to the list.

anshikab:
I have added two comments with design considerations, not necessarily to be addressed in this pull request, but perhaps in subsequent issues.

include/npc/rooms-npc.h
```diff
@@ -279,4 +279,18 @@ char* track_room(npc_mov_t *npc_mov);
...
+ * Paramters:
+ * npc_mov: The NPC movement struct
+ * 
+ * Returns:
+ * 0 if move in unsuccessful 
+ * 1 npc has reached the end of the path, reverse_path is called, but 
+ *   the move is not implemented
+ * 2 successful move to the next room
+*/
+int move_npc_def(npc_mov_t *npc_mov);
```
firatciftci, line 294: To be consistent with the overall style of the module, I'll suggest renaming this function to something like `npc_move_definite()`.

firatciftci: Also, looking at the functions again, I'll say that wherever you see `def` and `indef`, you should convert them to `definite` and `indefinite` to improve readability.

philipchenaux:
Just made all of the relevant changes you guys suggested!

firatciftci:
Issue Score: ✔️++

Comments:
Great work!

PR #777, opened 2020-06-07
Custom actions/ast next

emmamedeiros:
This pull request contains the updated AST block struct, which now includes an integer for the number of AST blocks within the struct, as well as a pointer to AST_block_t **next, which points to a list of subsequent AST blocks. This pull request also contains a minor update to the custom action struct, which has added a UT hash handle to aid in storing custom actions.


emmamedeiros:
Thanks Annabelle! I just updated the description so I think it is clearer now!!

src/custom-actions/include/ast_block.h
```diff
@@ -29,10 +29,15 @@ typedef union block {
...
 
+/* Forward declaration */
+typedef struct AST_block AST_block_t;
+
 /* Struct to contain a block, as well as its type */
 typedef struct AST_block {
     block_t *block;
     block_type_t block_type;
+    int num_AST;
+    AST_block_t **next;
```
MaxineK36, line 40: This shouldn't be done with a forward declaration, it should be
`struct AST_block **next`

MaxineK36: Is next going to be used with utlist? If so, why is it a double pointer?

src/custom-actions/src/action_block.c
```diff
@@ -45,7 +45,10 @@ AST_block_t* AST_action_block_new(action_enum_t action_type, int num_args,
     action_block_t *action;
     int new_action;
     block_type_t block_type = ACTION;
-
+    int num_AST = 1;
+    AST_block_t *next;
+    
+    next = malloc(sizeof(AST_block_t));
```
MaxineK36, line 51: If there is no next, you should set it to NULL

src/custom-actions/include/ast_block.h
MaxineK36: Nevermind; I see that it's seems to be used as an array (please correct me if I'm wrong though).
Please rename it to something other than "next" (maybe action_sequence?) if this is the case, since `next` usually indicates that something is being used as a linked list

src/custom-actions/src/ast_block.c
```diff
@@ -74,7 +79,11 @@ int AST_block_free(AST_block_t *ast)
                 branch_block_free(ast->block->branch_block);
             }
     }
-    free(ast->block);
+    if (ast->num_AST > 1) {
```
MaxineK36, line 82: You should make a note in the header for `AST_block_free` (in `ast_block`) that free also frees all the blocks in the sequence

src/custom-actions/src/branch_block.c
```diff
@@ -48,7 +48,10 @@ AST_block_t* AST_branch_block_new(int num_conditionals, conditional_block_t** co
     branch_block_t *branch;
     int new_branch;
     block_type_t block_type = BRANCH;
+    int num_AST = 1;
+    AST_block_t *next;
```
MaxineK36, line 52: If this is being treated as an array, you should malloc enough space for however many AST there are. In this case, maybe next should be null?

src/custom-actions/src/action_block.c
MaxineK36: (since num_AST is 1)

src/custom-actions/src/conditional_block.c
```diff
@@ -46,7 +46,10 @@ AST_block_t* AST_conditional_block_new(conditional_type_t conditional_type,
     conditional_block_t *conditional;
     int new_conditional;
     block_type_t block_type = CONDITIONAL;
-
+    int num_AST = 1;
+    AST_block_t *next;
+  
+    next = malloc(sizeof(AST_block_t));
```
MaxineK36, line 52: same note as before

src/custom-actions/src/control_block.c
```diff
@@ -43,7 +43,10 @@ AST_block_t* AST_control_block_new(control_type_t control_type)
     control_block_t *control;
     int new_control;
     block_type_t block_type = CONTROL;
-
+    int num_AST = 1;
```
MaxineK36, line 46: Same note

tests/custom-actions/test_ast_block.c
```diff
@@ -190,8 +234,13 @@ Test(AST_block_t, free_CONDITIONAL)
...
+    block_type_t block_type2 = CONDITIONAL;
+    int num_AST = 2;
+    int num_AST2 = 1;
+    AST_block_t* test;
+    AST_block_t* next = AST_block_new(block, block_type2, num_AST2, &test);
     
-    ast = AST_block_new(block, block_type);
+    ast = AST_block_new(block, block_type, num_AST, &next);
     
     cr_assert_not_null(ast, "AST_block_new() failed");
```
MaxineK36, line 246: You should have a test case for when num_AST is more than 2. I'm worried that the memory won't be set properly in that case, given that "next" is an array but is often treated as a double pointer to a single element

tests/custom-actions/test_custom_action.c
```diff
@@ -46,15 +50,19 @@ Test(custom_action_t, new_BRANCH)
...
+    
     char *action_name = "act_PUSH";
     char *context = "item";
     char *item = "obj_CHAIR";
     char *type = "paladin";
+    UT_hash_handle hh = hh;
 
     custom_action_t *new_action = custom_action_new(action_name, context, item, 
-                                                    type, head);
+                                                    type, head, hh);
```
annabelleperng, line 65: I realized just now that setting hh in custom_action_new is unnecessary - I think uthash will automatically handle all the hash handles. I don't think hh has to be a parameter in custom_action_new (the only place where it needs to there is in the struct definition for custom_action_t, in custom_action.h).

annabelleperng: It doesn't seem like it's doing any harm there; it just might make more sense for custom_action_new to be called without it.

src/custom-actions/src/ast_block.c
```diff
@@ -33,14 +36,17 @@ AST_block_t* AST_block_new(block_t *block, block_type_t block_type)
...
+                   AST_block_t *next, AST_block_t *prev)
 {
     assert(ast != NULL); 
     assert(block != NULL);
-
+ 
     ast->block = block;
     ast->block_type = block_type;
-    
+    ast->next = NULL;
```
MaxineK36, line 47: These lines are good

tests/custom-actions/test_custom_action.c
emmamedeiros: sounds good, just took it out of new/init functions and just left it in the struct!

emmamedeiros:
This pull request has now been approved, merging the updates to the ast block struct, as well as the addition of a hash handle to the custom action struct into dev. This pull request will now be closed and the custom-actions/ast_next block deleted.

MaxineK36:
Issue Score: ✔️++

Comments:
Great work! The closing summary on this is very clear and concise.

Issue #691, opened 2020-05-27
Implement test skill trees interaction in the chiventure CLI


jackandthebean:
# Implement test skill trees interaction in the chiventure CLI
### What
Now that our final few modules are nearly finished and are on the way to a final `rpg-skilltrees` team pull request, we will use the functions we've written to create a sample skill trees interaction set for the chiventure CLI. We will also pull from `action_management` and `cli` code to do so.
### Why
This test skill trees interaction is a good way to ensure our code works in the greater context of chiventure—beyond unit testing—and to have something to demonstrate in our final presentation.

jackandthebean:
Yesterday, @VannStrasen and I met to discuss our skill trees demo design/implementation and opened the branch [`skilltrees/demo`](https://github.com/uchicago-cs/chiventure/tree/skilltrees/demo) to complete our work

jackandthebean:
@VannStrasen and I have made the PR to merge our complete chiventure demo game that makes use of skill trees. #788 elaborates on the details of the added/changed files. We are also adding a custom command `SKILLS` to print all skills in a given inventory to the CLI.

jackandthebean:
I added a README to describe the Demo game and help those unfamiliar with its creation to navigate it.
https://github.com/uchicago-cs/chiventure/blob/skilltrees/presentation/src/skilltrees/examples/README.md

firatciftci:
Issue Score: ✔️+

Comments:
Great work on this issue! The only thing I will point out is that there should have been more detail put into the meeting update comment; ideally, it should include who the meeting was between, what topics were discussed, and what are the exact outcomes.

PR #781, opened 2020-06-08
Fixing warnings from most recent merge into dev

VannStrasen:
This small change fixes a few typing warnings when trying to make chiventure that were missed by a previous pull request. There is one check inside of one test that is being commented out as it requires a bit more work to fix, but that will be fixed in issue #691 (which will become a PR later tonight/early in the morning tomorrow) in order to prioritize fixing these warnings as soon as possible.


firatciftci:
Issue Score: ✔️++

Comments:
Great work!

PR #792, opened 2020-06-08
Custom actions/interface functions

annabelleperng:
This pull request includes all interface functions besides do_custom_action. The interface allows for storing a custom action into the game (currently into a Hash table in interface.c which will eventually be relocated into game-state/game.h) and searching for previously-defined custom actions by name. It also contains the skeleton functions for translating a custom action from object_t form (having just been parsed from a json file) into a custom_action_t struct.


src/custom-actions/include/interface.h
```diff
@@ -0,0 +1,101 @@
...
+ * (currently represented as a private hash table inside interface.c).
+ * 
+ * Parameters:
+ * - action: The custom action to be added
+ * 
+ * Returns:
+ * - SUCCESS if the custom action was successfully added to the list
+ * - FAILURE if otherwise
+ */
+int *add_custom_action_to_game(custom_action_t *action);
```
brentd313: I think the return type was intended to be
```
int
```
as opposed to
```
int *
```

brentd313: Nevermind. You just fixed this

annabelleperng: Thank you! You're correct, and I've since finished the includes, the checks pass now

PR #793, opened 2020-06-08
Custom Scripts Demonstration Example (customizing the output string for a test action TASTE)

hugo-chan:
We created an in-memory game for our presentation demo.

This code was inspired by last year's CLI team's example game where a new action called `TASTE `is added. We used our `custom_type` module by making the output string of this action dynamic. The value of this output string will be returned from a Lua script which will return different strings depending on the argument passed into Lua. Specifically, if '1' is passed into Lua, the output string will be "You can't taste the rock"; if '2' is passed into Lua, the output string will be "YOU CAN'T TASTE THE ROCK".


src/custom-scripts/examples/demonstration.c
```diff
@@ -0,0 +1,76 @@
...
+
+    /* Create a rock in room1 */
+    item_t *rock = item_new("ROCK","It is a rock.",
+                   "You were hoping this was The Rock but, alas, it is just a plain and ordinary rock");
+    add_item_to_room(room1, rock);
+
+
+    /* Where custom_type comes into play, create a dynamic string (hold different values) depending
+       on what the user enters at the start of the game */
+    char stringNum;
```
firatciftci, line 39: Very minor stylistic thing: You should convert this variable's name to be snake case instead of camel case (so it should be `string_num`).

hugo-chan: Fixed!

firatciftci:
Issue Score: ✔️+

Comments:

PR #776, opened 2020-06-07
Implementation of Basic Functionalities to Play Through a Quest

philipchenaux:
This is a further enhancement of the previous pull request on this branch that initially created the main structs for a quest. This pull request involves creating and playing through a quest. We have changed the following files:
```
include/quests/quest_state.h
src/quests/quests_state.c
tests/quests/test_quests_state.c
```
In this pull request we have included the following main functions:
```C
int add_achievement_to_quest(quest_t *quest, achievement_t *achievement_to_add);
int start_quest(quest_t *quest);
int complete_achievement(quest_t *quest, item_t *item_collected, npc_t *npc_met);
int is_quest_completed(quest_t *quest);
int quest_status(quest_t *quest);
item_t *quest_completed(quest_t *quest);
```
These are mostly still relatively basic function that allow for users to create a very simply quest.

We have also added the demo file for our presentation.

PR #794, opened 2020-06-08
RPG-NPC Quests - Bugfix 

shayna-k:
This is to fix the warnings that were accidentally pushed in the [PR](https://github.com/uchicago-cs/chiventure/pull/776) for the quests example code. The only file changed is the following
```
src/quests/examples/quest-example.c


firatciftci:
Issue Score: ✔️+

Comments:

PR #797, opened 2020-06-08
Action management/conditional rooms

victoriavillalba:
Ok so this pull request adds a list_action_type_t *conditions parameter to the path_t struct so that a path to a room can be initialized with a list of actions that need to be completed before being allowed entry to the room it leads to. This list of actions can be initialized with path_new_conditions() in path.c.
The action_type_t struct now has two new parameters: room_t *room and char *direction, which are necessary to find the path that it's a condition to by using path_search(). If the action isn't a condition to anything then room and direction are both NULL.
There are now two new functions that allow conditions to be removed as they are done. Remove_condition() in path.c with a helper function in game_action.c. The helper function is delete_action(), which deals with the task of removing the node from the linked list.
In order to implement these functions in do_item_action(), do_path_action(), and do_item_item_action(), there are a few more helper functions in game_action.c: find_act(), which finds the specific node with the action that is currently being done, and acetoincmp(), which compares the strings in two action_type_t structs (helper function for find_act()).
The tests have also been added and implemented for these new functions. 
The last thing that changed was an if statement in do_path_action() that verifies if the path can be taken. Another condition was added to make sure that the path does not have any pre-requisite actions listed as necessary before taking the path.


include/game-state/game_action.h
```diff
@@ -109,6 +110,44 @@ int add_action_inventory_condition(game_action_t *action, player_t *player,
...
 
+/*
+ * compares action strings
+ *
+ * Parameters:
+ * - list_action_type_t: points to node with one action
+ * - list_action_type_t: points to node of another action
+ *
+ * Returns:
+ * - int value that acts like boolean
```
MaxineK36, line 121: you should specify the convention here (0 if equal, same convention as strcmp)

src/action_management/src/actionmanagement.c
```diff
@@ -124,6 +136,24 @@ int do_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *i, char **ret_
         }
         else
         {
+	    //remove action from any conditions
+	    path_t *closed_path;
+            closed_path = path_search(a->room,a->direction);
+	    // only if action is a condition to something (action with
```
MaxineK36, line 142: this hsould be a block comment

src/action_management/src/actionmanagement.c
```diff
@@ -124,6 +136,24 @@ int do_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *i, char **ret_
...
+		closed_path = path_search(a->room,a->direction);
+		delete_node = find_act(closed_path->conditions,a);
+		condition = remove_condition(closed_path,delete_node);
+		if (condition != SUCCESS)
+		{
+	  	    sprintf(string, "condition was not removed");
+	 	}
+	    }
+
             // successfully carried out action
```
MaxineK36, line 157: block comment

src/action_management/src/actionmanagement.c
MaxineK36: (i.e. `/* comment here */`)

src/action_management/src/actionmanagement.c
```diff
@@ -166,7 +196,8 @@ int do_path_action(chiventure_ctx_t *c, action_type_t *a, path_t *p, char **ret_
         return WRONG_KIND;
     }
     // validate existence of path and destination
-    if ((path_found == NULL) || (room_dest == NULL))
+    // third condition checks if conditions have been met
```
MaxineK36, line 199: should be a block comment

src/action_management/src/actionmanagement.c
```diff
@@ -176,6 +207,25 @@ int do_path_action(chiventure_ctx_t *c, action_type_t *a, path_t *p, char **ret_
     /* PERFORM ACTION */
     int move = move_room(g, room_dest);
 
+    //remove action from any conditions
```
MaxineK36, line 210: same here

MaxineK36: in general, anything that's a comment above a line that refers to multiple lines of code should be ab block comment

src/action_management/src/actionmanagement.c
```diff
@@ -271,6 +321,24 @@ int do_item_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *direct,
         }
         else if (applied_effect == SUCCESS)
         {
+            //remove action from any conditions
```
MaxineK36, line 324: This chunk of code (getting closed path, etc) is the same in all three parts; it should be a helper. you can keep it locally in actionmanagement.c, no need to add it to a header file, just put a header for the function

src/game-state/src/game_action.c
```diff
@@ -147,6 +149,37 @@ int add_action_condition(game_action_t *action, condition_t *condition)
     return SUCCESS;
 }
 
+/* See game_action.h */
```
MaxineK36, line 152: this is a function used only locally, so you should take the header out of game_action.h and just put the header comment here

victoriavillalba: For game_action_init()?

MaxineK36:
Issue Score: ✔️+

Comments:
Good work overall! One thing for future issues: When a PR is merged, you should include a final comment summarizing what was included.

Issue #529, opened 2020-05-11
Parse the zipped files into our representation of objects.


dtbukowski:
Communicate with Nam with how the .wdz files look, then parse those files into the representation of objects that blake and noah decided on.

dtbukowski:
Made progress on the current state of libzip in the CS Machines with a commit on wdl/parsing_zipped (thought the branch doesn't show up and I may need to push it). Although share objects exist in repositories /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libzip.so/usr/lib/jvm/java-11-openjdk-amd64/lib/libzip.so
/usr/lib/vlc/plugins/access/libzip_plugin.so, I cannot find the zip.h file necessary to complete. There is either a mistake in my Makefile or libzip needs to be installed in usr/lib/x86_64-linux-gnu, so I will use zziplib for now and try asking techstaff to download libzip

dtbukowski:
Was able to successfully open and close a zip file since I finally had the right libraries for zip.h and the header installed in the Virtual Machine. Libzip isn't installed on the CS machines so as of now I am not able to use libzip on the CS machines.

Will push my progress once I am able to fix errors pushing onto Github. Apparently there are errors with remotes.

dtbukowski:
Pushed progress on using libzip where I figured out how to get file indexes and get files from name (has to be specific directory) or index. 

Next I will use Noah's implementation to make JSON objects and push that onto master

dtbukowski:
I have pushed code that will make JSON objects, but specifically only for the 5 JSON files in the sandbox in wdl in parsed_zipped directory. 

However, the errors present are because of JSON not being present in the VMs so I will either install JSON myself tonight or, depending on how fast techstaff acts, use my own computer with libzip installed.

I have implemented a way to loop through all files in a zip directory, however I will need to figure out how to:
Get files that end in .json (check if filename ends in .json)
Get files that are images and files that are sounds (looked into it here: https://stackoverflow.com/questions/670546/determine-if-file-is-an-image) 
Loop through all JSON objects in a .json file (current implementation of json_parse_test that uses the name of the field will not work because new field will require this file to constantly be updating.)

dtbukowski:
Link to code: https://github.com/uchicago-cs/chiventure/tree/wdl/parsing_zipped/src/wdl/sandbox

dtbukowski:
[...]

PR #798, opened 2020-06-08
Implemented .wdz zip file opening and verified that the JSON files can be read

namanhd1nh:
~~The branch implements some basic wdz file loading and prints out the contents of the JSON files contained inside to the terminal behind chiventure's main UI. This serves as both a demo for the team, a first introduction of libzip and json-c dependencies to the chiventure codebase, and a proof of concept that the .wdz file loading functionality is on the right track.~~

~~This does not, however, convert the parsed JSON objects into game objects, which should be implemented later. (in branch `wdl/parsing-wdz-namanh`).~~

~~Addresses #529 but does not fulfill it completely. We will make a separate backlog issue for final conversion of json objects into game objects.~~


namanhd1nh:
PR opened by mistake; this was supposed to be merging branch `wdl/parsing-wdz-stabledemo`, not `-namanh`.

PR #803, opened 2020-06-08
Initial implementation for player-controlled ball/cursor

seanrryan:
This PR builds off of what was merged a few days ago in #747.

I've added a few lines to support a new object in the map window that can serve as a cursor or simply a way to bring attention to a certain part of the screen. It supports arrow keys and WASD as the input keys for moving it around on the screen. It currently does not interact with any of the other polygons, such as the player or the rooms implemented by @damonhines but the "cursor" is continuously drawn and redrawn to accurately respond to feedback so it can be checked whether or not it lies within/on top of certain regions later on.

I decided to first implement keyboard-controlled cursor support in keeping with the spirit of text-adventure games. WASD controls allow for the user to keep both hands on the keyboard without even needing to think about the mouse. However, raylib has support for mouse input if we want to eventually implement mouse controls down the road.

Here's how to see the example, starting from the `chiventure` directory:
`cd build`
`make`
`make examples`
`src/ui/examples/./map_raylib_example`

Once the window opens, the arrow keys should move the small white circle on the screen. There are two control schemes:
- Arrow keys: up, down, left, right
- WASD
    - W: up
    - A: left
    - S: down
    - D: right

The player circle was changed to lightt gray to prevent confusion with the white cursor, which should be visible against the colored rooms and the black background.


src/ui/examples/map_raylib_example.c
```diff
@@ -57,7 +73,7 @@ int main(void)
             radius = 15;
 
             DrawRectangle(screenWidth/4*2 - 90, 170, roomWidth, roomHeight, RED);
-            DrawCircle(centerXPos, centerYPos, radius, WHITE);
+            DrawCircle(centerXPos, centerYPos, radius, DARKBLUE);
```
xmvalencia, line 76: this is minor but maybe change the indicator to LIGHTGRAY, it provides better contrast and I think it'll make it easier to distinguish if it was in the first room. this is also a pure aesthetic choice

seanrryan: I agree. That would provide a better contrast. Thanks, changing it now

src/ui/examples/map_raylib_example.c
```diff
@@ -44,6 +48,18 @@ int main(void)
 
     while (!WindowShouldClose())   
     {
+
+        /* Updates ball cursor with arrow keys or WASD*/
+        if (IsKeyDown(KEY_RIGHT)) cursorBallPosition.x += 5.0f;
```
ecoble, line 53: Can you make 5.0f a variable? maybe "step" or "move_unit" or something

src/ui/examples/map_raylib_example.c
```diff
@@ -89,10 +105,13 @@ int main(void)
             DrawRectangle(posX, posY, roomWidth, roomHeight, GRAY);       
         }
 
+        /* Draws the ball cursor in the current position according to user input */
+        DrawCircleV(cursorBallPosition, 8, cursorBallColor);
```
ecoble, line 109: Variable to define what 8 is too

seanrryan: just added a cursor radius variable!

src/ui/examples/map_raylib_example.c
seanrryan: just fixed! "move_unit" is now defined up with the rest of the cursor variables

seanrryan:
To conclude, this PR is a small addition to our existing raylib map example to include a cursor object that can be controlled by the keyboard, which exists as an object above the other shapes generated by the map. With the arrow keys or WASD, the user moves the ball around the map. The coordinates of the cursor are updated to reflect user input, so in the future function can be called to retrieve the location of the cursor (i.e. the center of the circle) from the map environment.

ecoble:
Issue Score: ✔️++
Comments:
Great job on this PR, especially with such a thorough summary!

PR #804, opened 2020-06-08
Merge in a redesign of Chiventure code so that the new object interfaces can be used to load games while still maintaining use of old WDL

### LINKED ISSUE #689: Implement the new interfaces for accessing libobj to create the game_t and room_t objects in game-state and use as example code
The final task for setting up the new libobj interface is to provide an example of how to use these new interface and objects to create something saved in the game-state. This code should implement the interface to take the global game-state object and set up the basic game-object and then create room-objects from libobj objects. This may, depending on the current status of other teams, involve working with other teams, but theoretically could be written properly without need for much cross-team consultation.

This will serve two functions. First, the code may serve as an example for other teams to be able to implement their own use of the interface. Second, this will allow us to replace the previous libobj's implementation that would set up the game and rooms directly from the wdl document.

noahklowden:
This merge adds a `load_objects` module within `common` that is designed to convert WDL objects into game object. Unlike the old WDL parsing, which had all functions for converting objects to the game within WDL code, this module is designed to allow for all teams to be able to access it so that the WDL team does not need to be directly involved in changes to object to game conversion code.

This meant splitting the `load_wdl()` function within `load_game` into two parts: one that parses a WDL file or WDZ file into `obj_t`s and `object_t`s respectively, and then another (`load_objects()`) that converts those into a game.

closes #689


include/wdl/wdl_ctx.h
```diff
@@ -0,0 +1,53 @@
+#include "wdl/wdl_common.h"
+#include "wdl/objstore.h"
+#include "game-state/game.h"
+
+#ifndef INCLUDE_WDL_CTX_H
+#define INCLUDE_WDL_CTX_H
+
+typedef struct wdl_ctx
+{
+    //a hashtable of objstores containing all parsed objects from a wdl file
```
MaxineK36, line 10: these struct comments should be block comments

noahklowden:
With the merge complete, once parsing is finished, it should be possible for teams to use `load_objects()` to load in various aspects of the game into the `game_t` object.

Moving forward, it is advised the WDL team consider putting together a new `add_rooms_to_game()` function for WDZ to demonstrate this functionality. Otherwise, this could also be taken up by a future game game sate.

MaxineK36:
Issue Score: ✔️++

Comments:
Great work! The closing summary is particularly well done.

Issue #813, opened 2021-01-30
Design and implement new object store


borjasotomayor:
Based on the experiences from the last two years, it has become clear that chiventure needs a unified object store where all the various components can store their data. This will largely involve the following:

- Flesh out the WDL++ format developed in Spring 2020
- Create an in-memory object store that has a nearly 1:1 correspondence with the WDL++ file format.
- Design an API for accessing/modifying the data from other components.

dwahme:
Below is a summary of the existing WDL++/WDZ documentation and some proposed updates/clarifications to how the directory structure is handled such that it can be cleanly coupled with a revamped libobj object store. The following is subject to updates and review


## Existing WDL++/WDZ Format Docuemntation Summary

The following is a compilation of all documentation for WDL++

### WDZ Sandbox

Source code from- https://github.com/uchicago-cs/chiventure/tree/dev/src/wdl/sandbox

Characteristics:
- Appears to be a normal single JSON file- just need to support all common JSON functionality
- Had the major categories like `"ROOMS"`, `"GAME"`, etc defined directly in the file


### WDL++ Design Doc

Reference- https://github.com/uchicago-cs/chiventure/wiki/What-WDL---will-look-like

Characteristics:
- Appears to be normal single JSON file- just need to support all common JSON functionality


### WDL++/WDZ JSON Directory Structure Documentation

Reference- https://github.com/uchicago-cs/chiventure/wiki/WDL:-Looking-into-options-to-store-JSON-objects

Characteristics:
- Use a zip folder to hold multiple JSON files
- Appears to want to support multiple directories
- Wants to store media such as images and audio inside the zip folder
- Needs to be able to support multiple files corresponding to a major category to support game "libraries" (such as plug and play RPG elements like a base monster set, base spell set, etc that can easily be added in to an existing game)
- Wants to store games zipped instead of unzipped


## Proposed WDL++/WDZ Requirements (and Rules for Parsing into Object Store)

Characteristics
- WDZ is a zip folder containing the following:
    - Some form of "main" or "default" file (for single-file games or small components like "ITEMS")
    - JSON files for a given component (`./ROOMS.json` gets translated to the `"ROOMS"` component in the object store)
    - Folders for further component groupings (`./ITEMS/staff.json` gets translated to `"ITEMS.staff"`)
    - Media (maybe in a media folder?)- needs to be able to be referenced from any JSON file
- Maybe allow single-file JSON files as WDL++ so that they don't have to be zipped?

[...]

PR #814, opened 2021-03-19
Libobj 3.0 and JSON to obj_t

dwahme:
Related to #813

This PR contains code for Libobj 3.0 and a way to load JSON into an obj_t struct.

Changes:
- Added a sample WDZ-format game at `games/test_game`
- Revamped libobj for version 3.0:
    - Added a LIST datatype for doubly-linked lists
    - Added an OBJ datatype (to differentiate from the NONE type)
    - Refactoring of the obj struct for how children were held and added utlist doubly-linked list support
    - Bug fixes for libobj
    - Removed unnecessary functions from libobj
    - Added tests for libobj
- Refactored a bit of the wdl module:
    - Updated some code for libobj 3.0 behavior
    - Updated tests for compatibility with libobj 3.0
    - Updated some object iteration behavior to use utlist and uthash macros for safety
- Added a way to load JSON into an obj_t along with tests


include/libobj/load_json.h
```diff
@@ -0,0 +1,36 @@
...
+ * 
+ * Recursive helper for load_obj_json. 
+ * 
+ * Parameters:
+ *  - path: The path to the JSON file
+ * 
+ * Returns:
+ *  - EXIT_SUCCESS on success, EXIT_FAILURE on failure
+ */
+int _load_obj_json(obj_t *parent, struct json_object *json);
```
borjasotomayor, line 21: The correct way of exposing this function is to add a header file in `src/libobj/include` and then include that from the tests. You can see an example of this here: https://github.com/uchicago-cs/chiventure/blob/dev/tests/custom-actions/CMakeLists.txt#L17

include/libobj/obj.h
```diff
@@ -30,38 +35,30 @@ typedef struct obj
...
     } data;
 
-    // The object's attributes (a hash table)
-    struct obj *attr;
+    // For if the obj is part of a list
+    struct obj *next;
+    struct obj *prev;
     
     // Required uthash identifier for making the hash table
     UT_hash_handle hh;
```
borjasotomayor, line 56: The more correct way of writing this union would be like this:

```
    union
    {
        bool b;
        char c;
        int  i;
        char *s;
        struct {
            struct obj *attr;
            UT_hash_handle hh;
        } obj;
        struct {
            struct obj *head;
            struct obj *next;
            struct obj *prev;
       } lst;
    } data;    
```

This way, the `data` union contains all the information associated with each type. Making this change would require using the utlist and uthash macros for structs with non-default next/prev and hh fields. If making this change is easy, we should do it now. If not, it doesn't have to block this PR, but it has to be changed at some point.

borjasotomayor (changes requested):
This looks good overall. A few overarching comments:

- load_json.h should be renamed to load.h. I think it's ok for the implementation to have different load_* modules (since there will be separate code for loading from a ZIP file, etc.) but, from the user's perspective, it makes more sense to think of there being a single "load.h" header with all the functions for loading from disk.
- Do parser.h and stack.h actually have to be public header files? They seem internal to libobj so, if they're not used anywhere else, they should be moved to `src/libobj/include`
- The examples in `src/libobj/examples` will need to be updated at some point, but that doesn't have to be part of this PR.

include/libobj/load_json.h
dwahme: Done

include/libobj/obj.h
dwahme: Regarding the hh change: Unless I'm misunderstanding the uthash library, I think that the hh field needs to be present in all objects, as the child nodes need to have hh (instead of the parents).

Regarding the next/prev fields: I played around with it for a bit but I didn't see a clean and easy way to shift over to nesting them in the `data` union. I think it might be better to hold off on making the change now.

dwahme:
- The requested load* changes have been made.
- WDL still depends on parser.h and stack.h. I think it might be better to not touch these for now, and then when we're reworking some of WDL for the new libobj/wdz parsing, we can go in and clean those up.
- I did a quick touchup of the obj example to reflect the new changes

include/libobj/obj.h
borjasotomayor: Sounds good, let's make this change later. That said, could we change the union to something like this?
```
    union
    {
        bool b;
        char c;
        int  i;
        char *s;
        struct {
            struct obj *attr;
        } obj;
        struct obj *lst;
    } data;    
```

That makes it clearer that "there is an object type of data, and that type has one piece of information: its attributes". This also future-proofs this a bit, in case we end up having to associate more information with an object type.

borjasotomayor:
The changes look good to me, and I agree that we can defer moving parser.h and stack.h until we rework the WDL code. I have just one more requested change (see my comment about the `data` union)

include/libobj/obj.h
dwahme: I think that this is a good solution. Done

dwahme:
Data change comment addressed, this should be ready for re-review

borjasotomayor (approved):

PR #815, opened 2021-04-08
Adding Zip Parsing to Libobj 3.0

dwahme:
Related to #813

This PR contains code to read a zip file into an obj, in addition to a few minor changes.

Changes:
- Zip parsing
    - Added a function to load a zip file into an obj from a given path
    - Added functionality to load DEFAULT.json files directly into the object (bypassing storing them inside the filename)
    - Added tests for loading the zip file
    - Added test files for the zip tests
- Refactoring/touchups
    - Renamed internal hidden helper file load_json.h to load_internal.h and added zip loading helpers
    - Minor updates to libibj 3.0 documentation
    - Fixed dump_obj layout for displaying lists


borjasotomayor:
This looks good to me, but it looks like, despìte passing all the tests in CI, the ZIP tests don't pass on the CS Linux machines:

```
[----] /home/borja/repos/chiventure/tests/libobj/test_load_wdz.c:259: Assertion failed: Could not close zip file; check archive code
[FAIL] test_load_wdz::zip_default: (0.00s)
[RUN ] test_load_wdz::zip_multiple
[----] /home/borja/repos/chiventure/tests/libobj/test_load_wdz.c:334: Assertion failed: Could not close zip file; check archive code
[FAIL] test_load_wdz::zip_multiple: (0.00s)
[RUN ] test_load_wdz::zip_nested
[----] /home/borja/repos/chiventure/tests/libobj/test_load_wdz.c:212: Assertion failed: Could not close zip file; check archive code
[FAIL] test_load_wdz::zip_nested: (0.00s)
[RUN ] test_load_wdz::zip_simple
[----] /home/borja/repos/chiventure/tests/libobj/test_load_wdz.c:165: Assertion failed: Could not close zip file; check archive code
[FAIL] test_load_wdz::zip_simple: (0.00s)
```

I get the same result when running the tests on my own Ubuntu 20.04 machine. Any idea what could be causing this?

dwahme:
Hmm, that's strange; it was working fine on my ubuntu 20.04 machine. I'll dig into it in the next day or so. From here, I see two ways to solve this:
1. Debug exactly what's going wrong with closing the zip file on the CS linuz machines. If I remember correctly, closing the zip is what saves the file to disk. In the tests, this would be the first time the file is saved. It might just be a libzip setting thing or a linux thing; I'm not quite sure.
2. Refactor a bit of the library and the tests (which is something I was thinking about doing before anyways). At the moment, `load_obj_zip` takes in a `char *` path to the file in question, and then it loads the zip into memory in the function. We could modify this to take in a `zip_t *` instead, and leave putting the zip into memory to the caller of `load_obj_zip` instead (this is just done by calling `zip_open`). This has the added benefit of allowing a zip to be built in memory and then sent in to `load_obj_zip`directly  instead of having to save, close, and reopen. This would simplify the tests and probably make them more flexible for cross-platform stuff, as right now each of the tests needs the additional boilerplate of 1) creating a new temporary folder, 2) building the zip and then writing that to disk (so that it can be loaded later), and 3) deleting that folder as cleanup at the end of the test.

I personally think option 2 might be the better choice, but I will look into the error on the CS Linux machines a bit to see what's up. However, let me know if you think option 1 is better long term.

dwahme:
After doing some testing, it looks like it's an issue with creating folders on the fly during the tests. Since this will probably cause issues when moving to different platforms, I'm going to switch to option 2 instead.

dwahme:
I've restructured the tests and updated the `load_zip` definition to take in a zip file now. I've also verified that it works locally on my machine and the CS Linux machines (and it looks like it's passing on CI now).

PR #816, opened 2021-04-11
General code cleanup for Spring 2021

borjasotomayor:
This pull request includes a number of changes to tidy up the code for the Spring 2021 offering of CMSC 22000:

- Removed the checkpointing module and the dependency on protocol buffers. We are moving towards a low-level game representation (WDL++) where doing a dump of the game state in that same format will be much easier than with the original WDL.
- Made Lua an optional dependency, and introduced the notion of "optional components" that don't get built if an optional dependency is not met.
- Updated the dependencies documentation, with concise instructions for Ubuntu 20.04 and Ubuntu 16.04.
- Cleaned up the top-level games/ directory
- Small fixes here and there


docs/dependencies.md
```diff
@@ -3,101 +3,60 @@
...
-sudo apt-get install libncurses5-dev libncursesw5-dev
+sudo apt-get install apt-transport-https ca-certificates gnupg software-properties-common wget
+wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | sudo tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null
+sudo apt-add-repository 'deb https://apt.kitware.com/ubuntu/ xenial main'
+sudo apt-get update
+sudo apt-get install cmake
 ```
 
-## Installing `criterion`
+Criterion must be installed from their PPA:
```
ecoble, line 54: Maybe put in parenthesis what PPA stands for?

ecoble (approved):
I have one small suggestion but otherwise LGTM!

ecoble:
Also just a general question, are we removing check pointing because it doesn't work?

docs/dependencies.md
borjasotomayor: Fixed.

borjasotomayor:
It is being removed largely because we are transitioning to having information stored in an in-memory "object store" that has a practically 1-to-1 correspondence with a JSON-based format (WDL++), which means we don't need a separate format (like protocol buffers) to checkpoint the state of a game. That means that, with WDL++, you can checkpoint a game simply by dumping the entire object store to a WDL++ file (although this could be optimized further by detecting what values in the object store changed since the game was loaded, and only dumping those). Additionally, I'm trying to reduce the number of dependencies that chiventure has (including dividing dependencies into "required" and "optional"), and this has the added benefit of removing a dependency on protocol buffers.

PR #827, opened 2021-04-23
Updating WDL to Use Libobj 3.0

dwahme:
Related to #813

This refreshes WDL to use the new WDZ parsing functionality built in libobj 3.0 instead of the legacy YAML file parsing. It also has some minor updates to parsing the format to simplify and clarify writing WDZ files.

Changes:
- Libobj
    - Removed legacy `parser.*` and `stack.*`
    - Removed dependencies on libyaml
- WDL module
    - Swapped out legacy libobj yaml to obj parser with the libobj 3.0 wdz to obj parser
    - Swapped the spec from having a list of rooms to a dictionary of rooms (a room can now be found by `obj_get_attr(game, "ROOMS.<room_id>")` instead of iterating through a linked list of anonymous objects that have an id field)
    - Removed parsing redundant "id" field from the room object (this can be found with just `obj->id`)
    - Swapped the spec from having a list of items to a dictionary of items (an item can now be found by `obj_get_attr(game, "ITEMS.<item_id>")` instead of iterating through a linked list of anonymous objects that have an id field)
    - Removed parsing redundant "id" field from the item object (this can be found with just `obj->id`)
    - Swapped game end conditions to be a dictionary specifying conditions instead of a single-element list
- WDL tests
    - Updated tests to use new spec changes
    - Recreated the old YAML test files as a WDZ file
    - Added WDZ building/parsing logic to build the zip to later be parsed
    - Other minor simplifications/cleanups


borjasotomayor:
I've made a first pass through the changes, and noticed that the tests were passing with the provided `tests/wdl/examples/DEFAULT.json` file, but that this failed:

```
$ ./chiventure ../tests/wdl/examples/DEFAULT.json 
game object not found
```

It looks like there were some exit() calls that made it look like the `game/create_game` test passing (which was actually never getting to the assertion that checks whether the game was loaded correctly). See https://github.com/uchicago-cs/chiventure/pull/827/commits/c2b60167178441d46e6adb2102fc93e86d32691c for more details.

Looking through the code, it looks like `get_doc_obj` assumes that the provided path is always a ZIP file. In general, it looks like some of the old object code is not yet fully disentangled from the new code. I think what we need is a high-level game-loading function that takes a path as a parameter, and does the following:

- If the path is a `.wdl` file, treat it as a single JSON file and load that one JSON file into the object store (if I understand correctly, this is equivalent to having a ZIP file with a single DEFAULT.json file in it)
- If the path is a `.wdz` file, treat it as a ZIP file, and load the contents of that file into the object store.
- If the path is a directory, treat it as an uncompressed ZIP file (i.e., load everything in that directory as if they were files inside a ZIP file)

In all three cases, all we're doing is loading into the object store. Loading the object store into the game state structs should happen in a separate function.

I've reorganized the example files a bit to make it easier to test this both from the automated tests and when running chiventure from the command-line:

-  `tests/wdl/examples/wdl/` should contain only standalone WDL files (in the new JSON format). It currently contains a single `connected-rooms.wdl` file.
- `tests/wdl/examples/wdz` should contain either `.wdz` files or directories representing the contents of `.wdz` files. It currently contains a single directory `connected-rooms` with a `DEFAULT.json` file (which is actually just a symlink to `connected-rooms.wdl`)

Ultimately, we need to be able to run chiventure like this:

```
$ chiventure tests/wdl/examples/wdl/connected-rooms.wdl
```

```
$ chiventure tests/wdl/examples/wdz/connected-rooms/
```

```
$ chiventure tests/wdl/examples/wdz/connected-rooms.wdz
```

(this last file doesn't yet exist, but would just be the ZIP'd contents of `tests/wdl/examples/wdz/connected-rooms/`)

To get this pull request into `dev`, I think it's enough to support the first one, as we're still working on polishing up the WDZ implementation.

borjasotomayor:
Separate from the above issue, I've also removed references to YAML in the code (including sample files) to avoid confusion about what file format(s) are supported in chiventure. This is not to say we can't reintroduce the YAML format later on as something that gets "compiled down" into the new JSON format, but we should keep that YAML out of the repository until that happens.

dwahme:
I've fixed the hidden parsing bugs and rewired some of the logic so that chiventure can detect the file type and parse the game accordingly by adding a high level function called load_obj() that takes in a filepath and handles it based on extension (so now WDL only has to call this).

Changes:
- Added load_obj() to take in a path to an arbitrary extension and parse it accordingly
    - Did a bit of rewiring inside libobj and WDL to accomodate the changes
- Renamed the test file from *.json to *.wdl
- Renamed the WDL test files from *.json to *.wdl

I was successfully able to play and win the game when running
```
$ ./chiventure ../tests/wdl/examples/wdl/connected-rooms.wdl
```

borjasotomayor:
Now that chiventure can actually runs the JSON files, I think this is good to go as far as merging it into dev. However, I think the code still needs some structural changes (although I don't think they should block this PR). I'll create a separate issue for this, and will then approve this PR.

Issue #615, opened 2020-05-20
Change class-based item attributes to use linked-lists


simrenk:
Edit functions for adding class item restrictions + checking whether a class is restricted from item use to be linked-list based. Dependent on the implementation and approval of game-state functionality for attributes being set to linked lists

ideclue:
Currently, the mechanism for adding and checking **class item restrictions** (#614) is outlined in [https://github.com/uchicago-cs/chiventure/blob/dev/include/playerclass/class_item_restriction.h](url). The two methods include functionality for:
- Determining if a class is restricted from using an item, given the class and item
- Adding an item restriction to a class, given the class and item 
This process currently uses the `set_bool_attr` and `set_bool_attr` methods in the Game State module. However, it could be streamlined through the use of a list-based control system, where items can have multiple class restrictions (as outlined in #624).

ideclue:
Spoke with @LapsededPacifist to discuss the `attribute_list_t` struct that is currently being built in the Game State module (#624). This struct will provide the functionality needed to add a list of class restrictions (attributes) to a single item.

In order to modify the current class item restriction framework to use `attribute_list_t`, the following will be necessary:

- Change the methods in `class_item_restriction.c` to use the `attribute_list_t` struct
- Include additional options beyond the basic getter/setter for the game developers to view what other player classes are restricted from certain items

To implement these changes, I will have to work in a new branch built off of the `action_management/attribute_list` branch which contains the current working code for the `attribute_list_t` struct.

ideclue:
Reviewed the unmerged changes in the `rpg-playerclass/class-item-restriction` branch in order to check which of these changes are relevant to this issue. Most of the unmerged changes are deletions from the Action Management module, and will be largely irrelevant to this issue. However, an important addition I noticed were the pre-written test cases testing class item restrictions. These will be useful when testing my new implementation using attribute lists.

ideclue:
86d92d0928b33e4d81a31a0732f85ad506f274e3

ideclue:
Created a new branch `playerclass/class_item_restriction_lists` off of `action_management/attribute_list` to implement the changes listed above. `item` now includes a `class_restrictions` field of type `attribute_list_t` that is a linked list representing all class restrictions on a certain item. This way, class restrictions are easily accessible and are managed outside of the current framework set up for item attributes. Remaining elements of this task include:
- Adding a function to retrieve all class restrictions from a certain item
- Adding a function to remove a class restriction from an item
[...]

PR #846, opened 2021-04-26
Implement Attribute Linked List Functionality

### LINKED ISSUE #624: Creating functionality for setting attributes to linked lists in game-state (so that #615 can be implemented)
Adding a linked list struct (and associated malloc, free, init functions) and a set_linkedlist_attr function in game-state/item.c and item.h so that rpg-playerclass can eventually set items to have multiple class restrictions using just one attribute (issue #615)

jasulcaf:
(Closes #624)
This is connected to @LapsededPacifist and my Issue (#624) which set out to create a linked list struct that could contain multiple attributes alongside functions to manipulate said list. The eventual goal for this functionality would be related to Issue(#615) for the eventual goal of having multiple restrictions placed on one item. 

We discovered that there already existed a struct for this `attribute_list_t` as well as an associated delete function called `delete_attribute_llist`. As such, below are the changes we implemented.

Overview of Changes: 
- Implement `create_list_attribute`, to return an empty list of attributes that can be manipulated.
- Implement `list_contains_attribute`, which checks a list to see if a given attribute is found.
- Implement `add_attribute_to_list` and `remove_attribute_from_list` to add or remove an attribute to a list.


WPDolan (changes requested):
Looks good! My only suggestion is to include a few lines in `create_list_attribute` to handle cases where mallocing rv returns NULL (when memory is full) to prevent segfaults when setting `rv->attribute` and `rv->next`

WPDolan (approved):
Everything looks good!

src/game-state/src/item.c
```diff
@@ -571,6 +571,87 @@ int delete_attribute_llist(attribute_list_t *head)
...
+    }
+
+    attribute_list_t *tmp = head->next;
+  
+    /* Checks the continous case of having x number of attributes in given list */
+    while(tmp != NULL)
+    {
+        if(strcmp(tmp->attribute->attribute_key, attr->attribute_key) == 0)
+        {
+            LL_DELETE(head, tmp);
```
afranzese99, line 645: Does the LL_DELETE free the attribute that is being removed?

MaxineK36: @afranzese99 @LapsededPacifist Nope, it does not! See here for more details about utlist, and some example use cases that illustrate that you want to call `LL_DELETE` and *then* `free`: https://troydhanson.github.io/uthash/utlist.html#_using_utlist

afranzese99 (approved):
After the addition of freeing the element that was being deleted from the linked list everything looks good.

include/game-state/item.h
```diff
@@ -371,6 +371,49 @@ attribute_list_t *get_all_attributes(item_t *item);
...
+ * 
+ * Parameters:
+ *  linked list of pointers to attributes
+ *  name of a specific attribute
+ *
+ * Returns:
+ *  FAILURE for does not contain, SUCCESS for contains
+ */
+
+int list_contains_attribute(attribute_list_t *head, char *attr_name);
```
MaxineK36, line 391: `SUCCESS` and `FAILURE` are status codes; since this is explicitly checking whether or not an attribute is contained in the list (essentially answering a yes/no question) it should return a boolean value

include/game-state/item.h
```diff
@@ -371,6 +371,49 @@ attribute_list_t *get_all_attributes(item_t *item);
...
+
+/*
+ * Function to add an attribute to a linked listed (utlist) of attributes
+ *
+ * Paramaters:
+ *  Linked list of pointers to attributes
+ *  An attribute to add to the list
+ * 
+ * Returns:
+ *  FAILURE if it failed to add the attribute, SUCCESS if it add/the list contains the attribute
```
MaxineK36, line 401: nit: *adds

src/game-state/src/item.c
```diff
@@ -566,11 +566,93 @@ int delete_attribute_llist(attribute_list_t *head)
...
+	free(elt);
     }
     return SUCCESS;
 }
 
+/* See item.h */
+attribute_list_t* create_list_attribute()
+{
+    attribute_list_t* rv = malloc(sizeof(attribute_list_t));
+    if(rv == NULL)
```
MaxineK36, line 578: nit: put a space after `if` (below as well)

src/game-state/src/item.c
```diff
@@ -566,11 +566,93 @@ int delete_attribute_llist(attribute_list_t *head)
...
+    }
+
+    rv->attribute = NULL;
+    rv->next = NULL;
+
+    return rv; 
+}
+
+/* See item.h */
+int list_contains_attribute(attribute_list_t *head, char* attr_name)
```
MaxineK36, line 590: This will become `bool`

src/game-state/src/item.c
```diff
@@ -566,11 +566,93 @@ int delete_attribute_llist(attribute_list_t *head)
...
+{
+    if(attr_name == NULL || head->next == NULL)
+    {
+        return FAILURE;
+    }
+    attribute_list_t *tmp = head->next;
```
MaxineK36, line 596: You should be able to do this using the built-in `LL_SEARCH` function for utlist: see an example here https://github.com/uchicago-cs/chiventure/blob/e9d9e8e66977a5a76a358abd83a300bba6d65dbc/src/game-state/src/game_action.c#L174

src/game-state/src/item.c
```diff
@@ -566,11 +566,93 @@ int delete_attribute_llist(attribute_list_t *head)
...
+    {
+        return FAILURE;
+    }
+
+    /* Create the to-be appended struct of new attribute */
+    attribute_list_t *tmp = create_list_attribute(); 
+    tmp->attribute = attr;
+    tmp->next = NULL;
+
+    if(list_contains_attribute(head, attr->attribute_key) == SUCCESS)
```
MaxineK36, line 621: this if statement should happen before you call `create_list_attribute`

src/game-state/src/item.c
```diff
@@ -566,11 +566,93 @@ int delete_attribute_llist(attribute_list_t *head)
...
+{
+    if(attr == NULL || head->next == NULL)
+    {
+        return FAILURE;
+    }
+
+    attribute_list_t *tmp = head->next;
+  
+    /* Checks the continous case of having x number of attributes in given list */
+    while(tmp != NULL)
```
MaxineK36, line 641: You should use LL_SEARCH here as well

MaxineK36 (changes requested):
A couple of individual things to change, but the general design is great!

src/game-state/src/item.c
```diff
@@ -566,11 +566,111 @@ int delete_attribute_llist(attribute_list_t *head)
...
+/* See item.h */
+int add_attribute_to_list(attribute_list_t *head, attribute_t *attr)
+{
+    if (attr == NULL)
+    {
+        return FAILURE;
+    }
+    
+    /* General Case where there could be n elements in the list */
+    if (list_contains_attribute(head, attr->attribute_key) == true)
```
MaxineK36, line 633: nit : you don't need the `== true`

MaxineK36 (approved):
One tiny nit, otherwise looks great!

jasulcaf:
Summary:
create_list_attribute() will output an attribute linked list that can be manipulated by the following functions:
- add_attribute_to_list: Add an attribute to list
- remove_attribute_from_list: Remove an attribute (if it exists) from list 
- list_contains_attribute: Returns Boolean (for ease of utility for users) if an attribute is found in a list or not
- delete_attribute_llist: Free entire list

Additional note about an attribute_list_t: There exists the usage of an initial dummy node in the implementation of an attribute_list_t. While this is not something that will have to be known by users (function calls to add, remove, delete an entire list, or check if list_contains_attribute is true, do _not_ have to know this information i.e. users can simply pass entire list as a function call argument), if future updates are done on the implementation of the above mentioned functions, this should be kept in mind.

MaxineK36:
Issue Score: ✔️++

Comments:
Great work! This is an excellent example of a well-described PR with helpful reviews and a great closing statement that summarizes the work completed.

PR #847, opened 2021-04-26
Changing Player Class Item Restrictions to be Managed via Linked Lists

### LINKED ISSUE #615: Change class-based item attributes to use linked-lists

ideclue:
(Closes #615) (Relates to part (1) of #851)

Previously, restricting a Player Class from using a certain item was managed through the conventional item attribute system. However, **player class restrictions** are relatively self contained and it is desirable for them to be maintained separately through a linked-list based framework. These changes allow for player class restrictions to be accessed and managed easily through this new framework.

Summary of Major Changes:

- Added a `class_restrictions` field to `item` that keeps track of and manages class item restrictions separately from other item attributes
- Added a `remove_item_restriction` method so item restrictions can be removed from a certain Player Class
- Changed the `remove_attribute_from_list` function in `item.c` to allow attributes to be passed in by name and consequently allow for easier removal of class item restrictions
- Updated all relevant documentation in the Player Class Module to reflect these changes

Notes:

-  This Pull Request depends on #624 being successfully merged, as the `attribute_list_t` struct methods defined and implemented there are used extensively throughout my code
- In the future, the framework for creating attributes will be modified so that attributes do not have to be created directly in the Player Class Module


src/playerclass/src/class_item_restriction.c
```diff
@@ -1,22 +1,57 @@
...
 #include <stdbool.h>
 #include <stdlib.h>
+#include <string.h>
 #include "game-state/item.h"
 #include "playerclass/class.h"
 #include "playerclass/class_structs.h"
 
 /* see class_item_restriction.h */
-int set_item_restriction(item_t* item, class_t* class) {
+int add_item_restriction(item_t* item, class_t* class) {
```
jackcasey067, line 10: This is a stylistic nitpick, really just a suggestion (I am still going to approve, in other words):

You have a few functions with this sort of structure:
 - You check some conditions, often compound conditions, in an if statement.
 - The real logic of the function is in the body of the if statement.
 - You deal with the possibility of invalid inputs at the bottom.

I think this creates unnecessary nesting, and also puts unnecessary distance between the error condition (item or class being NULL) and the response to that error.  Instead, you might consider the pattern of guards:
- Check for invalid input at the top of the function, and deal with it right there, returning an error code.
- Continue with the rest of the function after that, without any additional nesting in an else block.

This has the following benefits:
- Less nesting.  It doesn't really matter here, but could help later with functions that have trickier logic.
- Error-catching conditions are close to error handling code.
- You can easily break up compound conditions into separate guards and handle them differently.  I actually recommend doing this if you do decide to adopt guards for this code.

src/playerclass/src/class_item_restriction.c
```diff
@@ -1,22 +1,57 @@
...
+        }
+
+        /* Attribute creation will be refactored in game-state/item to allow for streamlined attribute creation */
+        attribute_t* restriction = malloc(sizeof(attribute_t));
+        restriction->attribute_tag = BOOLE;
+        restriction->attribute_value.bool_val = true;
+        restriction->attribute_key = strndup(class->name, 100);
+
+        if (restriction == NULL) {
+            fprintf(stderr, "Failed to allocate memory");
```
jackcasey067, line 23: Two things here: 
1. I think you should return immediately if this error occurs.  We probably do not want to add NULL to `item->class_restrictions`.
2. I think it is recommended to add the location to an error print statement, to make it bit easier to pin down the bug if it does occur.  So like:
```suggestion
            fprintf(stderr, "Failed to allocate memory in class_item_restriction");
```
This was at least the custom adopted by whoever wrote the code I worked on (class.c).  It seems like a good idea.  You should do this everytime you `fprintf` an error message.

LapsededPacifist (approved):
Looks like the only change to the action_management code was changing `remove_attribute_from_list` to take a char* instead of an attribute* as the thing to remove. The test dependencies were changed and still passed, so the action_management modifications look good to me.

jackcasey067 (changes requested):
Overall, looks pretty good.  I am going to request that you change the error printing thing and the error return thing, but if you think I am wrong about those then I am open to changing my mind on them.

src/playerclass/src/class_item_restriction.c
ideclue: This is a good suggestion. I decided to go along with what the last person who worked on the code had did, but you're right that checking in different places would improve readability error-checking.

src/playerclass/src/class_item_restriction.c
ideclue: Thank you for pointing this out!

jackcasey067 (approved):
Looks quite good now.  I would still wait for #846 to be pulled before this gets pulled.

firatciftci:
Let's wait until PR #846 is reviewed and merged to `dev`, and then we can merge all of the changes from their end to here and review this PR as a whole.

firatciftci (approved):
Everything looks good to me, amazing work!

ideclue:
Closing Statement:

Upon review and implementation of peer feedback (specifically, improving error handling), my code has been successfully merged with `dev`. Class item restrictions are now managed through a  linked-list framework can be manipulated in the `playerclass` module.

firatciftci:
Issue Score: ✔️++

Comments:
Great work!

PR #897, opened 2021-05-02
Implementing mode functionality to support NPC dialogue

### LINKED ISSUE #864: Begin implementing "mode" functionality in the CLI in order to integrate NPC dialogue
Currently, the CLI can only interpret user input as game commands (push, pick up, look, etc). To implement dialogue, we would like to isolate user input from this function by creating a mode struct or field accessible by the game context. In "conversation mode" the CLI does not treat the user input as a command, but only checks for dialogue options.
Assignees: Peter, James
team: npc2

pkasner:
(Closes #864 )

Previously, the CLI only interpreted user input as commands for actions. To implement NPC dialogue, we needed a way to have user input not treated as a command. By having a game mode option, chiventure can switch between operating through the normal CLI and operating through separate functions, such as a run_conversation function, while still obtaining user input in the same way.

Changes:
- Added mode.h and mode.c file with struct definitions and init/new/free functions
- Added a mode field to the game struct
- When not in normal mode, the print_cli function calls the mode's run_mode function instead of do_cmd.
- The load_yaml_game function also loads the game mode in normal mode.


jeyeong (approved):

src/ui/src/print_functions.c
```diff
@@ -170,14 +170,23 @@ void print_cli(chiventure_ctx_t *ctx, window_t *win)
         return;
     }
```
pkasner, line 172: This section of the code is the only part that technically interacts with the CLI, although it is perhaps better described as a diversion than an interaction.

jeyeong:
This looks good to me.

src/game-state/src/mode.c
```diff
@@ -0,0 +1,64 @@
...
+
+#include "game-state/mode.h"
+
+
+/* see mode.h */
+int game_mode_init(game_mode_t *mode, mode_type_t curr_mode, 
+                   mode_operation run_mode, char* mode_ctx)
+{
+    assert(mode != NULL);
+    strcpy(mode->mode_ctx, mode_ctx);
```
lilyehsani, line 12: Do not use `strcpy` here (and in fact try to avoid using it in general). Instead, use `strncpy(mode->mode_ctx, mode_ctx, MAX_IS_LEN);` as it is a safer option which protects against unterminated strings.

lilyehsani (changes requested):
Added a couple relatively minor comments. Overall great job!

pkasner:
I've fixed the formatting mistakes for * and switched strcpy to strncpy

pkasner:
Just realized from last class that some of chiventure's tests don't actually load a WDL file, so the load_yaml_game function won't be called in this case. To ensure that a mode is always initialized, I can add a condition to do so in the context's init function.

lilyehsani (approved):
LGTM! Approved.

lilyehsani:
LGTM! Approved.

pkasner:
The mode functionality is currently working, but there are currently no mode operations or actions implemented for switching mode. The next steps for implementing npc dialogue are writing a mode operation function for conversation mode, which will call run_conversation_step() in the dialogue module, and implementing a "TALK" action to initiate conversations with NPCs.

lilyehsani:
Issue Score: ✔️++

Comments:
Great work overall! The summary, updates, and closing of this PR are very thorough.

Issue #875, opened 2021-04-27
Update relevant code to player integration according to updated player needs


afranzese99:
After the completion of issue #835 (meeting with the relevant RPG teams (playerclass, skill trees, and possibly others) and updating the player integration plan documentation) we will need to implement the desired features into the chiventure code. The player struct will likely be updated, in addition to functions that may become outdated with new additions to the player module/wishlist (e.g. `player_new()`in player.h (.c)).

WPDolan:
@afranzese99 and I have completed documentation of the player struct. When implementing the initialization, new, and free functions, we were unsure if `player_free` should also free the associated `class_t` field. After discussion with the playerclass team, we determined the `class_t` struct will not be freed given the playerclass team's potential desire to use classes for multiclassing, descriptions, or other purposes. Given future implementation of the `class_t` struct by the playerclass team, we may need to free this field in the future

afranzese99:
@WPDolan and I have updated functions that take in the `player_t struct` as a parameter (`player_new()`, `player_init()`, `player_free()`). However, the health field in the `player_t` struct is outdated and will be turned into a stat held within the `stats_hash_t *player_stats`, this will break many functions that currently use the health field in the player struct (e.g. `change_health()` in the player module, `chiventures_ctx_init()` in the common module, etc.). We will also need to update wherever `player_new()` is called now that this function takes in different parameters than before.

WPDolan:
After discussing with Professor Borja, I have updated the `player_new` function to only set a the player's ID, instead of every field present in the player struct. I have also updated all function calls to `player_new` from setting both the health and id to only setting the player's health. Currently, the implementation for the health functions remain in the code, but they will most likely be removed once health is fully integrated as a stat in the player's `player_stats` field.

WPDolan:
The pull request has been approved, as the associated branch with this issue, `action-management/player-integration`, has been successfully merged. 

We have removed the old `player_init` function which set the fields of the player struct. In its place, we have created multiple functions to set individual fields in the player struct such as player_set_race and player_set_class. More functions will be added in future issues (#944 and #952) to set other fields in the player struct. This allows developers to ignore fields in the player struct that may not pertain to their game.

We also removed the health field from the old player struct. We intend that both a player's maximum and current health will be stored as respective stats in the player's player_stats hash. This will allow a player's health to be affected by functions currently implemented in the stats module, and provide additional options for its exclusion should a game developer be uninterested in health as a mechanic for their game.

[...]

PR #898, opened 2021-05-02
Implemented player integration plan into game-state module

WPDolan:
We have implemented action-management's player integration [plan](https://github.com/uchicago-cs/chiventure/wiki/Action-Management-Player-Integration-Plan) into the existing game-state module as outlined in #875. 

This required us to update the established player struct to accommodate requests from other RPG teams. We have also updated the associated `player_new` and `player_free` functions to accommodate the new fields of the player struct. Updating the function call to `player_new` necessitated updating previous calls to `player_new` in various testing modules as well as the common module.

Two new functions and associated tests have been created, `player_set_race` and `player_set_class`. Getter/setter calls for other fields in the player struct will need to be implemented in a future issue depending on whether the `class_t` field is going to be dynamically or statically updated over a game.


src/game-state/src/player.c
```diff
@@ -2,38 +2,65 @@
...
+    plyr->player_id = malloc(MAX_ID_LEN * sizeof(char));
 
-    int check = player_init(plyr, player_id, health);
+    assert(player_id != NULL);
 
-    if (plyr == NULL || plyr->player_id == NULL)
-    {
-        return NULL;
-    }
+    strncpy(plyr->player_id, player_id, strlen(player_id));
```
LapsededPacifist, line 52: strlen(player_id) needs to be replaced with MAX_ID_LEN (or strnlen). Currently this doesn't properly terminate the string and could theoretically not properly initialize it.

src/game-state/src/player.c
```diff
@@ -61,26 +103,6 @@ int delete_all_players(player_hash_t* players)
...
 }
 
-/* See player.h */
-int get_health(player_t* plyr)
-{
-    return plyr->health;
-}
-
-/* See player.h */
-int change_health(player_t* plyr, int change, int max)
```
LapsededPacifist, line 71: Asked in the slack about this, currently it looks like you are losing the functionality to cap HP at the max hp. Not sure if this is intended and max hp is being implemented a different way.

src/game-state/src/player.c
```diff
@@ -44,7 +71,22 @@ int player_free(player_t* plyr)
...
+    if (plyr->player_class != NULL)
+    {
+        class_free(plyr->player_class);
+    }
+
+    if (plyr->inventory != NULL)
+    {
+        delete_all_items(&plyr->inventory);
+    }
+
```
LapsededPacifist, line 89: player_stats, player_combat_skills, player_non_combat_skills, and player_effects all currently remain unfreed. I didn't see reference to this in the plan, so I am not certain, but it doesn't seem like all of these are global. For example, will player_stats outlive the player?

Also not really important, but delete_all_items should just return on NULL, so you don't need to check.

borjasotomayor (approved):
The changes to the `player_t` initialization functions (and their effects on the rest of the chiventure code) look good to me. Please note that this is the only aspect of this PR that I have reviewed (i.e., you should still wait on @MaxineK36's review before considering this PR to be approved)

LapsededPacifist (approved):
Updates look good to me. They met all my original comments and I think are solid.  Might be a bit confusing to people not expecting the health field to be gone, but that is just a quick documentation fix instead of a code thing.

include/game-state/player.h
```diff
@@ -30,30 +56,16 @@ typedef struct player {
...
  * Parameters:
- *  plyr: A player. Must point to already allocated memory.
- *  health: The starting health of the player
- *  player_id: the unique string ID of the player. currently
- *   this will always be "1" since there is only one player
- *
- * Returns:
- *  SUCCESS on success, FAILURE if an error occurs.
- */
-int player_init(player_t *plyr, char* player_id, int health);
```
MaxineK36, line 44: Why was this function deleted?

src/game-state/src/player.c
```diff
@@ -2,38 +2,64 @@
...
+/* See player.h */
+int player_set_class(player_t *plyr, class_t *player_class)
+{
+    if (plyr == NULL || player_class == NULL)
+    {
+        return FAILURE;
+    }
+
+    plyr->player_class = player_class;
```
MaxineK36, line 34: How did you land on shallow-copying here?

src/game-state/src/player.c
```diff
@@ -44,7 +70,19 @@ int player_free(player_t* plyr)
...
     free(plyr->player_id);
+
+    if (plyr->player_race != NULL)
+    {
+        free(plyr->player_race);
+    }
+
+    if (plyr->player_class != NULL)
+    {
+        class_free(plyr->player_class);
```
MaxineK36, line 81: Is the function adding a player class expected to create and allocate one for the player?

include/game-state/player.h
```diff
@@ -15,11 +15,37 @@
...
+
+    /* A string containing the player's race */
+    char *player_race;
+
+    /* The player's current class. class_t contains the base stats, and skills for that class at
+    the beginning of a game. These may change throughout the game, so their current states are stored 
+    in the health, player_stats, player_skills fields in this player struct */
     class_t *player_class;
+
+    /* All of the stats, with their values, the player has */
```
MaxineK36, line 36: Can you note that this includes health?

MaxineK36 (commented):
This is looking good overall! I left a few comments where I'd like some clarification, and then I'll take a look again

include/game-state/player.h
WPDolan: The functionality for `player_init` will be/is inherited by smaller functions like `player_set_class,` `player_set_race`, and future, similar functions for other fields for the player struct. This will give developers the option to create players without having to think about fields that may not be relevant to their current game.

src/game-state/src/player.c
WPDolan: Currently, there exists no functionality for creating deep copies of any of the structs associated with classes, skilltrees, stats, etc, so shallow-copying was the only way to have this information be accessible to the struct. In the future we may talk with relevant teams and ask them to create copy functions for their modules so that we could obtain the necessary information without risk of freeing something that was not intended to be freed. This was also the reason that we have not yet created functions to set stats or skills, due to the fact that we would most likely pull them from the given class at player creation.

src/game-state/src/player.c
WPDolan: We expect that all non-string information will already be created and allocated somewhere before any calls to `player_new`.

include/game-state/player.h
MaxineK36: Great, that makes sense; please just also note that in your closing statement

MaxineK36 (approved):
Looks awesome! There are two changes in here that may seem counterintuitive—removing the health field and removing player_init—even though they are logical design choices, to please make sure to mention and explain them explicitly in your closing statement

WPDolan:
The `action-management/player-integration` branch has been successfully merged into the `dev` branch. 

Alongside the updates to `player_new` and `player_free` mentioned in the opening summary, we have also removed the old `player_init` function which set the fields of the player struct. In its place, we have created multiple functions to set individual fields in the player struct such as `player_set_race` and `player_set_class`. More functions will be added in future issues to set other fields in the player struct. This allows developers to ignore fields in the player struct that may not pertain to their game.

Another change that needs mention is our removal of the health field from the old player struct. We intend that both a player's maximum and current health will be stored as respective stats in the player's `player_stats` hash. This will allow a player's health to be affected by functions currently implemented in the stats module, and provide additional options for its exclusion should a gave developer be uninterested in health as a mechanic for their game.

Future issues #944 and #952 will document and implement more functionality into the player interface based off of the changes made to the player struct by this merge.

MaxineK36:
Issue Score: ✔️++

Comments:
Excellent work! Great job linking to the future issues in your closing statement.

Issue #884, opened 2021-04-28
Clean up npc code according to issue #828


amymorrill:
When reviewing existing NPC code, we found several small inconsistencies (such as repeated parameters in the NPC struct). We will fix those inconsistencies to clean up the code.

northofsophia:
A new npc/clean-structs branch has been checked out where the code in the npc struct has been cleaned up, as well as creating a new npc_battle struct.

amymorrill:
We are continuing this issue into Sprint 3, and we have decided to temporarily close the associated pull request (see #899 for more details). This is because our teammates also submitted a pull request (#904) editing the main npc struct, which conflicts with our edits to the npc struct. After our teammates' pull request goes through, we will revisit this issue and see if the npc code / npc struct need to be cleaned up further, or if we can close this issue based on their edits.

amymorrill:
I have just finished cleaning up the code for the npc.c struct in the npc/clean_structs branch. The most important task related to this was deleting the duplicated class fields. While I was doing this, I also noticed that the npc_new, npc_init, and npc_free functions do not match the updated (or even the previous, non-updated) npc struct, so I also changed those to match the struct. 

The next step will be to deal with merge conflicts between the dev branch and npc/clean_structs branch and to re-open pull request #899.

amymorrill:
I have resolved merge conflicts between the dev branch and npc/clean_structs branches and have re-opened pull request #899 which is pending review.

Notably, as I mentioned in my last comment, I thought that npc_new and npc_init did not match the npc struct because they were missing dialogue and inventory fields. However, I have since realize that an npc's inventory and dialogue are meant to initialize to NULL, and should be filled in later. Thus, I again got rid of the dialogue and inventory fields in npc_new and npc_init in this last round of edits.

lilyehsani:
Issue Score: ✔️++

Comments:
Great work overall! The updates of this issue are very thorough.

Issue #883, opened 2021-04-28
Integrate the npc and battle modules according to the design done in issue #828


amymorrill:
In the last sprint, we made plans to integrate the npc and battle modules. Now, we want to implement those plans, in consultation with the battle team. Specifically, we will move information in the npc-enemy struct in the battle module to be contained in the npc struct, allowing the battle team to get rid of that struct. We also will discuss with the battle team any further ways that npcs and battles can be integrated.

amymorrill:
After talking with @Elsa-Athiley and @lczielinski from the Battle team, we have decided that our plan to integrate NPCs and Battle structs will work (moving info from the npc-enemy struct into an npc-battle struct, which will be a part of the main npc struct). The next step is to actually create the npc-battle struct, modify the existing npc struct, and then modify and functions explicitly reference the fields of npc struct (like npc_new, etc.).

amymorrill:
We have decided to carry this issue over into Sprint 3. The main reason for this is that we are temporarily closing the associated pull request (see #899 for more details). 

We also have realized that we need to edit our code to avoid breaking tests that other teams have written which include npc_new and npc_init. To address this, we will make npc_battle default to NULL in npc_new and npc_init and will then create additional functions to add an npc_battle struct to an npc. We will implement these functionalities before reopening pull request #899.

amymorrill:
We are still waiting on two pull requests to go through (#904, #941) before editing our code and re-opening our previous pull request (#899). In the meantime, I have updated the npc design and planning wiki page (https://github.com/uchicago-cs/chiventure/wiki/NPC-Design-and-Planning#integration-of-NPC-module) to reflect our planned changes to the npc struct (and those implemented in #904) as well as more details about our plans for npc/battle integration.

amymorrill:
I have just updated the npc/clean_structs branch with a few changes from the original plan. 

First, we decided to have a separate npc_battle module incorporating the definition for the npc_battle struct, new/init/free functions for that struct, and any other battle related functions we create in the future. Accordingly, I have create an npc_battle.h and an npc_battle.c file. 

[...]

PR #899, opened 2021-05-02
Cleaning up npc.h and implementing integration plan for npc and battle modules.

amymorrill:
We have cleaned up the npc struct in npc.h as outline in #884. We also have implemented our plan to integrate the npc and battle modules as outlined in #883. 

First, we created a new struct called npc_battle_t which holds the information about npcs that is only pertinent to battles. We also created npc_battle_init, npc_battle_new, and npc_battle_free functions.

Second, we updated the existing npc struct, getting rid of repetitions and adding an npc_battle field. We also updated npc_init, npc_new, npc_free, get_npc_health, and change_npc_health to be compatible with the updated npc struct. 

Notably, there are still some warnings and errors that come up in the npc module. However, these warnings/errors either arise from problems with the battle/battle_structs file (item_t is defined here and in game-state/item.h) or problems with the get/set functions for the npc's inventory, which were not edited by us (this may be because of the problem with the battle file because the inventory deals with items). These problems should be addressed by the battle team or by us in a future issue.


amymorrill:
After discussing with the rest of our team, we are going to temporarily close this pull request. 

There are three reasons for this. The first is that there is a pull request from other members of our team (#904) that also edits the main npc struct, so it will make integration easier if we add our edits to the dev branch after their code has been added. The second is that our branch currently does not build due to a dependency on a file from the battle team (battle_structs.h) that contains an error (redefining item.h which is already defined in game-state). The third is that because we changed the npc_new and npc_init functions, our code will break tests that other teams have written which include those functions. To solve this problem, we will make npc_battle default to NULL in npc_new and npc_init and will then create additional functions to add an npc_battle struct to an npc.  

Thus, we will wait until our team members' pull request goes through (#904), the battle_structs.h file is modified, and we fix our npc_new/npc_init problem before reopening this pull request.

amymorrill:
Both of the pull requests that we were waiting on were merged with the dev branch, and we've edited our code to resolve all merge conflicts with dev, so this pull request is now being re-opened. Below are a list of all of the changes made in this pull request (both before the initial pull request was made and before re-opening it):
- We created a new struct called npc_battle_t which holds information about npcs that is only pertinent to battles.
- We create a new module called npc_battle which holds the npc_battle_t struct as well as the new/init/free functions for that struct.
- We edited the npc struct to no longer include health, which now is held in the npc_battle struct. We added will_fight to the npc struct, which is a boolean telling whether or not an npc engages in battle (for our reasoning for this, see #883).
- We updated the npc new/init/free functions to reflect the new npc struct. 
      - Note that at first, I added an inventory and dialogue field to the npc_new and npc_init functions, not realizing that those 
         are supposed to be initialized to NULL. Since then, however, I have edited those functions to no longer take those fields. 
      - Also note that I initially made npc_new and npc_init take an npc_battle_t struct as a field, but I later changed it so that 
         npc_battle initializes to NULL, and I removed the npc_battle field from npc_new and npc_init. 
- We updated the test_npc.c and test_rooms_npc.c files to get rid of conflicts with the updated npc_new function. I added @torresszy to review these changes as someone from the npc team.
- We updated the test_quests_state.c file to get rid of conflicts with the updated npc_new function. I added @dillonoloughlin to review these changes as someone from the quest team.

src/npc/src/npc.c
```diff
@@ -58,8 +61,9 @@ int npc_free(npc_t *npc)
     free(npc->npc_id);
     free(npc->short_desc);
     free(npc->long_desc);
-    free(npc->class);
+    convo_free(npc->dialogue);
```
torresszy, line 64: There seems to be a `convo_free` function call in line 55 already

torresszy (changes requested):
Looking good overall! There is a minor issue I commented on, but correct me if I am not understanding the code correctly.

src/npc/src/npc.c
amymorrill: Oops I didn't notice that. I've removed the extra line.

torresszy (approved):
Looks good to me now!

src/npc/src/npc.c
```diff
@@ -58,8 +61,8 @@ int npc_free(npc_t *npc)
     free(npc->npc_id);
     free(npc->short_desc);
     free(npc->long_desc);
-    free(npc->class);
     delete_all_items(&npc->inventory);
+    class_free(npc->class);
```
lilyehsani, line 65: I believe we should we be freeing the npc_battle_t in the npc struct here as well.

tests/npc/test_npc.c
```diff
@@ -153,47 +155,6 @@ Test(npc, get_ldesc_npc)
                  "get_ldesc_npc() didn't get npc's long_desc");
 }
 
-/* Checks that get_health() returns the health of the npc */
-Test(npc, get_npc_health)
```
lilyehsani, line 157: Why were these two tests deleted? Are they going to be rewritten in a future test_npc_battle module?

lilyehsani (changes requested):
Great job! Added a couple comments.

tests/npc/test_npc.c
amymorrill: I deleted these tests because health is now in the npc_battle module, and I haven't yet written a function to add an npc_battle field to the npc struct (which I will be making in my new issue), so it's impossible to assign npcs a health level right now. I will add back these tests when I write other npc_battle function tests!

src/npc/src/npc.c
amymorrill: Oops, that is true. I will add that now.

src/npc/CMakeLists.txt
```diff
@@ -2,11 +2,13 @@ add_library(npc
             src/npc.c
             src/rooms-npc.c
             src/dialogue.c
-            src/npc_move.c)
+            src/npc_move.c
+	    src/npc_battle.c)
 
 target_include_directories(npc PRIVATE src/)
+target_link_libraries(npc battle)
```
lilyehsani, line 9: Is there a reason you added this line? I don't believe it is necessary.

lilyehsani (changes requested):
One more thing

src/npc/CMakeLists.txt
amymorrill: I was getting an error in the compiler where it said that move_free (a function in the battle_moves module, which was included in npc_battle.h) was an undefined reference. I went to office hours and Borja recommended that I should add this line, which fixed the problem.

lilyehsani: Awesome! Thanks.

lilyehsani (approved):
LGTM! Approved.

amymorrill:
Closing statement: This pull request has been merged into dev. In addition to the changes described in previous comments, we also modified CMakeLists.txt to fix unresolved reference errors.

lilyehsani:
Issue Score: ✔️++

Comments:
Great work overall! The summary and updates of this PR are very thorough.

Issue #876, opened 2021-04-27
Make functions that return default items, moves, and stats for battles


lczielinski:
Summary: The current battles code lacks an easy way for game authors to implement general battle events in their games. We want to add functions that return default items, moves, combatants, and settings to easily allow implementation of battles with random or default settings.

lczielinski:
After meeting with each other, we decided to start with creating functions for making a default move, item, and stat. These functions would be semi-random, e.g. numerical fields would be picked randomly, and names would be picked from an array of names (which for now would only include 1 name). Other structs: players, enemies, combatants, and the battle struct would be more complicated to make defaults for, but would depend on the defaults for moves, items, and stats, so after finishing these functions we will discuss further how to implement the more complex default objects.

lczielinski:
We created a branch, battle/default_objects, in order to work on a new battle src file, battle_default_objects.c. In this file, we have functions to return a semi-randomly-generated item, stat, and move. We will now meet with our senior developer to work on next steps, and we will consider integrating these default objects in a battle example as well as creating more complex default objects.

sonnykaiyuenlee:
Met with Firat and discussed our current code. As of right now, we return randomized fields for moves, stats, and items structs. We decided we should make these concrete (eg we have an item sword with x damage and y defense), rather than randomized numbers for the fields. To finish this issue we are going to make well defined moves, stats, and items. We need to create a header file and a test file to ensure that our functions are returning items, moves, stats with expected fields.

lczielinski:
We created a pull request with our completed file, battle_default_objects.c, as well as an include file and test file (with two tests that pass). The next steps would probably be to use these functions in the example battle code, and create new functions for more complex default items (combatant_t, battle_t, etc).

lczielinski:
The pull request has been approved and merged into dev, containing new files battle_default_objects.c, battle_default_objects.h, and test_battle_default_objects.c. To summarize, the src file contains functions that return a randomized/default item, stat, and move to be used in an example/general battle.

firatciftci:
Issue Score: ✔️+

Comments:
Great work on this! I only have one thing to comment on, which is regarding the summary:
- The summary for this issue is too terse. A good rule of thumb is that you should be writing at least one full paragraph in your summary (see https://uchicago-cs.github.io/cmsc22000/project/policies.html#tasks-issues for examples).

PR #900, opened 2021-05-02
Battles/default objects

lczielinski:
We created a new battle src file, battle_default_objects.c, which has three functions for returning a default item, stat, and move. The default stat is more random, by choosing random numbers within a reasonable range for each field. There are 10 default items and moves each, which currently have very simple methods of determining fields (basically, the 10 items/moves increase in overall power). This could be tweaked in the future, but we think it is sufficient for the current purpose of being used in example battles. We also included a header file and a test file (with 2 new tests that pass). #876


Elsa-Athiley (approved):
Looks Awesome!

firatciftci (approved):
Great job! Everything looks good to go.

lczielinski:
The pull request containing our code, new files being battle_default_objects.c, battle_default_objects.h, and test_battle_default_objects.c, has been merged. To summarize, the src file contains functions that return a randomized/default item, stat, and move to be used in an example battle.

firatciftci:
Issue Score: ✔️++

Comments:
Great job!

PR #902, opened 2021-05-02
Making split screen show images rather than a rectangle.

eao10000:
@grkapoor17 and I worked inside of the ui/examples directory working on changing split screen example to utilize images rather than rectangles as a graphic. Utilizing function such as LoadImage DrawTecture, ect we were able to successfully draw an image. For the image to work it has to be png and images under 500 x 500 pixels. Then adding 2 more pictures in the sandbox/images directory that works allowing for there to be 3 photos total.


grkapoor17 (approved):

MaxineK36 (changes requested):
One small thing: can you note in a comment in your code what the requirements are for the images (size + extension)? Otherwise, looks awesome!

MaxineK36 (approved):
Looks great!

grkapoor17 (commented):
Looks good!

eao10000:
The change from rectangles to images in split-screen has been implemented, we had to utilize more function from raylib to complete this. While doing this we found that the images has to be 500 X 500 pixels or a png to work, this information has been included in the code.

Issue #893, opened 2021-04-30
Update npc testing frameworks after the integration between rooms-npc and npc module


torresszy:
The rooms-npc module has integrated into the npc module by adding another field into the npc struct, and relevant functions are updated as well. We need to now update the tests for npc and rooms-npc to reflect the latest changes in the source codes

torresszy:
I have created a new test file, `test_npc_move.c` and refactor the `test_rooms_npc.c` so the tests related to `npc_mov` struct will now go to `test_npc_move.c` and tests related to `npcs_in_room` struct will stay in `test_rooms_npc.c`. I will also be writing new tests to any new functions we added from last sprint.

torresszy:
I have also written a new test for `delete_npc_from_room` in `test_rooms_npc.c`. All the tests have passed now with the new updates, and I have mde a pull request to this issue

lilyehsani:
Issue Score: ✔️++

Comments:
Great work overall! The updates of this issue are very clean and to the point.

PR #904, opened 2021-05-03
Integrating npc_mov struct into npc struct and splitting rooms-npc module

torresszy:
After investigating the `npc` and `rooms-npc` module, we have decided to integrate the `noc_mov` struct from `rooms-npc.h`  into the `npc` struct in `npc.h` as an additional field. However, since `rooms-npc.h` uses definitions and methods from `npc.h` already, including `rooms-npc.h` in `npc.h` will create circular dependencies between the two module. In order to solve this problem, I have splitted `rooms-npc` into itself and a new module called  'npc_move`. The `rooms-npc` module now only has the struct `npcs_in_room' struct and its related methods, and the `npc_move` module has the struct `npc_mov` and other related structs and methods. These changes allow more flexibility in developing npc movements in the future without worrying about circular dependencies between modules.

Summary of major changes:
- Added the `npc_mov` field to `npc` struct
- Removed `npc_id` field in the `npc_mov` struct since this struct will be part of a npc struct now
- Updated the `add`, `init`, `free`, and other relevant functions for these two data types to incorporate the changes
- Added a new function, `delete_npc_from_room` in `rooms-npc` since the original module only had a `add_npc_to_room` function
- Moved the structure `npc_mov` and its related structs and methods from `rooms-npc` to `npc_move`, leaving only the struct `npcs_in_room` and its methods in `rooms-npc`
- Updated `test_npc.c`, `test_rooms_npc.c`, and `test_quests_state.c` to incorporate the new field. However, this update is temporary, and more thorough changes in the existing tests and well as adding new tests should be done in future sprints.
- Updated the Makefile to include the new module `npc_move`
- Updated the `npc_example.c` and `quest-example.c` so the example can be built successfully.

Note:
- I have already made a new issue (#893) for further updates on the npc tests frameworks in the future. However, other parts of the chiventure, for example, quests, also need updates since they are using the npc structs as well.


xlewellen (approved):
Splitting up `rooms-npc` seems like the most straightforward way to avoid circular dependencies between `npc` and `room-npc`. It is important to keep in mind that this will have implications for other npc reliant teams, such as the tean-npc-2 and the quest team (in instances of npc reliant quests). With these changes implemented, it might be a good idea to add a method to `rooms-npc` that calls both `add_npc_to_room` and `delete_npc_from_room` that will effectively move an existing npc from one room to another, allowing for easier use of the movement capabilities of npcs.

mbutera17:
The existing NPC Movement Design wiki was outdated, and I updated it to reflect progress made during this sprint detailed in Torres' comment above. These updates can be seen [here](https://github.com/uchicago-cs/chiventure/wiki/RPG-NPC:-Independent-Feature:-NPC-Movement-Design-Document#npc-movement-through-the-world). Further, on the [NPC Design and Planning Wiki](https://github.com/uchicago-cs/chiventure/wiki/NPC-Design-and-Planning), there was no existing mention or link to the NPC Movement Design wiki (which I added).

src/npc/src/npc_move.c
```diff
@@ -0,0 +1,218 @@
+#include "npc/npc_move.h"
+
+// STRUCT FUNCTIONS -----------------------------------------------------------
+/* See npc_move.h */
+int npc_mov_init(npc_mov_t *npc_mov, npc_mov_enum_t mov_type,
+                 room_t *room)
+{
+    char* room_id  = room->room_id;
```
lilyehsani, line 8: Move * (a few times in this file)

src/npc/src/npc_move.c
```diff
@@ -0,0 +1,218 @@
...
+/* Helper function for move_npc_def function */
+int room_id_cmp(room_list_t *room1, room_list_t *room2)
+{
+    return (strcmp(room1->room->room_id, room2->room->room_id));
+}
+
+/* See npc_move.h */
+int move_npc_definite(npc_mov_t *npc_mov)
+{
+
```
lilyehsani, line 193: Delete empty line

lilyehsani (approved):
LGTM! Approved.

torresszy:
This pull request has been merged successfully, and this is the closing statement. After this merge, the npc struct now contains a new field called npc_mov, which contains the type of movements and a list of rooms for a npc to visit. There is a following pull request #958, which updates the testing frameworks for the changes in the npc struct.

lilyehsani:
Issue Score: ✔️+

Comments:
Great work overall! 

- For a task of this size, we would've expected to see more status updates and/or status updates that were more detailed, particularly regarding why the PR was closed and reopened. 
- When you create a PR, you are requesting a review of the code in the branch you want to merge. As such, you should not push any commits to the branch right after creating the PR; you should only do so in response to a code review. Make sure that, when you create a PR, the code is ready for review and passes all the tests.

Issue #881, opened 2021-04-28
Add functionality that allows players to use items (consumables) during battles


Elsa-Athiley:
Currently players are only able to use 3 commands in battle,  MOVE USE move_name ON enemy_name , MOVE LIST and HELP . Players are  unable to use items in battle, specifically consumables that would increase health or attack and defense strength. I need to add functionality for USE item_name.

Elsa-Athiley:
Added a print_items function that is called when the command HELP is used. It will print the available battle items and list name, attack, defense, and hp values.  There is currently  find_item and use_item functions that I will use to create the functionality for the USE <item_name> command. These are being tested in the battle _command_example.c file.

Elsa-Athiley:
Added functionality for USE <item_id> had to change it to take in the id instead of name because of the way the commands are parsed. Also printed stat changes due to using the item. The only thing left is to clean up the spacing of print statements in the terminal and clean up the code so there is less repetition.

Elsa-Athiley:
Submitted a pull request with the changes (#905 ). I was able to implement a command that allows users to use items during battle in the `battle_command_example.c` file and also updated other commands to include items. The HELP command will now also list the available items and their associated information. Another team member is now working on implementing temporary effects with the items consumed during battle.

firatciftci:
Issue Score: ✔️+

Comments:
Great work on this! I only have one thing to comment on, which is regarding the summary:
- The summary for this issue is too terse. A good rule of thumb is that you should be writing at least one full paragraph in your summary (see https://uchicago-cs.github.io/cmsc22000/project/policies.html#tasks-issues for examples).

PR #905, opened 2021-05-03
Added use <item> functionality to example and updated readme to reflect updates

Elsa-Athiley:
I implemented the USE <item_id> command in the `battle_command_example.c` so now we can test the functions related to item use in battle. Players are able to use consumables in a battle. The changes in hp, attack, and defense of the player is printed to the screen to show the results of the move. I also added a print_items function, so the available items can be displayed alongside the list of moves when the HELP command is used.  The associated README was updated to reflect these changes
(#881)


src/battle/examples/battle_command_example.c
```diff
@@ -106,6 +107,26 @@ int print_moves(chiventure_ctx_battle_t *ctx)
...
+/* Prints out the avaliable items for the player
+ * Parameter:
+ *  ctx: the main structure of the game
+ * Returns:
+ *  Always SUCCESS
+ */ 
+int print_items(chiventure_ctx_battle_t *ctx)
+{
+    item_t *temp;
+    printf("\n AVAILABLE ITEMS LIST:\n");
```
lczielinski, line 119: Tiny thing, but the space after \n messes up the formatting in the console

src/battle/examples/battle_command_example.c
```diff
@@ -167,8 +188,40 @@ int read_move(char **args, chiventure_ctx_battle_t *ctx)
...
         printf("MOVE USE <move_name> ON <enemy_name>\n");
         printf("MOVE INFO <move_name>\n\n");
+        printf("USE <item_id>\n");
+        printf("Here is the list of available items!\n");
+        print_items(ctx);
+        printf("\n");
         return SUCCESS;
     }
+    //   // handles the command USE <item>
+     else if (strncmp(args[0], "USE", MAX_COMMAND_LENGTH) == 0)
```
lczielinski, line 198: Just an idea for the future, but maybe it would be a good idea to add functionality to allow the user to input "USE <move_name>" instead of move id

src/battle/examples/battle_command_example.c
```diff
@@ -167,8 +188,40 @@ int read_move(char **args, chiventure_ctx_battle_t *ctx)
...
+            return FAILURE;
+        }
+
+        res = use_item(ctx->game->battle->player, atoi(args[1]));
+        if (res == FAILURE) {
+            return FAILURE;
+        } else 
+        {
+        printf("New Hp is %d\n", ctx->game->battle->player->stats->hp);
+        printf("New Strength is %d\n", ctx->game->battle->player->stats->strength);
```
lczielinski, line 220: I think "Strength" should be "Attack" instead

lczielinski: Never mind, I see how the field in stat is called "strength"--maybe we should change the item_t struct's field to be called "strength" instead of "attack"?

src/battle/examples/battle_command_example.c
```diff
@@ -256,11 +309,39 @@ int main()
...
+    p_item2->quantity = 10;
+    p_item2->durability = 3;
+    p_item2->name= "Elixr of life";
+    p_item2->description = "Adds 80 to your hp!" ;
+    p_item2->battle = true;
+    p_item2->attack = 20;
+    p_item2->defense = 60;
+    p_item2->hp = 40;
+    p_item2->next = NULL;
+    p_item2->prev = NULL;
```
lczielinski, line 337: Just in this example, I think p_item2->prev should be p_item1. Also for the future, I'm looking through the item functions in battle_logic and I don't think it updates the linked list of items when one item gets used up. That could be something to fix in the future

src/battle/examples/battle_command_example.c
Elsa-Athiley: Yea im currently changing some naming stuff in another issue so I can go ahead and do this as part of that issue .

src/battle/examples/battle_command_example.c
Elsa-Athiley: I assume you mean "USE <item_name" instead of item_id and I also had the same thought initially and I realized the find_item() function that I use to implement the command takes in the item_id as one of the parameters so It was simpler to go with this. But for sure this could be apart of another issue to update the the item related functions.

Elsa-Athiley:
Okay so I fixed all the things Laura pointed out. The only things not changed were things that would take a bit more time or can be done in another issue

firatciftci (approved):
Everything looks good to me!

lczielinski (approved):
Yes, looks good now and I like the ITEM LIST functionality!

firatciftci:
Issue Score: ✔️+

Comments:
Great work! I only have one thing to comment on:
- When a pull request is done, you should include a final comment summarizing what was merged to the `dev` branch.

Issue #392, opened 2019-11-22
Make all input text case insensitive


ecoble:
Currently input text is case sensitive, e.g. if there is a chair in a room, "CHAIR" will not be a valid object. Make all text case insensitive

ecoble:
Input is now case insensitive, as input is converted to all caps before being parsed. WDL files must specify objects/directions/commands in all caps. Caps were picked as this was the convention for actions, however it could be switched to all lowercase instead.

ecoble:
This is issue was closed with PR #385

borjasotomayor:
Reopening this issue and putting it on the backlog because it looks like making chiventure case-insensitive is going to require deeper modifications in the code.

For example, consider the following lines from the CLI extra command example: https://github.com/uchicago-cs/chiventure/blob/23ea6d186f938a99ed5bc9ee88396ec0e7c7a904/src/cli/examples/extra-command.c#L33-L35

If you change `ROCK` to `rock` then the following happens in the CLI:

```
  > look rock
   specified item not found
  > look ROCK
   specified item not found
```

This is because the item is added to the items hash table with key `rock`, but the CLI always converts the parameter to `look` to uppercase, so it always looks up the key `ROCK` (which is not found).

In this specific case, it may be enough to modify `item_new` to convert the item identifier to uppercase, but it would be good to actually survey the rest of the code to ensure that all identifiers are internally converted to either lowercase or uppercase (preferably lowercase, since that's what case-insensitive text is usually converted to)

samdonio:
Currently, the cli parser makes all inputs fully uppercase, but this can have problems as outlined above. A way to make the cli case insensitive is to have any identifier key be made fully lowercase before being added to the respective hashtable. It seems most of these changes would be made in the game-state module i.e. probably item.h, path.h, room.h, stats.h, game_action.h, game.h, as they take in something to be added to or created in the game, and at times the hashtable. The parser however can be changed very quickly to make every input lowercase.

mjgoodman321:
Modified the parser to automatically lowercase, and added a general helper function to game-state-common so it could be applied to all of game state easily. The helper function should be moved to a more appropriate spot, preferably one that allows it to be implemented across multiple modules as needed.

mjgoodman321:
Moved case_insensitize helper function to a new module in cli for utility functions. Implemented the function inside of other cli modules.

mjgoodman321:
[...]

PR #907, opened 2021-05-03
Cli/case insensitivity

mjgoodman321:
@samdonio and I implemented case insensitivity as outlined by Issue #392

Created a cli-utility module and included a case-insensitize function within. Implemented that function within the cli and game-state functions as needed in order to create case insensitivity, as well as modified existing code to support it.

The assumption was made that all instances of (char *tokens[TOKEN_LIST_SIZE]) were created through the parser, and thus didn't need to be case-insensitized


mjgoodman321:
Closed Pull Request due to complications within the code making it not ready. Will be making a new pull request later

PR #919, opened 2021-05-04
Memory management in game-state and action_management

### LINKED ISSUE #877: Memory management in game-state, action-management, and related tests
Currently a lot of tests in game-state and action-management are leaking memory. We need to determine if only the tests are leaking memory or if the functions themselves are responsible. Using valgrind on the test executables, it is possible to find uninitialized memory leakage which can lead to a debugging phase.

jasulcaf:
(Closes #877)
This is connected to @LapsededPacifist and my Issue (#877) which was focused on clearing up the memory leakage that was found in the tests for game-state and action_management. The purpose of clearing the tests was so we could potentially uncover bugs in free functions (or the lack of free functions) which could prove to become detrimental as Chiventure continues to grow.

Overview of implemented changes:
- All tests run without leaking memory (except for issues with chiventure_ctx_free and related objects that where outside our scope of game-state and action_management functions. More details can be found here: https://github.com/uchicago-cs/chiventure/issues/877#issuecomment-831684904)
- delete_all_items now _successfully_ deletes duplicate items without memory leakage
- Fixed memory leakage where strings where being malloc'ed instead of using strndup

Something to note was that before, if valgrind was simply ran as: `valgrind ./(Executable)` on the test executables, then it would show no memory leakage. However, when you ran `valgrind --tool=memcheck --trace-children=yes ./(Executable)` you would get a much more descriptive outcome that showed how much memory was actually being lost.


src/game-state/src/item.c
```diff
@@ -537,8 +537,18 @@ int delete_all_items(item_hash_t** items)
     item_t *current_item, *tmp;
     HASH_ITER(hh, *items, current_item, tmp)
     {
-      	remove_item_from_hash(items, current_item); /* deletes (items advances to next) */
```
WPDolan, line 540: What does the new iteration loop achieve that the old iterator failed to accomplish? Uthash documentation makes the old method seem like a correct approach.

WPDolan:
When running the valgrind command provided in the request statement, all of the tests in `tests/game-state` run without memory leakage. However, many of the tests in `tests/action_management` leak memory. I have brought this up in the team slack.

src/game-state/src/item.c
jasulcaf: As can be found in more detail in our Issue Page https://github.com/uchicago-cs/chiventure/issues/877#issuecomment-830861590, with the current implementation of how adding duplicate items to the hash currently works, this loop is necessary to delete all duplicated items. When removing a duplicated item currently, only the "head" item is being removed while the "tail" is being rehashed to the same place in the hash table. However, HASH_ITER would have already moved on to the next hash spot. The purpose of the loop is to thoroughly delete all re-addition (i.e. all duplications) to the same hash spot _before_ continuing the process of iterating through the rest of the hash.

WPDolan: That makes sense. Thanks for the clarification!

WPDolan (approved):
Changes successfully achieve their goal, valgrind is showing that the proper blocks have been freed in game-state and action_management tests. My original questions have been clarified, and I see no further issues. Approved.

jasulcaf:
As a summary of what we talked about in the team slack, the tests that had memory leakage in action_management tests (test_item_action, test_item_item_action, and test_path_actions) occurred due to these initializing a chiventure_ctx_t struct which contains a lookup_t** field called table. The free function, lookup_t_free does not function properly and correcting this would require talking with the CLI team and learning what is actually inside (and what should be freed). I have edited the initial summary accordingly to reflect this.

src/game-state/src/condition.c
```diff
@@ -203,6 +204,7 @@ condition_t *inventory_condition_new(player_t *player, item_t *expected_item)
     new_condition->expected_item = expected_item;
 
     condition_t *condition_wrapper = malloc(sizeof(condition_t));
```
MaxineK36, line 206: Instead of using `malloc` and then `memset` with 0, you should use `calloc`
e.g. `condition_t *condition_wrapper = calloc(1, sizeof(condition_t));`
See the [manual page](https://linux.die.net/man/3/calloc) for more details

src/game-state/src/stats.c
```diff
@@ -98,6 +98,7 @@ int stat_effect_init(stat_effect_t *effect, effects_global_t *global)
 stat_effect_t *stat_effect_new(effects_global_t *global)
 {
     stat_effect_t *effect = malloc(sizeof(stat_effect_t));
+    memset(effect,0, sizeof(stat_effect_t));
```
MaxineK36, line 101: Ditto here; use `calloc`

tests/action_management/test_get_actions.c
```diff
@@ -7,10 +7,32 @@
...
 #define NUM_ACTIONS (17)
 
+/* free_get_supported_actions: Free's list_action_type_t list that contains ALL
+ *                             supported actions
+ * 
+ * Inputs:
+ *      - head: a list containing all supported actions. MUST be built with 
+ *              get_supported_actions
+ */
+void free_get_supported_actions(list_action_type_t* head)
```
MaxineK36, line 17: Why is this only necessary for tests, and not included in a source file?

tests/action_management/test_get_actions.c
```diff
@@ -20,37 +42,43 @@ Test(get_actions, count)
...
 }
 
 
-action_type_t *search_supported_actions(char *query)
+action_type_t *search_supported_actions(list_action_type_t* head, char *query)
 {
-    list_action_type_t *head, *temp;
-    head = get_supported_actions();
+    list_action_type_t *temp;
+    //head = get_supported_actions();
```
MaxineK36, line 53: Please remove any commented-out code that is no longer used!

tests/action_management/test_item_actions.c
```diff
@@ -137,7 +137,11 @@ int execute_do_item_action(char *act_name, enum action_kind kind, char *allowed_
         rc = do_item_action(ctx_test, a, item, &string);
         break;
     }
-
+    
+    
+    game_free(ctx_test->game);
+    lookup_t_free(ctx_test->table);
+    free(ctx_test->table);
```
MaxineK36, line 144: do we know why this isn't included in `lookup_t_free`?

tests/game-state/test_player.c
```diff
@@ -227,28 +242,29 @@ Test(player, add_item_effect_to_player)
...
   "item for testing add_item_to_player");
+  new_item->stat_effects = NULL;
   effects_global_t *g1 = global_effect_new("effect 1");
   stat_effect_t *e1 = stat_effect_new(g1);
   stat_effect_t *e2 = stat_effect_new(g1);
   class_t *class = class_new("class", "short", "long",
                    NULL, NULL, NULL);
   player->player_class = class;
   add_stat_effect(&player->player_class->effects, e1);
+  stats_global_t *health = malloc(sizeof(stats_global_t));
```
MaxineK36, line 253: is there a `global_stats_new` function?

tests/game-state/test_player.c
```diff
@@ -258,7 +274,17 @@ Test(player, add_item_effect_to_player)
                      "add_item did not add stat_mod to effect");
   cr_assert_eq(player->player_class->stats->modifier, 1.125, 
                "add_item did not update modifier");
-
+  free_stats_table(class->stats);
+  delete_all_stat_effects(class->effects);
+  class_free(class);
```
MaxineK36, line 279: Fyi, looks like with the updated player functions, you may  not need to separately free the class anymore

MaxineK36 (changes requested):
Awesome work! Just a couple nit-picky things to fix

src/game-state/src/condition.c
jasulcaf: Incorporated change, didn't think about this, thanks!

tests/action_management/test_get_actions.c
jasulcaf: To be quite frank, I didn't know if we should have included this in a source file or not (didn't know if it merited this since I am not sure how often nor the context of how get_supported_actions is called in other places in Chiventure). I know that in src/get_actions.c lies `get_supported_actions()` function, so if you think that it would be a good idea to move it as a function that can be used by other teams in Chiventure, then I can definitely move free_get_supported_actions to be in the same source file as `get_supported_actions()`.

tests/action_management/test_item_actions.c
LapsededPacifist: No. From what I have seen, I would think that it should be in `lookup_t_free`, but `lookup_t_free` is also not freeing the table very well. I think the entire function needs to be written.
If I had to guess, I would say the probably could potentially be similar to our problem of only freeing the first item in each bucket of the hash, but the function currently doesn't seem to be freeing all of the first item. Just the name and the struct itself (I think it also contains an action).  
Basically, I don't know why much of what is happening in `lookup_t_free` is happening.

tests/game-state/test_player.c
jasulcaf: Yep, there was! Thank you for pointing that out, I have changed the test to use stats_global_new so this is now resolved.

tests/action_management/test_get_actions.c
jasulcaf: In the end, the functions was both renamed and moved to the source file. The new name is now `free_supported_actions ` and is found in src/get_actions.c

tests/game-state/test_player.c
LapsededPacifist: I removed the class_free call. The test will leak memory for a little bit, but the memory side of this test is going to be fixed when freeing effects no longer double frees everything.

include/action_management/actionmanagement.h
```diff
@@ -82,6 +83,15 @@ int action_type_init_room_dir(action_type_t *a, room_t *room, char *direction);
  */
 list_action_type_t *get_supported_actions();
 
+/* free_get_supported_actions: Free's list_action_type_t list that contains ALL
```
MaxineK36, line 86: nit: *frees

MaxineK36 (commented):
One typo, then it should be good to go!

MaxineK36 (approved):
Looks great!

jasulcaf:
As a conclusion to our Pull Request:
We dug through the test code to fix memory leakage related to action_management and game-state functions and in the process we added a new free function called `free_supported_actions` (header file in actionmanagement.h, implemneation found in src/get_actions.c)  to free all supported actions; and we also fixed the following mistake with delete_all_item: 
- delete_all_item was not freeing all resources due to a complication with HASH_ITER and the way that adding duplicated items in a hash table worked. This required incorporating a loop to remove all duplications before continuing the iteration in the hash table

We have also identified some issues that could be looked at in future issues:
1.  the current free function to free a chiventure struct is not implemented 
2. There exists complications with the lookup_t_free functions with the full table not being fully freed. 
3.  Stats and Stats global seem to have some dependencies/relationship that complicates how they are successfully freed.

For more detailed information on these issues, can find at: https://github.com/uchicago-cs/chiventure/issues/877#issuecomment-831684904

MaxineK36:
Issue Score: ✔️++

Comments:
Great work! Your conclusion is particularly useful, especially as it details potential future work.

PR #943, opened 2021-05-05
Adding type-specific attribute creation functions to streamline Player Class/Attribute integration

### LINKED ISSUE #851: Update Player Class/Attribute Integration
Currently, the only way Player Classes can interact with items is through class item restrictions. In order to add to this functionality and clean up existing class item restriction code, it will be helpful to improve the Player Class/attribute integration in order to eventually brainstorm and add more features. This issue will consist of two primary components:

1. Updating the current class item restriction code to reflect changes from Pull Request #846: The return and function parameters of many of the attribute linked list functions were altered after the review of this Pull Request 
2. Altering the method for creation of attributes of an item (by adding `attribute_new` and `attribute_init` functions) in order to make it easier to add class item restrictions inside `playerclass` code, and eventually, add more interactions between player classes and items

ideclue:
(Closes #851)

In the current class item restriction code, it is necessary to create an attribute directly in order to add it to the `class_restrictions` field of an item (a linked list of attributes). However, the code in `item.c` only contains type-specific `set_attr` functions that combine the creation of an attribute and its addition to the `attributes` hash table. Since class item restrictions are handled via a linked-list, not the `attributes` hash table, I divided up the two processes occurring in the `set_attr` functions in order to allow for creation of attributes in modules other than `game-state`.

Summary of Major Changes:

- Five type-specific attribute creation functions have been added to `item.c` 
- Relevant tests were added and documentation was updated
- Minor changes in `class_item_restriction.c` reflect the addition of these type-specific attribute creation functions

### DIFF
--- a/src/game-state/src/item.c
+++ b/src/game-state/src/item.c
@@ -213,14 +213,15 @@ attribute_t *get_attribute(item_t *item, char* attr_name)
 attribute_t* str_attr_new(char* attr_name, char* value)
 {
     attribute_t* new_attribute = malloc(sizeof(attribute_t));
-    new_attribute->attribute_tag = STRING;
-    new_attribute->attribute_value.str_val = value;
-    new_attribute->attribute_key = strndup(attr_name, 100);
 
     if (new_attribute == NULL) {
         fprintf(stderr, "Failed to allocate memory for new attribute");
         return NULL;
     }
+
+    new_attribute->attribute_tag = STRING;
+    new_attribute->attribute_value.str_val = value;
+    new_attribute->attribute_key = strndup(attr_name, MAX_ID_LEN);
 
     return new_attribute;
 }
@@ -230,14 +231,15 @@ attribute_t* str_attr_new(char* attr_name, char* value)
 attribute_t* int_attr_new(char* attr_name, int value)
 {
     attribute_t* new_attribute = malloc(sizeof(attribute_t));
-    new_attribute->attribute_tag = INTEGER;
-    new_attribute->attribute_value.int_val = value;
-    new_attribute->attribute_key = strndup(attr_name, 100);
 
     if (new_attribute == NULL) {
         fprintf(stderr, "Failed to allocate memory for new attribute");
         return NULL;
     }
+
+    new_attribute->attribute_tag = INTEGER;
+    new_attribute->attribute_value.int_val = value;
+    new_attribute->attribute_key = strndup(attr_name, MAX_ID_LEN);
 
     return new_attribute;
 }
@@ -247,14 +249,15 @@ attribute_t* int_attr_new(char* attr_name, int value)
 attribute_t* double_attr_new(char* attr_name, double value)
 {
     attribute_t* new_attribute = malloc(sizeof(attribute_t));
-    new_attribute->attribute_tag = DOUBLE;
-    new_attribute->attribute_value.double_val = value;
-    new_attribute->attribute_key = strndup(attr_name, 100);
 
     if (new_attribute == NULL) {
         fprintf(stderr, "Failed to allocate memory for new attribute");
         return NULL;
     }
+
+    new_attribute->attribute_tag = DOUBLE;
+    new_attribute->attribute_value.double_val = value;
+    new_attribute->attribute_key = strndup(attr_name, MAX_ID_LEN);
 
     return new_attribute;
 }
@@ -264,14 +267,15 @@ attribute_t* double_attr_new(char* attr_name, double value)
 attribute_t* char_attr_new(char* attr_name, char value)
 {
     attribute_t* new_attribute = malloc(sizeof(attribute_t));
-    new_attribute->attribute_tag = CHARACTER;
-    new_attribute->attribute_value.char_val = value;
-    new_attribute->attribute_key = strndup(attr_name, 100);
 
     if (new_attribute == NULL) {
         fprintf(stderr, "Failed to allocate memory for new attribute");
         return NULL;
     }
+
+    new_attribute->attribute_tag = CHARACTER;
+    new_attribute->attribute_value.char_val = value;
+    new_attribute->attribute_key = strndup(attr_name, MAX_ID_LEN);
 
     return new_attribute;
 }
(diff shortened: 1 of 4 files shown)


src/game-state/src/item.c
```diff
@@ -208,126 +208,173 @@ attribute_t *get_attribute(item_t *item, char* attr_name)
...
+{
+    attribute_t* new_attribute = malloc(sizeof(attribute_t));
+    new_attribute->attribute_tag = STRING;
+    new_attribute->attribute_value.str_val = value;
+    new_attribute->attribute_key = strndup(attr_name, 100);
+
+    if (new_attribute == NULL) {
+        fprintf(stderr, "Failed to allocate memory for new attribute");
+        return NULL;
+    }
```
jackcasey067, line 223: In each of these functions, you malloc a new attribute, then use it, and then only at the end do you check for if malloc fails.  `new_attribute->anything` will error if malloc really did fail (I think), so you should move the if statement up to before you access the fields of `new_attribute`.

src/game-state/src/item.c
```diff
@@ -208,126 +208,173 @@ attribute_t *get_attribute(item_t *item, char* attr_name)
...
+}
+
+
+/* see item.h */
+attribute_t* int_attr_new(char* attr_name, int value)
+{
+    attribute_t* new_attribute = malloc(sizeof(attribute_t));
+    new_attribute->attribute_tag = INTEGER;
+    new_attribute->attribute_value.int_val = value;
+    new_attribute->attribute_key = strndup(attr_name, 100);
```
jackcasey067, line 235: This is maybe debatable, but the `100` here and around your code feel like magic numbers; you should maybe use `MAX_ID_LEN` (which I found used in this file, even though I don't know where it is defined) instead; that way if it changes to allow a different size identifier, your code automatically updates.

tests/game-state/test_items.c
```diff
@@ -405,7 +467,7 @@ Test(attribute, set_str_attr)
     cr_assert_eq(rv, SUCCESS, "change_str_attr: did not successfully set attr");
     int num_in_hash = HASH_COUNT(test_item->attributes);
     cr_assert_gt(num_in_hash, 0, "change_str_attr: no elements added to hash");
-    attribute_t* test_attr = get_attribute(test_item, "Attribute_Test_Name");
+     attribute_t* test_attr = get_attribute(test_item, "Attribute_Test_Name");
```
jackcasey067, line 470: Looks like you accidentally added a random space here.

src/game-state/src/item.c
```diff
@@ -208,126 +208,173 @@ attribute_t *get_attribute(item_t *item, char* attr_name)
...
-    }
-    else if (res != NULL && res->attribute_tag != STRING)
-    {
+    } else if (res != NULL && res->attribute_tag != STRING) {
         return FAILURE; // skeleton for not overriding type
-    }
-    else
-    {
+    } else {
         res->attribute_value.str_val = value;
```
LapsededPacifist, line 310: I don't know what effect this would have if any, but my initial thought was this should be `strndup(value)`. It looks like they were not using it before, so it is probably fine. You would have to pay attention and free the value before you changed it and have a special case in attribute free. Is there a reason that the string isn't being saved or is it just not necessary as it will always be static?

jackcasey067 (changes requested):
Pretty good work overall; two things to change IMO (see my other comments for more details on both).

1. Make sure to check for NULL right after you malloc.
2. Avoid magic numbers.

src/game-state/src/item.c
```diff
@@ -208,126 +208,173 @@ attribute_t *get_attribute(item_t *item, char* attr_name)
...
+    new_attribute->attribute_value.bool_val = value;
+    new_attribute->attribute_key = strndup(attr_name, 100);
+
+    if (new_attribute == NULL) {
+        fprintf(stderr, "Failed to allocate memory for new attribute");
+        return NULL;
+    }
+
+    return new_attribute;
+}
```
LapsededPacifist, line 294: There is a lot of shared code in these functions. I am not sure if this is a change you want to make, but I would have considered a single attr_new function that also takes as input a tag (And the type of value would need to become a union).
The only functionality comment I have for these functions, I think the NULL check needs to be directly after the malloc. You won't be able to access the memory currently if it is NULL

LapsededPacifist (changes requested):
The only thing that I think needs to be changed is when the NULL check needs to happen. Otherwise, I think things look good.

src/game-state/src/item.c
LapsededPacifist: MAX_ID_LEN is defined as 20 in [game_state_common.h](https://github.com/uchicago-cs/chiventure/blob/dev/include/game-state/game_state_common.h)
I agree with this. This was a pre-existing problem in the code, but it would be nice to actually give it a name instead of just tossing around a random 100.

src/game-state/src/item.c
ideclue: I also noticed this, but I thought in light of the type-specific "set" functions, it would be best to continue the current pattern of implementation.

src/game-state/src/item.c
ideclue: I will definitely make this change. Thank you both for pointing this out!

src/game-state/src/item.c
ideclue: I think it's the latter (I played around with `attribute_free` and no notable changes occurred).

jackcasey067 (approved):
Good work!  Approving.

LapsededPacifist (approved):
Changes to the NULL check look good.

src/game-state/src/item.c
firatciftci: Although the redundancy is not ideal, I think it is fine in terms of its extent – also, as @ideclue mentioned, the `set` functions follow the same pattern, so I think having the file follow the same logical pattern is a better idea than to try and come up with a different function design. That being said, if anyone would like to refactor this file to be less redundant in the future, that could be a good idea.

firatciftci (approved):
The comments above by the other reviewers are really on point! With all the changes made, I think this is good to go!

ideclue:
Closing Statement:

After implementing peer feedback, all code has been successfully merged with `dev`. Type-specific attribute creation functions can now be used across modules and will be helpful in managing linked-lists of attributes (and other relevant data structures) for the purpose of developing more complex relationships between items and player classes.

firatciftci:
Issue Score: ✔️++

Comments:
Great job!

Issue #918, opened 2021-05-04
Set up CMake file to compile battle_flow_example


adinilfeld:
Currently the CMake file in `battle/examples`only compiles one example file, `battle_command_example`. I'll be modifying the CMake to compile the `battle_flow_example` file as well, which has already been written to run a sample battle using hard-coded moves. 

See also issues #839 and #892, where this problem was discovered.

adinilfeld:
After getting advice from @firatciftci, I figured out how to update the CMake file to compile `battle_flow_example` (commit 2f5946acb901fca855708a3256b37e616a382d9d). That caused a bunch of errors + warnings, which I fixed in commit 63f9a7ef92d1fd2043491d6bec1dcb8189e9e65f. However, there's some kind of error where `battle_flow_example` ends after one turn, and with the enemy HP seeming to have overflowed (it's at -2147483578), so I'll look into how to fix that.

adinilfeld:
After a lot of fiddling, I finally figured out what was wrong with `battle_flow_example`. Firstly, some player/enemy stats (i.e. level, defense, strength) weren't being set, so at some point during the damage calculation there was division by 0, which caused the weird overflow error. I then also had to adjust `battle_flow.c` to use a default move for the enemy -- previously the enemy wasn't given any moves, so the example battle would end in failure after the player's turn.

Now I just have to update the README to account for the fact that `battle_flow_example` is functional.

adinilfeld:
The issue is finished: `battle_flow_example` now compiles and runs successfully, and the README in `battle/examples` has been updated with the appropriate instructions. This task will be closed once the pull request is approved.

firatciftci:
Issue Score: ✔️+

Comments:
Good job! I only have one thing to comment on:
- When an issue is done, you should include a final comment summarizing what was done, and why the issue is being closed.

PR #956, opened 2021-05-06
Fixed battle_flow_example and updated README

adinilfeld:
Previously, the CMake file in `battle/examples` only compiled one example file, `battle_command_example`. It now also compiles `battle_flow_example` (which had to be slightly modified to run without error). I also updated the README in `battle/examples` to match the fact that `battle_flow_example` is functional.

Closes issue #918.


src/battle/src/battle_flow.c
```diff
@@ -132,7 +132,8 @@ int battle_flow(chiventure_ctx_battle_t *ctx, move_t *move, char* target)
     }
     else
     {
-        return FAILURE;
+        dmg = damage(b->player, move, b->enemy);
```
lczielinski, line 135: Could you clarify this? This is the else case for "enemy_move != NULL", so if the enemy has no moves, then the player just keeps attacking the enemy?

lczielinski (approved):

lczielinski:
Looks good overall, I'm glad this example works now! I just have one question about a change you made

src/battle/src/battle_flow.c
adinilfeld: The idea behind this if-else is that if the enemy is given an `enemy_move` from the `give_move` function (line 127), then the enemy attacks the player using that move (lines 131-132). Otherwise, the enemy attacks the player with the same move the player used to keep the demo moving (lines 135-136).

Elsa-Athiley (commented):
looks good to me as well!

Elsa-Athiley (approved):

firatciftci (approved):
Looks good to me!

adinilfeld:
To summarize this merge: the CMake file now compiles `battle_flow_example` (which had already been written). I also fixed some bugs in `battle_flow_example` that had caused it to crash. The README in `battle/examples` has also been updated with instructions to run the example.

firatciftci:
Issue Score: ✔️++

Comments:
Great work!

PR #964, opened 2021-05-08
Cli/quit working solution

### LINKED ISSUE #273: Support for Quit Command

samdonio:
Previously chiventure could only be closed using the keystroke CTRL+D, but now the user can close it by typing "QUIT" into the cli followed by any keypress. This was done by adding an out parameter to many of the ui functions that interacted with the cli, and this out parameter would end the ui loop if the user inputted "QUIT".

Closes #273


include/ui/window.h
```diff
@@ -7,10 +7,10 @@ typedef struct chiventure_ctx chiventure_ctx_t;
 
 typedef struct window {
     WINDOW *w;
-    void (*print)(chiventure_ctx_t *ctx, struct window *win);
+    void (*print)(chiventure_ctx_t *ctx, struct window *win, int *quitval);
 } window_t;
 
-typedef void (win_print)(chiventure_ctx_t *ctx, window_t *win);
+typedef void (win_print)(chiventure_ctx_t *ctx, window_t *win, int *quitval);
```
ecoble, line 13: Instead of calling this quitval for all win_print functions, it would be better to name it something more generic like ret or retval. That way in the future it can be used for its own return values if need be, without it being specific to quitting the game

src/ui/src/ui.c
```diff
@@ -27,6 +27,10 @@ void start_ui(chiventure_ctx_t *ctx, const char *banner)
     signal(SIGINT, SIG_IGN);
 
     int ch;
+    int *quitval; 
+    quitval = malloc(sizeof(int));
```
ecoble, line 31: I don't think the value needs to be malloced, have you tried just lines 30 and 32?

src/ui/src/ui.c
```diff
@@ -27,6 +27,10 @@ void start_ui(chiventure_ctx_t *ctx, const char *banner)
     signal(SIGINT, SIG_IGN);
 
     int ch;
+    int *quitval;
```
ecoble, line 30: To clarify, naming this value quitval still makes sense.

src/ui/src/ui.c
samdonio: I have tried that, but after removing the lines where I malloc and also free the int pointer, typing QUIT just results in a seg fault.

include/ui/print_functions.h
```diff
@@ -34,32 +34,32 @@ void print_banner(window_t *win, const char *banner);
  * Parameters:
  *   - ctx : chiventure context struct
  *   - win : window where to print the info
+ *   - quitval: not used in this context
  *
- * No value is returned
  */
-void print_info(chiventure_ctx_t *ctx, window_t *win);
+void print_info(chiventure_ctx_t *ctx, window_t *win, int *quitval);
```
ecoble, line 40: All of the parameter names in this file should also be named retval, sorry for being unclear the first time.

ecoble (changes requested):

src/ui/src/print_functions.c
```diff
@@ -210,11 +215,10 @@ void print_cli(chiventure_ctx_t *ctx, window_t *win)
 }
 
 /* see print_functions.h */
-void print_map(chiventure_ctx_t *ctx, window_t *win)
+void print_map(chiventure_ctx_t *ctx, window_t *win, int *quitval)
```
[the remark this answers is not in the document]
ecoble: Great question! All of the printing functions are part of a "win_print" function type defined by the UI team, and they all have to have the same input parameters and output type. So unfortunately in this case, if we want to change the parameters of one win_print function we have to change the type of all of them.

include/ui/print_functions.h
mberamendiconde: Just changed the parameter names from quitval to retval!

include/ui/print_functions.h
```diff
@@ -34,32 +34,32 @@ void print_banner(window_t *win, const char *banner);
  * Parameters:
  *   - ctx : chiventure context struct
  *   - win : window where to print the info
+ *   - quitval: not used in this context
```
ecoble, line 37: The header comments should also be updated to reflect the quitval to retval change.

include/ui/print_functions.h
```diff
@@ -34,32 +34,32 @@ void print_banner(window_t *win, const char *banner);
...
  *
  * Parameters:
  *   - ctx : chiventure context struct
  *   - win : window where to print the info
+ *   - quitval: not used in this function
  *
- * No value is returned
  */
-void print_map(chiventure_ctx_t *ctx, window_t *win);
+void print_map(chiventure_ctx_t *ctx, window_t *win, int* retval);
```
ecoble, line 62: Small nit, everywhere else we are doing int *retval, it would be good to be consistent.

ecoble (changes requested):

include/ui/print_functions.h
mberamendiconde: Fixed this issue!

include/ui/print_functions.h
mberamendiconde: Fixed this!

mberamendiconde:
Revised @ecoble comments! Fixed the header comments to reflect the change from quitval to retval

include/ui/print_functions.h
ecoble: It looks like maybe this change didn't get pushed through, I still see it as "int* retval" not "int *retval"

src/ui/src/print_functions.c
```diff
@@ -121,7 +121,7 @@ void print_banner(window_t *win, const char *banner)
 }
 
 /* see print_functions.h */
-void print_info(chiventure_ctx_t *ctx, window_t *win)
+void print_info(chiventure_ctx_t *ctx, window_t *win, int* quitval)
```
ecoble, line 124: Same small nit here, about changing the placement of the * to be "int *quitval" for consistency.

ecoble (changes requested):

src/ui/src/print_functions.c
mberamendiconde: Apologies! I misunderstood your earlier comments, the asterix placement update should be pushed now, and consistent

include/ui/window.h
```diff
@@ -70,10 +70,10 @@ int window_free(window_t *win);
  * Parameters:
  *   - ctx : chiventure context struct
  *   - win : window which we want to print
+ *   - quitval: an out parameter to make the game close on "QUIT"
  *
- * does not return anything
  */
-void window_print(chiventure_ctx_t *ctx, window_t *win);
+void window_print(chiventure_ctx_t *ctx, window_t *win, int *quitval);
```
ecoble, line 76: Sorry for not catching this on the last review, but the parameters names for functions should match across files, meaning this should also be updated to retval. Basically everything should be updated to retval, except for when it is used as the name of a variable in ui.c

src/ui/src/print_functions.c
```diff
@@ -121,7 +121,7 @@ void print_banner(window_t *win, const char *banner)
 }
 
 /* see print_functions.h */
-void print_info(chiventure_ctx_t *ctx, window_t *win)
+void print_info(chiventure_ctx_t *ctx, window_t *win, int *quitval)
```
ecoble, line 124: Update quitval to retval here, and for the rest of the win_print functions in this file.

src/ui/src/window.c
```diff
@@ -52,7 +52,7 @@ int window_free(window_t *win)
 }
 
 /* see window.h */
-void window_print(chiventure_ctx_t *ctx, window_t *win)
+void window_print(chiventure_ctx_t *ctx, window_t *win, int *quitval)
```
ecoble, line 55: It should be updated here as well.

ecoble (changes requested):

mberamendiconde:
Should be fixed now!

ecoble (approved):
LGTM!

mjgoodman321 (approved):
Looks like you guys got it. Go for it!

mberamendiconde:
Closing statement: Success! This issue was resolved, now the user can input "QUIT" into the command line and the game will quit and the terminal will close equivalently to the "Ctrl+D" input. This implementation mainly affected UI code.

ecoble:
Issue Score: ✔️+

Comments:
Great work fixing this, I know the UI code can be super tricky to work with! Be sure all issues and PRs are properly labelled and have the right milestones.

PR #965, opened 2021-05-08
Create and implemente free_condition

### LINKED ISSUE #932: (small issue) Add free_condition function to condition.c
There currently is no function to free conditions as mentioned here: https://github.com/uchicago-cs/chiventure/issues/877#issuecomment-831014707. As such, this Issue is being created to develop a solution to properly free resources associated with a condition

jasulcaf:
(Closes #932)  

This is connected to my Issue #932 which set out to implement a free_condition function since in Issue #877 it was discovered that there was no designated function for this (and as a result, manual freeing was needed every time in all tests). 

Overview of Implemented Changes:
- Added and implemented free_condition
- Fixed tests in test_condition.c to use free_condition instead of the manual freeing
- Fixed a spot in code where `malloc` was being called and then `memset` with 0 instead of just calling `calloc` (an example of this is something that @MaxineK36 pointed out in another Pull Request: https://github.com/uchicago-cs/chiventure/pull/919#discussion_r626233932)


src/game-state/src/condition.c
```diff
@@ -124,8 +138,7 @@ condition_t *attribute_condition_new(item_t *item_to_modify, char *attribute_nam
     new_condition->attribute_to_check = attribute;
     new_condition->expected_value = new_value;
 
-    condition_t *condition_wrapper = malloc(sizeof(condition_t));
-    memset(condition_wrapper, 0, sizeof(condition_t));
+    condition_t *condition_wrapper = calloc(1, sizeof(condition_t));
```
LapsededPacifist, line 141: I was curious why you calloced condition_wrapper, but not new_condition. I am not sure you have to calloc new_condition, but Iam not sure. Was curious if there was a reason.

src/game-state/src/condition.c
```diff
@@ -88,23 +88,37 @@ int valid_condition(game_t *game, condition_t *condition)
...
+/* See condition.h */
+int free_condition(condition_t* condition)
+{
+    assert(condition != NULL);
+
+    /* Free the condition regardless of the type */
+    switch (condition->condition_tag)
+    {
+    case (ATTRIBUTE):
+        free(condition->condition.attribute_type);
```
LapsededPacifist, line 100: Does there need to be a null check here? Can this ever be NULL?

LapsededPacifist (approved):
I think it looks good. I had 2 comments, but I am not sure they require changes.

src/game-state/src/condition.c
jasulcaf: Technically it shouldn't be NULL but this was just something I added as a just-in-case precaution. When writing free functions in previous CS classes as well as just other code I've seen (such as the Libgeometry HW's in this class), this is usually included as a just-in-case scenario

src/game-state/src/condition.c
jasulcaf: For the calloc, I just knew that it is a better option than having to malloc and then memset to 0 again. This was just something I spotted in the code that I figured I'd "fix" along the way since Maxine had mentioned this in a prior instance to me (link to this can be found in the summary of this Pull Request).

src/game-state/src/condition.c
LapsededPacifist: I had been asking about `condition->condiotion.attribute_type`. You don't check if either of the types are NULL and I was curious if there was ever an instance where that could happen. I don't think there is, so it should be fine.

jasulcaf: Oh! I see what you mean now. Yes, there should not be a situation where this happens since both `inventory_condition_new` and `attribute_condition_new` will always create a `condition->condition.inventory_type` or `condition->condition.attribute_type` respectively.

jasulcaf:
In summary:
This Pull Request allows for the function `free_condition` to exist within Chiventure and so there are no longer manual frees for a condition floating around repeatedly in different code. The actual implementation detail is to use a switch statement so that if in the future more types of conditions are added, then this should be a simple change to the current function without the caller having to know the specifics (currently only two types of conditions: Inventory and Attribute)

MaxineK36:
Issue Score: ✔️++

Comments:
Great work! Your PR summary is particularly well-written; the bullets make it very clear and easy to read.

PR #983, opened 2021-05-11
created an example that spliced together split screen and map examples to form a visual with a split screen with a small map in the top left corner

zoetorok:
Created a new visual example in the UI examples folder. In this example, we parsed the split-screen and map examples together, then modified the code which created the map in order to supersede it on top of the split-screen and appear in the top left corner of the top of the split-screen. We also had to modify the code so that the windows only initialized once and rework the map sizing and location based on the window sizing input into the split-screen initialization.


grkapoor17 (approved):

src/ui/examples/split_screen_map_example.c
```diff
@@ -0,0 +1,272 @@
...
+            for(int i = 0; i < length; i++) {
+                letterCount--;
+                name[letterCount] = '\0';
+            }
+
+            if (letterCount < 0)
+                letterCount = 0;
+        }
+
+        //Draw Image
```
MaxineK36, line 167: Is this drawing the map or the whole image?

MaxineK36 (changes requested):
Looks good overall! A couple of small code changes, and it looks like you have a few empty files that should be deleted (and a 2021-pr-tutorial submodule that shouldn't be there)

Also -- is the white dot meant to represent the player? When I ran it, it was frequently outside of the map diagram

MaxineK36 (approved):
Looks great!

MaxineK36:
Issue Score: ✔️+

Comments:
Great work overall! One area of improvement for next time:
- The reviewer for this PR approved the PR without comment. Please make sure that you always do a full review of the code, even if the senior developer has already provided their review. In the rare cases where there truly isn't anything to improve in the PR, make sure to leave a comment confirming that you reviewed the code and don't have any changes to suggest. If the senior dev has left their review, it can also be useful to indicate whether you agree with it or not.

Issue #982, opened 2021-05-11
Refactor level-oriented generation code with adding a function for filtering speclist


wmingyan:
Currently, the level-oriented generation algorithm is a part of the multi_room_level_generate function (#882 ), which means it cannot be used for other generation functions like recursive_generate. We decide to put this functionality (filtering the given speclist with a given difficulty level) into a separate function so that we can take a filtered speclist as a parameter for other autogeneration functions, and this might also be useful for later features.

wmingyan:
The relevant [pull request](https://github.com/uchicago-cs/chiventure/pull/987) has been approved, and the refactored code, including a new function filter_speclist and tests for it, has been merged into the dev branch. This marks the end of this issue.

MaxineK36:
Issue Score: ✔️++

Comments:
Great work! Your summary is particularly detailed and helpful.

PR #987, opened 2021-05-12
Refactor level-oriented generation code

wmingyan:
This Pull Request is for refactoring the level-oriented generation code, which puts the functionality of filtering a speclist with a difficulty level into a separate function so that the filtered speclist can be used for other level-oriented autogeneration functions in addition to multi_room_level_generate. 
This involves updating the autogenerate module. Tests are added to test_autogenerate.c. Updates in the autogenerate module include adding filter_speclist_with_difficulty function and modifying multi_room_level_generate function. 
This PR is related to issue #982.

### DIFF
--- a/include/openworld/autogenerate.h
+++ b/include/openworld/autogenerate.h
@@ -195,6 +195,7 @@ int roomspec_is_given_difficulty(room_level_t **room_levels,
 
 /* filter_speclist_with_difficulty
  * Creates a speclist by filtering the given speclist with a difficulty level
+ * so that the returned speclist only contains roomspecs of one level
  *
  * Parameters:
  * - speclist: pointer to the speclist we want to filter
--- a/include/openworld/autogenerate.h
+++ b/include/openworld/autogenerate.h
@@ -193,6 +193,22 @@ int roomspec_is_given_difficulty(room_level_t **room_levels,
                                   int difficulty_level);
 
 
+/* filter_speclist_with_difficulty
+ * Creates a speclist by filtering the given speclist with a difficulty level
+ *
+ * Parameters:
+ * - speclist: pointer to the speclist we want to filter
+ * - room_levels: pointer to the hash structure for room levels
+ * - difficulty_level: the difficulty level
+ * 
+ * Returns:
+ * - pointer to the filtered speclist, NULL if no spec matches the level
+ */
+speclist_t* filter_speclist_with_difficulty(speclist_t *speclist, 
+                                            room_level_t **room_levels, 
+                                            int difficulty_level);
+
+
 /* multi_room_level_generate
  * Level-oriented version of multi_room_generate.
  *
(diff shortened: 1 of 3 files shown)


tests/openworld/test_autogenerate.c
```diff
@@ -1062,6 +1062,78 @@ Test(room_level, lvl0_to_lvl1_roomlevels)
...
 
+/* Checks that filter_speclist_with_difficulty returns NULL
+ * if no roomspec in the speclist is of the given difficulty level */
+Test(speclist, filter_speclist_NULL)
+{
+    roomspec_t *rspec1 = make_default_room("school", NULL, NULL);
+    roomspec_t *rspec2 = make_default_room("farmhouse", NULL, NULL);
+
+    speclist_t *list1 = speclist_new(rspec1);
+    speclist_t *list2 = speclist_new(rspec2);
```
Justin08784, line 1073: make_default_room creates a hash table of 5 roomspecs–– rspec1 and rspec2 are not singular room specs. This might be a problem when you call speclist_new on them because some hash table slots (including the rspec pointer) may be NULL. 

Instead, I think you should HASH_FIND_STR to obtain a specific rspec from each table and then use roomspec_to_room to convert it into a room.

tests/openworld/test_autogenerate.c
```diff
@@ -1062,6 +1062,78 @@ Test(room_level, lvl0_to_lvl1_roomlevels)
...
+}
+
+
+/* Checks that filter_speclist_with_difficulty successfully filters speclist */
+Test(speclist, filter_speclist)
+{
+    roomspec_t *rspec1 = make_default_room("school", NULL, NULL);
+    roomspec_t *rspec2 = make_default_room("farmhouse", NULL, NULL);
+
+    speclist_t *list1 = speclist_new(rspec1);
```
Justin08784, line 1107: Same as above.

Justin08784 (changes requested):

Justin08784 (approved):
That works too, GJ!

include/openworld/autogenerate.h
```diff
@@ -193,6 +193,22 @@ int roomspec_is_given_difficulty(room_level_t **room_levels,
                                   int difficulty_level);
 
 
+/* filter_speclist_with_difficulty
+ * Creates a speclist by filtering the given speclist with a difficulty level
```
MaxineK36, line 197: Does this return only items of matching difficulty level? or does it return anything that's, for example, that level or lower? 
Either way, this detail should be specified in the header

MaxineK36 (changes requested):
You should also add a test that ensures that `filter_speclist_with_difficulty` runs correctly when the resulting list should be non-empty but not the whole list (i.e. input a list with rooms of multiple difficulty levels including the target, and make sure you get the right output)

include/openworld/autogenerate.h
wmingyan: The returned speclist only contains roomspecs of the one level. I've added it to the header and hope that is clear.

MaxineK36 (approved):
Looks great!

wmingyan:
Closing statement:
Comments by reviewers have been addressed and branch `openworld/refactor-level-gen` is now merged into `dev.`

MaxineK36:
Issue Score: ✔️++

Comments:
Great work! Your summary is very detailed and helpful; for the future, you could also consider using bullet points to improve readability.

Issue #925, opened 2021-05-04
Implement autogeneration for rooms within a radius for openworld module


Justin08784:
Currently, our autogeneration functions (room_generate and multi_room_generate) are able to load only rooms that are directly adjacent (one path away from) the current room. It would offer more flexibility to game authors if they could specify all rooms that are within an arbitrary radius/number of paths (or as time and memory allows) from the room to be generated. This is similar to how we can adjust render distance in minecraft (essentially 'radius') to determine how many chunks we load.

We anticipate the use of a queue to keep track of visited rooms as the implementation will likely involve a breadth-first search, and we would have to enable room_generation around rooms beside the current_room.

wmingyan:
We three had a meeting, discussed the effects we want to achieve with this implementation, and tentatively decided to use recursive algorithm without queue for implementing this functionality.

Justin08784:
Added tests for recursive_generate for cases when radius is 0 (do nothing), 1, and 3.

Justin08784:
Recursive_generate tests were crashing. Bug was discovered at ~random_item_lookup()` in autogenerate.c. It was incorrectly passing in `tmp->item_id` not `current->item_id`.

Justin08784:
Summarizing last two commits:

- Added `char **directions` and `int num_dirs` parameters to recursive_generate so that users can specify exactly which paths are filled with rooms at each level of recursion. 

Note: each directions in the `directions` array can only be part of the default six: NORTH, EAST, UP, SOUTH WEST, DOWN. Since the directions parameter is not guaranteed to be ordered like so, recursive_generate internally constructs an `all_directions` array and uses strcmp to find the correct `directions` index `all_directions` mapping. This eats up a lot of stack memory.

It might be worth considering defining a new struct to map each direction to its reverse and structuring the directions parameter as an array of these structs. This would also enable support for any custom directions.

- Added `char *dir_to_parent` parameter to recursive_generate which removes a lot of redundant computations. Previously, recursive_generate would backtrack and make recursive calls on already generated rooms. Benchmarking using criterion tests reveals a 36% decrease in time, or 50% faster computation. Performance gains were not as significant as expected.

Justin08784:
Added test for dir_to_parent parameter. Checks that recursive_generate–– when given dir_to_parent "SOUTH"––generates no rooms in the "SOUTH" direction of the original room.

Justin08784:
The pull request for 

[...]

PR #997, opened 2021-05-13
Openworld: Implement autogeneration for rooms within a radius (aka recursive generation)

Justin08784:
Introduces the recursive_generate function which generates arbitrarily many connected (in paths) layers of rooms around a given room in a branchwise fashion. The number of layers is specified in number of paths away from the original room, denoted as the radius.

Relevant tests have been added to test_autogenerate.c.
This PR is related to issue #925.

Some terminology
_Pivot:_ the room around which more rooms will be generated

_Branchwise:_
- for each pivot, we fill as many paths/branches around it with new rooms
- branches are disjoint from each other; we can cross from one branch to another only by travelling through the pivot


ahmed-sumaiya (approved):

src/openworld/src/autogenerate.c
```diff
@@ -181,7 +181,7 @@ int random_item_lookup(item_hash_t **dst, item_hash_t *src, int num_iters)
 
     HASH_ITER(hh, src, current, tmp) {
         if (i == num_iters) {
-            copy_item_to_hash(dst, src, tmp->item_id);
+            copy_item_to_hash(dst, src, current->item_id);
```
MaxineK36, line 184: was this a bug from earlier?

src/openworld/src/autogenerate.c
```diff
@@ -262,3 +262,83 @@ int multi_room_level_generate(game_t *game, gencontext_t *context,
     return result;
 }
 
+
+/* See autogenerate.h */
+int recursive_generate(game_t *game, room_t *curr_room, speclist_t *speclist, 
+                       int radius, char **directions, int num_of_dir, char *dir_to_parent) 
+{
+    // base case
```
MaxineK36, line 270: Style: Comment like these (that precede a block of code and exist on their own line) should be block comments (`/* comment */`)

src/openworld/src/autogenerate.c
```diff
@@ -262,3 +262,83 @@ int multi_room_level_generate(game_t *game, gencontext_t *context,
...
+        // opposite direction
+        int forwards = dir_index[i];
+        int backwards = (forwards + 3) % 6;
+
+        if (path_exists_in_dir(curr_room, all_directions[forwards])) {
+            next_room = find_room_from_dir(curr_room, all_directions[forwards]);
+        
+        // create room in path if it doesn't exist yet
+        } else {
+            roomspec_t *rspec = random_room_lookup(speclist);
```
MaxineK36, line 324: Could any of these be refactored into a helper? This function is currently quite long

src/openworld/src/autogenerate.c
```diff
@@ -262,3 +262,83 @@ int multi_room_level_generate(game_t *game, gencontext_t *context,
...
+            path_t* path_to_next_room = path_new(next_room, all_directions[forwards]);
+            assert(SUCCESS == add_path_to_room(curr_room, path_to_next_room));
+
+            // Path for the opposite direction
+            path_t* path_to_curr_room = path_new(curr_room, all_directions[backwards]);
+            assert(SUCCESS == add_path_to_room(next_room, path_to_curr_room));
+        }
+
+        // recursive case, decrement radius by 1
+        assert(SUCCESS == recursive_generate(game, next_room, speclist,
```
MaxineK36, line 338: Could this be `int rc = recursive_generate(game, next_room, speclist, `...?

src/openworld/src/autogenerate.c
```diff
@@ -262,3 +262,83 @@ int multi_room_level_generate(game_t *game, gencontext_t *context,
...
+            path_t* path_to_curr_room = path_new(curr_room, all_directions[backwards]);
+            assert(SUCCESS == add_path_to_room(next_room, path_to_curr_room));
+        }
+
+        // recursive case, decrement radius by 1
+        assert(SUCCESS == recursive_generate(game, next_room, speclist, 
+                                             radius - 1, directions, num_of_dir,
+                                             all_directions[backwards]));
+    }
+    return SUCCESS;
```
MaxineK36, line 342: does this function ever return FAILURE? In keeping consistent with the rest of chiventure, you may want to change line 338 to `rc` (see above comment) and then have `return rc` here

MaxineK36 (changes requested):
Looks good overall, a couple of style things + other small fixes

src/openworld/src/autogenerate.c
Justin08784: Yep! It was from last year's team.

src/openworld/src/autogenerate.c
Justin08784: Lines 314-334 checks if a path from the current_room is empty and generates a single room at the end of the path if it is empty.

Now this is essentially copy-pasted code from room_generate. The only reason we're redefining this code instead of simply calling upon room_generate is because room_generate can only generate around the 'curr_room' that's defined in the game struct. 

We figured that we could modify room_generate so that it generates around any given room, which could be passed in as a parameter, but we weren't sure if this was worth the effort. Advice?

src/openworld/src/autogenerate.c
Justin08784: Currently, there is no fail condition.

src/openworld/src/autogenerate.c
MaxineK36: Is there a reason this one has "UP" and "DOWN" as directions and `room_generate` doesn't? 

And I think that that should make sense in the long run, but if it's a change that will take longer you can create up a backlog issue for it (you don't need to add any assignees yet) and it can be done later on. For now, just note in a comment above that section that the code is similar to the `room_generate` code

src/openworld/src/autogenerate.c
MaxineK36: @Justin08784 Why not return `FAILURE` if `recursive_generate` returns failure?

src/openworld/src/autogenerate.c
Justin08784: The reason we have UP and DOWN is because we tried to support all six default directions (NESW + UP + DOWN). Actually, I am not 100% sure if there really are default directions hardcoded into chiventure.

MaxineK36 (approved):
Looks great!

Justin08784:
Comments by reviewers have been addressed and the relevant branch is now merged into `dev.`

MaxineK36:
Issue Score: ✔️+

Comments:
Good work overall! One area of improvement for next time:
- The reviewer for this PR approved the PR without comment. Please make sure that you always do a full review of the code, even if the senior developer has already provided their review. In the rare cases where there truly isn't anything to improve in the PR, make sure to leave a comment confirming that you reviewed the code and don't have any changes to suggest. If the senior dev has left their review, it can also be useful to indicate whether you agree with it or not.

Issue #271, opened 2019-06-04
Support for History Command


ecoble:
CLI previously supported "HIST" as a command, that printed out the user's history of commands. This command was causing a lot of issues during integration and testing, so we stopped supporting it. The goal of this issue is to debug the history operation and add support for it again.

Tianlel:
Closing statement:
I've successfully implemented the history command and I've added the hist option to the help_text.

dchen0258:
The goal is really the same as the one described before. Talking to Elizabeth, it seems like the committed code from above was close to finished but there simply wasn't enough time before the end of the project to complete a pull request.

dchen0258:
The old code in the cli/tianlel/history branch does not work with the updated chiventure. When building, there were undefined references to many of the history operations and function calls.

dchen0258:
Knowing this, the easiest thing to do is to use Tianle's old ideas and incorporate them into chiventure. 

This will include:
- Create a way to store all the operations that have been called
- Adding the HIST command back into the cmd.c file
- Printing the command when the HIST command is called 

This a high level way of looking at the problem. More details to come.

dchen0258:
The more detailed implementation:

- For storage: A linked list of strings (char arrays) might be the best way to store the operations, since we don't know how many commands they will call.
*Need to find where exactly a command is processed so it can be easily added to the linked list (or whatever data structure we end up choosing* - maybe we can take from cli/src/cmd.c line 231 with the variable char *s (decide whether to add commands that did not work or were NULL to list)
- For adding the HIST command back: simply uncommenting line 112 in the cli/src/cmd.c file should accomplish this
- For printing I saw the Tianle's use of "print _to_cli" for this which I think should still work. 

Other design decisions to be made: 
- If they call history, does that go in the history, does it reset it, or should it stay the same?
- I assume a NULL command call should not be included, but what about functions that are technically erroneous for chiventure? Should they be considered?

dchen0258:
Only thing left to consider is where exactly to initialize the linked list.

dchen0258:
Have implemented the function following most of the thoughts above with two main error/issues:
1. When HIST is called, it prints out the history, but only HISTs show up- the exact number of commands that would be in the history
2. It interprets "go north" as just "GO"

dchen0258:
Fixed the interpretation of multi-word commands so it interprets "go north" as "go north". Other issue is still being debugged.

dchen0258:
Submitted pull request.
[...]

PR #1009, opened 2021-05-16
Implementing support for the history command

dchen0258:
In relation to issue #271 
Goal: CLI previously supported "HIST" as a command, that printed out the user's history of commands. This command was causing a lot of issues during integration and testing, so we stopped supporting it. The goal of this issue is to debug the history operation and add support for it again.
Changes: Created new command list struct and new command list function in game state, add HIST entry to cmd.c, append command to linked list when new command is entered in cli in cmd.c, in operations.c for the hist_operation iterate through the linked list and print each command, changed the make file to include the new command.c file in game state folder


include/game-state/command.h
```diff
@@ -0,0 +1,24 @@
+#ifndef _COMMAND_H
```
ecoble, line 1: "Commands" are more related to the CLI/parsing aspect of Chiventure and don't actually affect the game state until later in the pipeline when they have been processed through action management. Instead of making this a game state module, I would move it into the CLI module.

ecoble (changes requested):
This looks great! There is a little bit of module refactoring to be done, but on the whole nice work!

include/cli/cmdlist.h
```diff
@@ -0,0 +1,22 @@
...
+#define _CMDLIST_H
+
+
+/* 
+ * The linked list is built to contain the command history
+ * of a user during a certain round
+ */
+typedef struct command_list
+{
+    char* command;
```
ecoble, line 11: Please be consistent with placement of *, looks like most other places you have type *name, so lets go with that.

include/cli/cmdlist.h
```diff
@@ -0,0 +1,22 @@
...
+    struct command_list *next;
+} command_list_t;
+
+
+/* 
+ * Creates a new command list stuct 
+ * sets the command, and sets the next command to NULL
+ */
+command_list_t* new_command_list(char* command);
```
ecoble, line 20: Same thing here about placement of *

include/game-state/game.h
```diff
@@ -45,7 +45,7 @@ typedef struct game {
 
     /* pointer to room that, when entered, ends the game */
     room_t *final_room;
-    
+
```
ecoble, line 48: Remove this whitespace change so that we aren't changing game state files.

src/cli/src/cmd.c
```diff
@@ -3,12 +3,14 @@
 
 #include "cli/cmd.h"
 #include "cli/operations.h"
+#include "common/utlist.h"
+#include "game-state/game.h"
```
ecoble, line 7: Do we still need game.h?

src/cli/src/cmd.c
```diff
@@ -232,11 +234,19 @@ cmd *cmd_from_tokens(char **ts, lookup_t **table)
 /* See cmd.h */
 cmd *cmd_from_string(char *s, chiventure_ctx_t *ctx)
 {
+
+    if (s != NULL) {
+        command_list_t* new_command = new_command_list(s);
+        LL_APPEND(ctx->command_history,new_command);
```
ecoble, line 240: Please put a space after the comma

src/common/src/ctx.c
```diff
@@ -41,6 +42,8 @@ int chiventure_ctx_init(chiventure_ctx_t *ctx, game_t *game)
     }
 
     lookup_t **table = lookup_t_new();
+    
+    ctx->command_history = new_command_list(NULL);
```
ecoble, line 46: I forgot to mention something important about using utlist, the head of the list should always be NULL. So instead of using new_command_list here, just set it equal to NULL.

src/cli/src/operations.c
```diff
@@ -26,11 +27,20 @@ char *help_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
     return NULL;
 }
 
-/* backlog task */
 char *hist_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 {
-    //print_history();
-    return "history operation not implemented yet\n";
+    command_list_t* temp = new_command_list(NULL);
```
ecoble, line 32: Same here for * placement

src/cli/src/cmd.c
```diff
@@ -232,11 +234,19 @@ cmd *cmd_from_tokens(char **ts, lookup_t **table)
 /* See cmd.h */
 cmd *cmd_from_string(char *s, chiventure_ctx_t *ctx)
 {
+
+    if (s != NULL) {
+        command_list_t* new_command = new_command_list(s);
```
ecoble, line 239: Same here for * placement

ecoble (changes requested):
A couple small things, mostly style changes.

src/cli/src/cmd.c
```diff
@@ -232,11 +233,18 @@ cmd *cmd_from_tokens(char **ts, lookup_t **table)
 /* See cmd.h */
 cmd *cmd_from_string(char *s, chiventure_ctx_t *ctx)
 {
+
+    if (s != NULL) {
```
mberamendiconde, line 237: I think Elizabeth mentioned in another PR that we should use the Allman style of indentation with,
 if (_____)
{
...
}

So I could just change the code to match that style. Otherwise this looks great!

src/cli/src/cmdlist.c
```diff
@@ -0,0 +1,19 @@
...
+#include <string.h>
+
+#include "cli/cmdlist.h"
+
+/* See cli/cmdlist.h for further documentation */
+command_list_t *new_command_list(char* command)
+{
+    command_list_t *new = malloc(sizeof(command_list_t));
+
+    if (command == NULL) {
```
mberamendiconde, line 11: Same thing here, I suggest Allman style indentation

src/cli/src/operations.c
```diff
@@ -26,11 +27,18 @@ char *help_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
 char *hist_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 {
-    //print_history();
-    return "history operation not implemented yet\n";
+    command_list_t *temp = new_command_list(NULL);
+
+    print_to_cli(ctx, "Start of command history: \n");
+    LL_FOREACH(ctx->command_history, temp)
+    {
+        if (temp->command != NULL) {
```
mberamendiconde, line 37: Suggest Allman style indentation here too!

mberamendiconde (commented):
Looks great overall! :-)

src/cli/src/cmd.c
dchen0258: ok makes sense, thanks!

ecoble (approved):

borjasotomayor (changes requested):
Adding the command history in `chiventure_ctx_t` sounds good to me, but I think we should instead create a `cli_ctx_t` type to store CLI-specific context information (similar to how there is a `ui_ctx_t` type), and place the command history there. That way, if we need to add additional CLI-related information to `chiventure_ctx_t`, it will be more clearly separated into component-specific contexts objects (if we do this, then it probably makes sense to move `lookup_t **table` in to that CLI context).

Since that will involve a fair amount of refactoring (given that the lookup table is used in several different places), we should just create a separate backlog issue for that refactoring. Please create that issue, and I will approve the PR once that's done.

dchen0258:
The new issue regarding the cli_ctx struct has been created. It is issue #1060.

borjasotomayor (approved):
LGTM! :shipit:

mberamendiconde (approved):
Looks great now! Appreciate the Allman style changes :-)

dchen0258:
This solution should work for the foreseeable future unless major changes are made. The only one I can think of is the adding a new cli struct, but that should only require some small refactoring of code!

ecoble:
Issue Score: ✔️++

Comments:
Great job implementing the HIST command, and using linked lists to do it! The summary and closing statement are great too, and great job @mberamendiconde for a thorough review!

PR #1010, opened 2021-05-17
Added fields to battle_item_t struct to distinguish between items/weapons

sonnykaiyuenlee:
I also added a function to generate one of ten random weapons. Added tests as well--all of which pass.


jintae412 (approved):
This seems to do exactly what you want it to. I wonder if as the Battle module continues to grow if it might be better to create an "Item Category" field of enums instead, but that's not strictly something that I think that we really need quite yet.

firatciftci (approved):
Everything looks good to me!

firatciftci:
Issue Score: ✔️

Comments:
Good work overall! I have two things to comment on:
- The summary for this PR is too terse. A good rule of thumb is that you should be writing at least one full paragraph in your summary (see https://uchicago-cs.github.io/cmsc22000/project/policies.html#tasks-issues for examples).
- When an PR is done, you should include a final comment summarizing what was merged to `dev` and what work remains to be done, if any.

PR #1014, opened 2021-05-17
Cli/action command

mberamendiconde:
Creating a pull request for cli/action_command branch, which implements the ability for the user to type "ACTIONS" into the command line and see a comprehensive list of the supported actions.


ecoble:
Following our discussion from the team meeting, this feature is basically a duplication of the existing HIST command, so we should close this issue/PR and potentially make the action commands something else, like showing the specific commands available on each item in the room.

Issue #981, opened 2021-05-11
Assign properties to items


tessied:
The DSL should allow properties to be attributed to specific items. Currently, the DSL grammar allows for properties to be associated with only the item above; this is not ideal and can be confusing when more items/rooms are added.

Assignee: Tessie and John
team: dsl

tessied:
I have modified the grammar defined for `property`, now we are able to assign the properties to a specific object (`[commit 64acc6a](https://github.com/uchicago-cs/chiventure/commit/64acc6a5ea9ee15b3879432efb864de6a39fd006#diff-af9289b34b6a239aea9e5dbda121cfbd39ae53c8f46d0a2456ffe9ace1efaaff)`). Instead of specifying properties using `property: value`, the dsl now supports `property FOR ROOM A: value` or `property FOR ITEM A: value`, for example. Now properties can be assigned to both rooms and items with a higher level of specificity.

This feature enhances the flexibility of assigning properties and decreases ambiguity. For example, if we declare a room followed by a declaration of an item, a property declared after that will be assigned to the item. But now with this feature, we are able to assign that property to the room itself or other items not immediately above the property.

tessied:
During a team meeting today, we decided that we wanted to support `property: value` without assigning them to specific items as well. `property FOR ITEM: value` and `property FOR ROOM: value` will be an option for users to specify properties if they choose to do so. 

I implemented this feature in the dsl grammar, now the grammar supports both `property: value` and `property FOR ITEM: value` (commit [`b7eda2f`](https://github.com/uchicago-cs/chiventure/commit/b7eda2fb43d8d465eed046989e5c30d4059a8763#diff-af9289b34b6a239aea9e5dbda121cfbd39ae53c8f46d0a2456ffe9ace1efaaff))

tessied:
The pull request linked to this issue reflects the changes made, that is, the grammar now allows users to assign properties to items if they want. Pull request has been merged, closing the issue.

tessied:
The pull request linked to this issue reflects the changes made, that is, the grammar now allows users to assign properties to items if they want. Pull request has been merged, closing the issue.

ecoble:
Issue Score: ✔️++

Comments:
Great job on this issue! Excellent documentation, especially connecting this to other issues and commits.

Issue #977, opened 2021-05-10
Support for escape characters in strings


jacknugent1529:
The dsl should allow users to include quotations marks, newlines, and other escape characters in strings using backslash syntax (e.g. `\"` for quotes).

jacknugent1529:
The dsl now supports escape characters. Simply use `\"`,`\t`, etc. and the character will be replaced by the Unicode equivalent. Now quotation marks can be included in strings without any issues.

jacknugent1529:
The pull request #1029 resolves this issue. The code for recognizing escaped characters in a string and replacing them with the proper unicode characters has been merged into dev, closing this issue.

ecoble:
Issue Score: ✔️++

Comments:
Great work, and nice examples!

Issue #931, opened 2021-05-04
Implement DSL grammar in code


jacknugent1529:
Our goal is to create a lark grammar file which describes the dsl language. This will occur after creating a document with the language specifications. As a result of this, lark will be able to parse the dsl into a parsing tree, which will later be transformed into the desired output.

Assignees: Jack and Tessie
team: dsl

r-e-stern:
We've defined the DSL grammar for its initial iteration in #937, and it should now be convert-able to a Lark grammar.

tessied:
We had a meeting where we discussed the current implementation of the DSL grammar. We brought up a few issues that we hope to discuss with the whole team:
- Newline-agnosticism is tricky to implement. We proposed that having a new line for each declaration might not be a bad idea as it increases readability and writability for the author as well. 
- The end condition is not necessarily the room in which the player ends the game, as Jack mentioned above. Therefore, we may need to modify the way in which we define a game at the top to accommodate the various end conditions we may be given.

Currently, we have all the basic features outlined in the DSL design guide. I will now attempt to add actions to the grammar.

tessied:
I've implemented the "actions" feature to the dsl grammar (commit [a0e8382](a0e83823147555d76fe15a6c626d4f2328f5b6bf)). I also added a short example dsl file to test the new actions feature. However, actions currently need to be defined separately. For example, if one wants to specify the actions `TAKE` and `EAT`, they will have to define two actions: `action: TAKE` and `action: EAT` and specify the conditions, success and failure outcomes separately for each action. We will figure out how to fix the implementation to support `actions: TAKE, EAT` and take into account the conditions that follow each of the actions.

In addition, I am currently only able to run the tree that lark produces. I will add to dsl_parser.py so that we can have actions in the python dictionary output.

jacknugent1529:
[...]

PR #1029, opened 2021-05-18
Added DSL grammar file: Implementation of the core features of the DSL grammar 

tessied:
We have implemented the grammar for a few of the core features of the DSL. This pull request consists of a lark grammar file which describes the DSL language (the file named `dsl_grammar.lark`), according to the [user stories](https://github.com/uchicago-cs/chiventure/wiki/DSL-~-User-Stories) and the [design](https://github.com/uchicago-cs/chiventure/wiki/DSL-~-DSL-Grammar) we came up with. 

Some of the features that were implemented in this grammar are:
- Declaring a game
- Declaring a room (with connections, properties, items)
- Declaring an item (with properties, actions)
- Declaring variables
- Adding escape characters for strings
- Attributing properties to items

This pull request also includes a README, a parser file as a placeholder for future implementation of the parser, and a few examples for demonstration. This pull request reflects the implementation done in issues #931, #977, and #981.


src/dsl/examples/distaff.dsl
```diff
@@ -0,0 +1,47 @@
...
+
+   ITEM sconce
+      short desc: "A sconce holding a candle"
+      # intentionally bad indentation
+long desc: "It looks a bit loose."
+
+ROOM room B
+   short desc: "A dungeon room."
+   long desc: "The walls are damp and moldy and, yet, the scent of freshly
+
```
ecoble, line 28: Are these extra lines intentional?

src/dsl/examples/distaff.dsl
```diff
@@ -0,0 +1,47 @@
...
+ROOM room B
+   short desc: "A dungeon room."
+   long desc: "The walls are damp and moldy and, yet, the scent of freshly
+   
+   
+   cut lavender flowers pervades the air."
+   connections: NORTH TO room A
+                 EAST TO room C
+   ITEM Door IN room B
```
ecoble, line 33: Is it intentional that this is the only item where the room is specified like this?

src/dsl/src/dsl_parser.py
```diff
@@ -0,0 +1,81 @@
...
+from lark.lexer import Token
+import json
+from pathlib import Path
+
+base_path = Path(__file__).parent
+grammar_f = open(base_path / "dsl_grammar.lark")
+dsl_grammar = grammar_f.read()
+grammar_f.close()
+
+class TreeToDict(Transformer):
```
ecoble, line 12: For all of these functions, I think it would be useful to write a comment saying what all these single letter variables are (s, k, v) and their type. If they are the same across all the functions, you can just put one comment for it at the top.

ecoble (changes requested):

src/dsl/examples/distaff.dsl
```diff
@@ -0,0 +1,47 @@
...
+      # intentionally bad indentation
+long desc: "It looks a bit loose."
+
+ROOM room B
+   short desc: "A dungeon room."
+   long desc: "The walls are damp and moldy and, yet, the scent of freshly
+   
+   
+   cut lavender flowers pervades the air."
+   connections: NORTH TO room A
```
r-e-stern, line 31: The [docs](https://github.com/uchicago-cs/chiventure/wiki/DSL-~-User-Stories#simple-declarations) seem to indicate that we want a comma-separated list for connections, but this works too if we want, we just have to update the docs accordingly.

src/dsl/examples/short_example.dsl
```diff
@@ -0,0 +1,15 @@
+GAME START room B END room C
+   intro: "Welcome!"
+
+
+ROOM room B
+   short desc: "A dungeon room."
+   ITEM Door IN room B
+    short desc: "A large wooden door"
+    action: OPEN, BREAK
```
r-e-stern, line 9: This doesn't actually parse correctly as a list of actions; instead it thinks OPEN is the property and the rest of the inside is the value.

r-e-stern (changes requested):
There's a lot more flexibility in the parser now than the documentation originally imagined (re: connections, etc.), so we should remember to note that.

Actions don't parse correctly quite yet.

src/dsl/examples/distaff.dsl
tessied: Yes, these lines are the first half of the room declaration for `room B`. We broke up the `long desc` into two lines with a few blank lines in between to demonstrate that we have defined our grammar generously so that the parser does not break even if there are extra lines here and there. I have made the comment  `#intentional line break` to make this clear in the file.

src/dsl/examples/distaff.dsl
tessied: Yes, this is the second half of the declaration for `room B`. We added the blank lines in between these two halves to show that the parser works with line breaks (see previous [comment](https://github.com/uchicago-cs/chiventure/pull/1029#discussion_r636410162) regarding line breaks in `long desc`) . For the `ITEM Door`, we also intentionally declared that it is in room B to show that our parser allows users to specify the locations of items if they wish.

src/dsl/examples/distaff.dsl
tessied: We have defined the grammar such that it gives the user the option to specify connections using either a comma-separated list or line breaks — both work properly. We could note this flexibility in the documentation.

src/dsl/examples/short_example.dsl
jacknugent1529: Thanks for pointing that out. Previously the parser correctly interpreted it, but the transforming function had errors. It should be fixed now so that all action ids are recognized and properties are associated with the correct actions.

r-e-stern (approved):

src/dsl/src/dsl_parser.py
jacknugent1529: The variable `s` represents what is passed to the function by Lark (it seemed to be common notation in the examples). As we're "compressing" the parsing tree, the key, `k` is always a string, while the value may take on different formats. We could list all of these, but in the future there might be additions. I added comments in commit eef06d1 to specify what the value type is in relevant locations.

src/dsl/examples/distaff.dsl
ecoble: Great! What does that mean for items that are not specified in a specific room? Are they still located in the room they are under?

src/dsl/distaff.dsl
```diff
@@ -0,0 +1,49 @@
+# This is a dsl version of the distaff game. Actions have not yet been
```
ecoble, line 1: Do we need this file? It looks the same as the one in the example directory.

tessied: Thank you for pointing that out! That was indeed an extra file that we did not have to include. I have since deleted the file (commit [`de2149a`](https://github.com/uchicago-cs/chiventure/pull/1029/commits/de2149a21ec1b354919391c8253be64260470d18)).

src/dsl/examples/distaff.dsl
tessied: Yes! For items that are not specified in a room, the default case applies and the item is located in the room that they are under. The same is also true for properties that are not assigned to a specific item - the default case associates them with the item/room they are under.

src/dsl/examples/short_example.dsl
```diff
@@ -0,0 +1,15 @@
...
+      OPEN condition: "Door is in front of you."
```
r-e-stern, line 10: The OPEN condition does not parse.

r-e-stern (changes requested):

src/dsl/examples/short_example.dsl
jacknugent1529: That was due to a slight indexing error. It is fixed now

r-e-stern (approved):

ecoble (approved):
LGTM!

tessied:
Closing statement: We added a DSL grammar file in `src/dsl` along with a parsing file and a few examples. Since opening the pull request, we made several changes to ensure that the features parse correctly. The changes have been merged and the `dev` branch now has the DSL grammar file, closing the pull request.

ecoble:
Issue Score: ✔️++

Comments:
Great job, the DSL is looking amazing! The documentation in the closing statement and summary are great as well, and thanks @r-e-stern for such a thorough review!

PR #1036, opened 2021-05-18
Fixing chiventure segfault when trying to load wdl file

### LINKED ISSUE #1030: WDL path seg fault
Currently when the user inputs an incorrect wdl file path chiventure causes a segfault, instead of just saying something along the lines of "Invalid file path!"

samdonio:
Previously chiventure would have a seg fault when the user tried using the load_wdl command at all. This is because the cli didn't correctly check if the path was valid, and also because the wdl module doesn't currently load wdl files correctly. Now, the cli returns that the path is invalid and that the wdl loading isn't complete yet and doesn't seg fault anymore.

Closes #1030


samdonio:
After more changes to the way that load_wdl_operation works, the LOAD_WDL command now checks whether the file exists, then checks if the file has the ending of .wdl, and if either of these are not true it tells the user that the path/file is invalid. Also, the function now correctly loads a chiventure game if the path is correct, meaning that the user can now use LOAD_WDL to load wdl files from inside the game whenever they want.

src/cli/src/operations.c
```diff
@@ -53,11 +54,56 @@ bool validate_filename(char *filename)
     }
 }
 
+bool validate_wdl_filename(char *filename)
```
ecoble, line 57: I think it is safe to remove the previous validate_filename function, I don't think we use .dat files anymore for Chiventure.

src/cli/src/operations.c
```diff
@@ -53,11 +54,56 @@ bool validate_filename(char *filename)
...
+{
+    int len = strlen(filename);
+    int min_filename_length = 4;
+    if(len < min_filename_length)
+    {
+        return false;
+    }
+    const char *ending = &filename[len - 4];
```
ecoble, line 65: 4 is a bit of a magic number here, can we create another variable, maybe something like "file_extension_length"

ecoble (changes requested):

src/cli/src/operations.c
samdonio: file_extension_length int has been made

src/cli/src/operations.c
samdonio: This function has been removed from the operations.c and operations.h file

src/cli/src/operations.c
```diff
@@ -33,31 +34,58 @@ char *hist_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
 }
 
-bool validate_filename(char *filename)
+
+bool validate_wdl_filename(char *filename)
 {
     int len = strlen(filename);
     int min_filename_length = 4;
+    int file_extension_length = 4;
     if(len < min_filename_length)
```
ecoble, line 43: Spacing after "if"

src/cli/src/operations.c
```diff
@@ -33,31 +34,58 @@ char *hist_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
-    if(cmp == 0)
+    const char *ending = &filename[len - file_extension_length];
+    int cmp = strcmp(ending, ".wdl");
+    if (cmp == 0)
     {
         return true;
     }
-    else
-    {
+    else{
```
ecoble, line 53: Looks like the rest of this file is using allman style brackets, lets stick with that.

src/cli/src/operations.c
```diff
@@ -33,31 +34,58 @@ char *hist_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
+    {
+        ch = tolower(tokens[1][i]);
+        tokens[1][i] = ch;
+        i++;
+    }
+
+
+    valid_path = access(tokens[1], F_OK);
+
+    if(valid_path == -1) //Triggers if file does not exist
```
ecoble, line 82: Spacing

src/cli/src/operations.c
```diff
@@ -66,7 +94,7 @@ char *load_wdl_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 
     if(game == NULL)
```
ecoble, line 95: Spacing

ecoble (changes requested):
Looks great! Just a couple style fixes, you can probably just fix them with astyle.

dchen0258 (approved):
Looks good! Very neat solution.

ecoble:
Issue Score: ✔️+

Comments:
Great job, thanks for fixing this! Be sure to put closing statements on your PRs as well, just as you would for issues.

PR #1038, opened 2021-05-18
Migrate CI from Travis to GitHub Actions

borjasotomayor:
This pull request removes the Travis configuration file and replaces it with a GitHub Actions workflow file, so that CI is done on GitHub Actions. Here is an example of a successful build on GitHub Actions: https://github.com/uchicago-cs/chiventure/runs/2613593469


borjasotomayor:
Note, once this pull request is approved, the branch protection rules for `master` and `dev` will need to be updated to require a GitHub Actions check, not a Travis check.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,59 @@
+name: Build and Test chiventure
+
+on:
+  push:
+    branches:
+      - master
+      - dev
```
firatciftci, line 7: Is there a reason for us to limit the GitHub Action to only execute for the `master` and `dev` branches? I am not sure if there is a free/paid tier distinction and/or a cap on how many actions can be run for a given repository, but if there is not, I think it could be useful for the students to utilize automated tests in their respective branches before they create a PR to merge to `dev`.

firatciftci (commented):
I have added a minor comment, but apart from that everything looks good!

.github/workflows/build-and-test.yml
borjasotomayor: It looks like we're allowed to have up to 60 concurrent jobs, and each build job shouldn't take more than 1 minute or so to run, which means we can probably handle building on every branch. I've updated the workflow file so it builds on every push (I initially set it to also build on pull requests, but that actually results in duplicate builds if you're already unconditionally building on all pushes)

firatciftci (approved):
Looks good to me!

Issue #1004, opened 2021-05-15
Provide input suggestions to user when input is misspelled 


mberamendiconde:
Want to create a feature by which if a user types in a word not supported by chiventure but similarly spelled to a supported input, it will be suggested to the user. 

For example, if the user meant to type 'take" but types "tak" , the user will see "did you mean: take" printed to the window

mberamendiconde:
Work for this happening in the cli/suggestions branch

mberamendiconde:
Currently when full words are input, the action word is returned to the screen.
Now, when loading an example game, there is a seg fault. No sure where it could be coming from. All new functions are in src/cli/src/operations.c

dchen0258:
The bug was fixed by using a for loop instead of a while loop.
Also a different use of strncat fixed another bug.

dchen0258:
Submitted pull request.

dchen0258:
Branch merged!

dchen0258:
There are possible enhancements to the issue: allowing multiple suggestions, not only based on the number of matching letters, but also the size of the words. This would require a much more length function, and probably can be an enhancement for the future!

ecoble:
Issue Score: ✔️++

Comments:
Great work! Nice documentation, and thoughts on possible enhancements!

PR #1041, opened 2021-05-18
Implementing support for input suggestions when input is misspelled

dchen0258:
In relation to issue #1004.
Goal: To create a feature by which if a user types in a word not supported by chiventure but similarly spelled to a supported input, it will be suggested to the user. For example, if the user meant to type 'take" but types "tak" , the user will see "did you mean: take" printed to the window, rather than "command not supported".
Changes: Created a new array that contains all possible actions. In the function "action_error_operation" rather than just printing out "command not supported", it compares the input to each action in the list, and returns the one that is most alike.


src/cli/src/operations.c
```diff
@@ -1,13 +1,69 @@
...
+#define min(x,y) (((x) <= (y)) ? (x) : (y))
+
+char* actions_for_sug[NUM_ACTIONS] = {"OPEN", "CLOSE", "PUSH", "PULL", "TURNON", "TURNOFF", 
+                        "TAKE", "PICKUP", "DROP","CONSUME","USE","DRINK",
+                        "EAT", "GO", "WALK", "USE_ON", "PUT", "QUIT","HIST", "HELP",
+                        "CREDITS", "LOOK", "INV", "MAP", "SWITCH", "LOAD_WDL", "NAME", 
+                        "PALETTE", "ITEMS"};
+
+
+int compare(char* word, char* action, int initial)
```
ecoble, line 22: Since these are helper functions without their own definition in the .h file, you should include header comments for these functions here.

src/cli/src/operations.c
```diff
@@ -1,13 +1,69 @@
...
+    for (int i = 0; i < min(strlen(word), strlen(action)); i++){
+        if(&action[i] != NULL && &word[i] != NULL) {
+            if (word[i] == action[i]){
+                current++;
+            }
+        }
+
+    }
+
+    if (current > initial){
```
ecoble, line 35: Is there a reason to set initial = current here? It seems like the point is just to return the greater of current and initial, in which case you can just return current if it's greater than initial or initial if it's not. Please correct me if I'm missing something.

src/cli/src/operations.c
```diff
@@ -1,13 +1,69 @@
...
+        if (action_input != NULL) {
+            temp = compare(strdup(action_input), strdup(actions[i]), initial);
+            if (temp > initial) {
+                index = i;
+                initial = temp;
+            }
+        }
+    }
+    
+    if (index == -1) {
```
ecoble, line 59: You do not need the else part of this statement, since you have a return statement within the if. You could also simplify this to use the ternary conditional operator (though it is not necessary if you don't want to).

src/cli/src/operations.c
```diff
@@ -261,9 +317,27 @@ char *kind3_action_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ct
...
+    suggestion = suggestions(strdup(tokens[0]), actions_for_sug);
+
+    if (suggestion != NULL) {
+        int str1 = strlen(suggestion);
+        int str2 = strlen("This action is not supported. Did you mean: ");
+        int len = str1 + str2;
+        char msg[] =  "This action is not supported. Did you mean: ";
+        print_to_cli(ctx, strncat(msg, suggestion, len));
+        return "";
+    } else {
```
ecoble, line 338: Same here about not needing this inside an else.

src/cli/src/operations.c
```diff
@@ -261,9 +317,27 @@ char *kind3_action_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ct
 }
 
 
+
 char *action_error_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 {
-    return "This action is not supported.";
+
+    if (tokens[0] == NULL) {
```
ecoble, line 324: This seems like a bit of a special case, it may be worth telling the user specifically that their action was NULL, because it could indicate a larger bug which is useful information.

ecoble (changes requested):

src/cli/src/operations.c
samdonio, line 35: Besides what Elizabeth said, the style of this should probably be corrected to fit with the rest of chiventure, as we are using Allman style so having the brace on the same line as the for and if's is incorrect. This is also a little pedantic but there maybe shouldn't just be the empty line on line 32. This can probably be fixed quickly with astyle.

src/cli/src/operations.c
```diff
@@ -1,13 +1,69 @@
...
+{
+    int i = 0;
+    int initial = 0;
+    int temp = 0;
+    int index = -1;
+    
+    for(int i = 0; i < NUM_ACTIONS; i++){
+        if (action_input != NULL) {
+            temp = compare(strdup(action_input), strdup(actions[i]), initial);
+            if (temp > initial) {
```
samdonio, line 52: Again, allman style.

src/cli/src/operations.c
samdonio, line 59: allman

src/cli/src/operations.c
samdonio, line 324: allman

src/cli/src/operations.c
samdonio, line 338: allman

samdonio (changes requested):
There didn't seem to be any problems besides what Elizabeth had pointed out, but just some style changes should be made. Otherwise, nice work!

ecoble (approved):
Great job!

samdonio (approved):
The style looks much better now, great job!

dchen0258:
Closing statement: 
There are possible enhancements to the issue: allowing multiple suggestions, not only based on the number of matching letters, but also the size of the words. This would require a much more length function, and probably can be an enhancement for the future!

ecoble:
Issue Score: ✔️++

Comments:
Great job on this issue! This was definitely a tougher task, and you implemented a great solution!

Issue #1042, opened 2021-05-18
Refactor the class_t struct to be a hashtable.


jackcasey067:
This is the first step for being able to load in classes from the WDL++ format.  We need someplace in memory to store all of the classes are in the JSON-style game files.

I am currently thinking about the following changes.  It is possible that some will not make it into this issue:
- Turn class_t into a hashtable.
- Add convenience functions for accessing and manipulating the hashtable.
- Add a field to the game_t struct to store the hashtable (this is where stats_hash_t and effects_hash_t lives).
- While I am refactoring the class struct, it might be a good time to finalize the class_t struct.  We have been thinking about combining the two skill inventories, for example.  Attributes may also be on the chopping block.  Also, we might want a field that denotes whether or not a class is for players or for NPCs/enemies.

Note that these changes would be followed by updating tests and documentation.

jackcasey067:
The action-management team approved of adding the hashtable to the game_t struct.  I will add them as a reviewer to any PR that results from this issue.

jackcasey067:
Update:
 -  The hashtable struct was added.
 - The following functionality exists and have passing tests:
   - Add classes to a hashtable
   - Add classes, replacing in case of a name clash (instead of returning FAILURE)
   - Count classes in hashtable
   - Delete and Free classes in hashtable
   - Find classes in hashtable
 Overall, I want using the class hashtable to feel less forbidding than using the MACROS.

jackcasey067:
Update on changes to the `class_t` struct:
- The two skill inventories (`combat` and `noncombat`) are being fused.  This will mean that a skill tree can have skills of both types with dependencies on each other.
    - This will require a change to the `player_t` as well, since it also has two skill inventories.
    - These changes were approved by the skill-trees and action-management teams.
- The `attributes` field stays, even though we currently don't use it much.  It could be especially useful in the future if classes are extended for use generating npcs and enemies (some enemies could have the "undead" trait, for example, and this could have gameplay effects regarding the use of certain items or skills (ie: "Holy Water")).
- The `effects` field also stays, more so as something that future teams could use.
- To clear up confusion, I am changing `stats` to `base_stats`, but the functionality is the same.

jackcasey067:
I implemented those changes, which meant changing very small parts of a lot of files.  I am going to open a PR, and then this issue will be closed.

jackcasey067:
Closing comment:
[...]

PR #1061, opened 2021-05-19
Playerclass/hashtable: Refactor class_t, and make it a hashtable.

jackcasey067:
This is connected to issue #1042.  Originally, I was looking at how we would load classes into memory from WDL, and I noticed that there is nothing currently that holds every class declared by a game author.  Therefore, I decided to make a class hashtable, which stores every class in the game.

Since doing this would involve modifying the class_t struct itself (adding a hashtable handle `hh`), I decided to use this as a moment to reconsider the fields in class_t.  This would need to be done before moving on to WDL class -> in-memory class conversion anyway, since doing that will involve specifying what JSON classes should look like.

The following changes were made (in broad strokes):
- I made class_t a hashtable by adding a UTHASH handle to the struct.
- I added functions for manipulating the hashtable, so that future users will not have to suffer through the somewhat confusing macros.
- I added tests for the new functions.
- I fused the noncombat and combat skill inventories into one `starting_skills` inventory.  I also did this in player_t.
- I renamed stats to base_stats, for clarity.  At one point, class_t contained a player's currently updated stats, but that is no longer the case.  In fact, all of the fields in class_t are meant to be static and unchanging under normal circumstances now (changing info exists in player_t, npc_t, and combatant_t).


WPDolan (approved):
I am reviewing this PR on behalf of the action management team, and am thus only looking at the changes that affect game-state modules. All of the changes that concern game-state (combining the skill inventory fields in the `player_t` struct in the player module) make sense and will reduce cognitive load when handling the player's skills. 

These changes will cause merge conflicts with [Pull Request 1011](url) because 1011 implements functions that depend on the old skill inventory fields. Fortunately, the merge conflict should only involve refactoring the player's skill inventory functions to consider the one combined inventory.

jackcasey067:
I merged dev into this branch, and fixed the problems that @WPDolan warned about above.

include/playerclass/class.h
```diff
@@ -77,4 +76,85 @@ int class_add_skills(class_t* class, skill_inventory_t* combat,
...
+ * 
+ * Parameters:
+ *  - hashtable: A pointer to a possibly NULL class hashtable pointer.
+ *  - class: The class being added to the hashtable.
+ * 
+ * Returns: 
+ *  - FAILURE if the class or its name field were NULL.
+ *  - SUCCESS on successful addition or replacement.
+ */
+int add_or_replace_class(class_hash_t** hashtable, class_t* class);
```
ideclue, line 114: I'm not sure if the language "add_or_replace" is the best choice here. I think "set" is typically used when it's assumed something will either be added or replaced.

include/playerclass/class.h
```diff
@@ -77,4 +76,85 @@ int class_add_skills(class_t* class, skill_inventory_t* combat,
...
+ * 
+ * Returns:
+ *  - A pointer to a the class that matches the provided name.
+ *  - NULL if no class was found, or if either parameter was NULL.
+ */
+class_t* find_class(class_hash_t** hashtable, char* name);
+
+/*
+ * Removes a class from a hashtable and frees it.
+ *
```
ideclue, line 131: Design question: Why is the free function not separated? It might be useful to have these operations separated for easier debugging. Also, if there is a scenario where we would like to use a class that is not in a hash table.
I would consider changing "delete" to two separated functions: "remove" and "free"

src/playerclass/src/class.c
```diff
@@ -106,14 +104,71 @@ int class_free(class_t* class) {
...
+
+/* See class.h */
+int add_class(class_hash_t** hashtable, class_t* class) {
+    if (class == NULL || class->name == NULL)
+        return FAILURE;
+
+    if (find_class(hashtable, class->name) != NULL)
+        return FAILURE;
+
+    HASH_ADD_STR(*hashtable, name, class);
```
ideclue, line 126: It's a little concerning that there's no check if this function, HASH_ADD_STR successfully adds the class. Would it be possible to return result code?

src/playerclass/src/class.c
```diff
@@ -106,14 +104,71 @@ int class_free(class_t* class) {
...
+/* See class.h */
+int add_or_replace_class(class_hash_t** hashtable, class_t* class) {
+    if (class == NULL || class->name == NULL)
+        return FAILURE;
+    
+    class_t* replaced_class = NULL;
+    HASH_REPLACE_STR(*hashtable, name, class, replaced_class);
+
+    if (replaced_class != NULL)
+        class_free(replaced_class);
```
ideclue, line 139: Same issues as with "add_class": no result code returned

tests/playerclass/test_class.c
```diff
@@ -110,3 +104,153 @@ Test(test_class, free)
...
+                 "delete_class failed to delete Warrior.\n");
+    cr_assert_eq(count_classes(&hashtable), 4, 
+                 "delete_class returned SUCCESS, but did not actually removed Warrior.\n");
+
+    /* Trying to delete a non existing class */
+    cr_assert_eq(delete_class(&hashtable, "Alchemist"), FAILURE, 
+                 "delete_class returned SUCCESS while trying to delete nonexistent Alchemist.\n");
+    cr_assert_eq(count_classes(&hashtable), 4, 
+                 "delete_class returned FAILURE, but there are now fewer classes than expected.\n");
+}
```
ideclue, line 256: This test majes no reference to the "freeing" component of "delete." Hence why "freeing" and "removal" should be separated.

ideclue (changes requested):
I reviewed the portions of the code relevant to the `playerclass` module. Overall looks good, but I think the function design for a couple of the hash table functions (delete and add_or_replace) should be reviewed and maybe updated. Documentation looks great!

include/playerclass/class.h
jackcasey067: I think I agree, this `set_class` is a better name.

src/playerclass/src/class.c
jackcasey067: HASH_ADD_STR doesn't return anything, unfortunately, so I don't have access to whether or not it failed.

Also, I trust the macros a lot.  If they fail, it's our fault for putting the wrong input in, which is why I check and return FAILURE in advance.  

Good idea though; if HASH_ADD_STR did return an error code, it would be a good idea to pass that in our return.

src/playerclass/src/class.c
jackcasey067: Similarly, HASH_REPLACE_STR does not return an error code.  See above.

include/playerclass/class.h
jackcasey067: I am iffy on this one, and could go either way.  I think freeing the memory in a class is something we should do in the `delete` function, to pull down complexity just a little bit.  I don't think it is likely that someone in the future will want to remove a class from the hashtable and still use it.  I am more concerned about someone assuming a function called delete (or remove) frees a struct when it actually does not free it, causing insidious memory leaks.  What do you think?

ideclue: @firatciftci What do you think of this design question?

ideclue (approved):
I am approving the PR as it now reflects some of the design changes I suggested. However, I still believe it might be useful to separate the "freeing" and "removing" elements of the `delete` function.

WPDolan (approved):
The functions associated with the now removed `player_noncombat_skills` and `player_combat_skills` skill inventories have been successfully refactored to account for the new `player_skills` field. 

I have no issues with the way that the functions themselves were changed, and the changes made to the skill inventory tests stick to the original philosophies that the action-management team had when designing them.

include/playerclass/class.h
firatciftci: I think this is fine as is – in terms of the use case of the function, I agree with @jackcasey067 that the expected use case will probably always be deleting the struct from the hash table and free'ing it at the same time. If a need for freeing the struct from memory separately arises in the future, the developers then can easily change this. For now, let's keep it as is.

firatciftci (approved):
Everything looks good to me, great work!

jackcasey067:
Closing comment: After making a few suggested changes, and resolving a few clashes that occurred with another PR, this branch was merged into `dev`.  Classes are now hashable.

firatciftci:
Issue Score: ✔️++

Comments:
Great work!

PR #802, opened 2020-06-08
Implenting Do Custom Action - Running Branch, Conditional, and Control Blocks

sonkinss:
This is a remake of a previous pull request listed here: https://github.com/uchicago-cs/chiventure/pull/782

Files that need to be checked: interface.c, interface.h, and test_interface_functions.c.

The files mentioned above are related to our function of do_custom_action, which is what our entire team relies around. Here, I've implemented functions to "run" 3 different types of Action Sequence Tree Blocks, which are the defining elements of a custom action in our implementation.

PR #1065, opened 2021-05-20
Implement Do Custom Action - Running AST Action Blocks

LapsededPacifist:
This pull request is the replacement of [PR 802](https://github.com/uchicago-cs/chiventure/pull/802).
It adds functions to evaluate conditional blocks, run AST_blocks (Which only is implemented for action blocks), and run custom actions. Tests were added to these functions in their respective test files.
Modifications were also made to the conditional_type enum, so that eval_conditional_block could distinguish between less and and greater than (and LTE/GTE).


src/custom-actions/include/interface.h
```diff
@@ -12,6 +14,30 @@
...
+/* 
+ * Given a custom action and its corresponding arguments, 
+ * attempt to execute the given action.
+ * 
+ * Parameters: 
+ * - action: A pointer to the custom action to be executed (most likely
+ *   acquired from search_for_custom_action)
+ * - args: An array of strings of the arguments passed to the action
+ *   (i.e. the words succeeding the action written in the command line)
+ * - num_args: The number of arguments being passed in
```
jasulcaf, line 32: There are 3 parameters specified but the function `do_custom_action()` only takes in one variable (the action)

src/custom-actions/src/ast_block.c
```diff
@@ -88,3 +88,29 @@ int AST_block_free(AST_block_t *ast)
 
     return SUCCESS;  
 }
+
+/* See ast_block.h */
+int run_ast_block(AST_block_t *block)
+{
+  int returnV;
+  if(block == NULL) return SUCCESS;
```
jasulcaf, line 96: A bit of a nit pick but shouldn't the check, if block == NULL, be the first line since there is no need to declare int returnV if that doesn't pass? (Also, the return SUCCESS should be on a new line, not on the same line as the if statement).

src/custom-actions/src/ast_block.c
```diff
@@ -88,3 +88,29 @@ int AST_block_free(AST_block_t *ast)
...
+      break;
+    case(BRANCH):
+      returnV = do_branch_block(block->block->branch_block);
+      if(returnV == 0) return run_ast_block(block->prev);
+      return run_ast_block(block->next);
+      break;
+    case(ACTION):
+      if(exec_action_block(block->block->action_block) == FAILURE)
+	return FAILURE;
+      return run_ast_block(block->next);
```
jasulcaf, line 110: The if statement "return" should be indented (and in a new line for the case BRANCH). At first I thought this was calling two returns on every case but upon looking at it more, I realized it was just an if statement return.

src/custom-actions/src/conditional_block.c
```diff
@@ -94,3 +95,30 @@ int conditional_block_free(conditional_block_t *conditional)
...
+  case GTB: // once conditional_block struct changed
+    return check_gt(block->left,block->right);
+    break;
+  case LTEB: // functioning like case LTE at the moment
+    return check_lte(block->left,block->right);
+    break;
+  case GTEB: // once conditional_block struct changed
+    return check_gte(block->left,block->right);
+    break;
+
```
jasulcaf, line 118: Extremely Nit-Picky but the extra line probably shouldn't be there since all the previous cases where one after another

src/custom-actions/src/interface.c
```diff
@@ -9,6 +9,60 @@
...
+            case WHILEENDWHILE:
+                // not implemented yet
+                break;
+            case FORENDFOR:
+                // not implemented yet
+                break;
+        }
+    }
+    return i;
+    */
```
jasulcaf, line 52: I am guessing this is work that can be done in the future correct? My only comment here would be if this future work has been documented as "to do" in some other place (maybe in a summary or even as a open Issue).

LapsededPacifist: This will be documented as to do soon. I just haven't quite gotten around to creating an issue for it. I will add documentation here when that issue is created as well as in the summary.
Related to this, I wasn't sure if I should drop this comment or not. This function will eventually be moved to a different file, and since branch_blocks are going to be totally redone, I don't think anything here will be helpful.

tests/custom-actions/test_interface_functions.c
```diff
@@ -0,0 +1,67 @@
...
+    branch_block_t* new_branch = branch_block_new(1, &conditionals, conditional_type, 2, &controls);
+    block_t* block = (block_t*)new_branch;
+    AST_block_t* ast = malloc(sizeof(AST_block_t));
+    ast->block = block;
+    ast->block_type = BRANCH;
+    ast->prev= NULL;
+    ast->next = NULL;
+    custom_action_t* ca = custom_action_new("act_PUSH","item","obj_CHAIR","paladin",ast);
+
+}
```
jasulcaf, line 67: I am a bit confused as to why there was the creation of all these blocks but there was no cr_assert_<> tests following. In simple words, I am confused as to what exactly this test is testing for.

LapsededPacifist: Currently it is testing nothing as branch blocks don't exist. This was a test written from the previous group, and I wasn't sure what to do with it. I probably should have just deleted it, but I had kind of thought that maybe it would be used again someday. 
I will just go back and delete it now.

jasulcaf: I see. In that case, either comment it out explaining what it could be used for in the future, or delete it seems like the best options. Either works fine.

jasulcaf (approved):
After changing the (minor) things I had pointed out, looks good now!

src/custom-actions/src/ast_block.c
```diff
@@ -88,3 +88,37 @@ int AST_block_free(AST_block_t *ast)
...
+    return SUCCESS;
+  }
+  int returnV;
+  switch(block->block_type)
+    {
+    case(CONTROL):
+      return FAILURE;
+      break;
+    case(BRANCH):
+      returnV = do_branch_block(block->block->branch_block);
```
MaxineK36, line 106: Conventionally, `returnV` should be `rc` for "return code"

src/custom-actions/src/ast_block.c
```diff
@@ -88,3 +88,37 @@ int AST_block_free(AST_block_t *ast)
...
+      if(returnV == 0)
+      {
+        return run_ast_block(block->prev);
+      }
+      return run_ast_block(block->next);
+      break;
+    case(ACTION):
+      if(exec_action_block(block->block->action_block) == FAILURE)
+      {
+	return FAILURE;
```
MaxineK36, line 116: nit: line 116 should be indented

src/custom-actions/src/ast_block.c
```diff
@@ -88,3 +88,37 @@ int AST_block_free(AST_block_t *ast)
 
     return SUCCESS;  
 }
+
+/* See ast_block.h */
+int run_ast_block(AST_block_t *block)
+{
+  if(block == NULL)
```
MaxineK36, line 95: nit: spacing

src/custom-actions/src/ast_block.c
```diff
@@ -88,3 +88,37 @@ int AST_block_free(AST_block_t *ast)
...
+      break;
+    case(ACTION):
+      if(exec_action_block(block->block->action_block) == FAILURE)
+      {
+	return FAILURE;
+      }
+      return run_ast_block(block->next);
+      break;
+    case(CONDITIONAL):
+      return FAILURE;
```
MaxineK36, line 121: why does conditional return failure?

src/custom-actions/src/ast_block.c
```diff
@@ -88,3 +88,37 @@ int AST_block_free(AST_block_t *ast)
...
+      break;
+    }
```
MaxineK36, line 123: You should have a default case

src/custom-actions/src/conditional_block.c
```diff
@@ -94,3 +95,29 @@ int conditional_block_free(conditional_block_t *conditional)
 
     return SUCCESS; 
 }
+
+/* See conditional_block.h */
+int eval_conditional_block(conditional_block_t *block)
+{
+  switch(block->conditional_type) {
+  case EQ:
+    return check_eq(block->left,block->right);
```
MaxineK36, line 104: nit: spacing after comma (on all lines)

MaxineK36 (changes requested):
A couple of small fixes/clarifications

src/custom-actions/src/ast_block.c
LapsededPacifist: I added a brief comment about this, but basically it is just because it isn't executable. This function is meant to run a custom-action, and the conditional blocks just become a Boolean not an action.

src/custom-actions/src/ast_block.c
```diff
@@ -92,33 +92,36 @@ int AST_block_free(AST_block_t *ast)
...
+  int rc;
   switch(block->block_type)
     {
     case(CONTROL):
       return FAILURE;
       break;
     case(BRANCH):
-      returnV = do_branch_block(block->block->branch_block);
-      if(returnV == 0)
+      //To be implemented
```
MaxineK36, line 106: What part is to-be-implemented? If nothing has been implemented for BRANCH yet, it might make sense for this to return failure

MaxineK36 (commented):
One last clarification

src/custom-actions/src/ast_block.c
LapsededPacifist: Do_branch_block was what wasn't implemented. I had set it up, so that it would just skip over branch blocks. However, I updated it so that it would return failure for now probably makes more sense in the short term.

MaxineK36 (approved):
Looks great!

LapsededPacifist:
Functionality for running Custom-Actions with only action blocks has been merged.
Functions do_custom_action and run_ast_block were added for this function.  Currently, they fail when run with anything besides a list of AST_action_blocks. List functionality for making this list will be added in [issue 1069](https://github.com/uchicago-cs/chiventure/issues/1069) and the ability to branch will be added in [issue 1068](https://github.com/uchicago-cs/chiventure/issues/1068). 
To support the functionality of branch blocks, eval_conditional_block was added. It returns a boolean based off of comparing the attributes in the conditional with the specified operator. This currently has no implementation and required a slight restructuring of the conditional_type_t enum, so that there was a way to differentiate between less than and greater than when evaluating a conditional.  Required minor modifications to the tests in test_branch_block and test_conditional_block.

MaxineK36:
Issue Score: ✔️++

Comments:
Excellent job taking over this existing branch and merging it to dev! You did a particularly good job linking the issues relating to future work in your closing statement.

PR #1081, opened 2021-05-23
Implement functions for Ast_block_t List Functionality

### LINKED ISSUE #1069: Implement List Functionality for AST_Blocks
Currently AST_blocks have prev and next pointers to other AST_blocks in order to be used as lists.
Functionality for adding, removing, and search with a list of AST_blocks needs to be added.
Note for adding to the list: The order that the actions appear in the list is often very important, so add needs to be able to add to a specific place in the list.

jasulcaf:
(closes #1069)

This is connected to @LapsededPacifist and my Issue #1069 which set out to implement functions that allowed for list manipulation of AST_block_t structs. 

Overview of Implemented Changes:

- `list_contains_AST_block`: This function checks whether or not an AST_block_t is found in a linked list already or not.
- `list_add_AST_block`: This function adds an AST_block_t at a _specific_ place in the linked list (denoted by an integer that uses the same indexing as standard array indexing in C aka. 0 denotes adding as the new 1st element of linked list).
       - `append_list_AST_block`: Function for adding a new AST_block to the _end_ of a linked list
       - `prepend_list_AST_block`: Function for adding a new AST_block to the _start_ of a linked list
- `list_how_many_AST_block`: This function checks how long an AST_block_t linked list is
- `list_remove_AST_block`: This function removes a specific AST_bock_t from a linked list (if it is found).

Aside from adding this functionality, we also decided to get rid of the 'prev' field from AST_block_t struct. This was done because this is currently unused (and will remain like so) and so make the developer experience more intuitive.


adishpan:
Looks good to me.

adishpan (approved):
Looks good to me.

adishpan (approved):

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,188 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
+ *
+ * Returns:
+ *          - 2 possible int value: 0 for when AST_blocks are the same
+ *                                  1 otherwise
+ * Note: The value for same MUST be 0 for this helper to be successfully used 
+ *       in LL_<function>
+ */
+int AST_cmp(AST_block_t* AST1, AST_block_t* AST2)
+{
+    if (AST1 == AST2)
```
MaxineK36, line 120: This will only return 0 if the two inputs are literally the same pointer. Is this the intended behavior? Generally speaking, comparisons are only useful if they look for the programmer's definition of two structs being the same (e.g. containing the same values)

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,188 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
+        return 0;
+    else return 1;
+}
+
+/* See ast_block.h */
+int list_how_many_AST_block(AST_block_t* head)
```
MaxineK36, line 126: Use `LL_COUNT` instead of doing this by hand

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,188 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
+    if (head == NULL || compare == NULL)
+    {
+        return false;
+    }
+
+    AST_block_t* tmp;
+    
+    LL_SEARCH(head, tmp, compare, AST_cmp);
+
+    if (tmp)
```
MaxineK36, line 154: style: in chiventure overall we tend to use brackets regardless of if the `if` statement is more than one line or not; please maintain that convention (and at least remain consistent inside of a file)

MaxineK36: Also, for readability, do a comparison like `if (tmp != NULL)` even though what you have is valid

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,188 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
+bool list_contains_AST_block(AST_block_t* head, AST_block_t* compare)
+{
+    if (head == NULL || compare == NULL)
+    {
+        return false;
+    }
+
+    AST_block_t* tmp;
+    
+    LL_SEARCH(head, tmp, compare, AST_cmp);
```
MaxineK36, line 152: Based on how you've written your compare function, this will only return true if the block `compare` is literally in the list (and not if an equivalent block is there)

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,188 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
+
+    if (tmp)
+        return true;
+    else return false;
+}
+
+/* See ast_block.h */
+int list_add_AST_block(AST_block_t* head, AST_block_t* add, int num_to_place)
```
MaxineK36, line 160: Why would you need to add AST_blocks at a specific location? This kind of manual rearranging somewhat interferes with the usefulness of using `utlist`
If you *do* determine that this is still needed, could you use `LL_INSERT_INORDER`?

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,188 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
+
+    /* Move Pointers around to accomodate new AST_block */
+    add->next = curr;
+
+    prev->next = add;
+
+    return SUCCESS;
+}
+
+/* See ast_block.h */
```
MaxineK36, line 201: See above comment: could these two functions just use `LL_PREPEND` and `LL_APPEND`?

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,188 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
+  return list_add_AST_block(head, add, 0);
+}
+
+/* See ast_block.h */
+int list_remove_AST_block(AST_block_t* head, AST_block_t* del)
+{
+    assert(head != NULL);
+    assert(del != NULL);
+
+    if (list_contains_AST_block(head, del) == false)
```
MaxineK36, line 219: If `del` is the *pointer* that exists in the list and you're trying to remove, you can just use `LL_DELETE` directly. If not, this function will always return `FAILURE` due to your `cmp` function

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,188 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
+        return FAILURE;
+
+    AST_block_t *elt, *tmp;
+    LL_FOREACH_SAFE(head, elt, tmp)
+    {
+        if (AST_cmp(elt, del) == 0)
+        {
+            LL_DELETE(head, elt);
+            
+            /* Set 'next' pointer to NULL do avoid recursive freeing from AST_block_free */
```
MaxineK36, line 229: This shouldn't be necessary

tests/custom-actions/test_ast_block.c
```diff
@@ -136,7 +133,386 @@ Test(AST_block_t, init_CONDITIONAL)
...
+
+    AST_block_t* second_ast = AST_block_new(brnc, second_block_type);
+    cr_assert_not_null(second_ast, "AST_block_new failed to create the second AST_block");
+
+    block_t *act = malloc(sizeof(action_block_t));
+    block_type_t third_block_type = ACTION;
+     
+    AST_block_t* third_ast = AST_block_new(act, third_block_type);
+    cr_assert_not_null(third_ast, "AST_block_new failed to create the third AST_block");
+
```
MaxineK36, line 161: Test your append functions first, and then use those in these tests instead of linking manually

tests/custom-actions/test_ast_block.c
```diff
@@ -136,7 +133,386 @@ Test(AST_block_t, init_CONDITIONAL)
...
+    AST_block_t* new_ast = AST_block_new(block, block_type);
+    cr_assert_not_null(new_ast, "AST_block_new failed to create a AST_block");
+
+    block_t *brnc = malloc(sizeof(block_t));
+    block_type_t second_block_type = BRANCH;
+
+    AST_block_t* second_ast = AST_block_new(brnc, second_block_type);
+    cr_assert_not_null(second_ast, "AST_block_new failed to create the second AST_block");
+
+    new_ast->next = second_ast;
```
MaxineK36, line 192: Same note here (don't link manually)

tests/custom-actions/test_ast_block.c
```diff
@@ -136,7 +133,386 @@ Test(AST_block_t, init_CONDITIONAL)
...
+    
+    AST_block_t* compare = AST_block_new(com, com_type);
+    cr_assert_not_null(compare, "AST_block_new failed to create a AST_block");
+
+    /* Check whether or not CONTROL is found in linked list*/
+    ret_val = list_contains_AST_block(new_ast, compare);
+    cr_assert_eq(ret_val, false, "list_contains_AST_block mistakingly found an AST_block not yet added");
+
+    /* Check whether or not BRANCH is found in linked list */
+    ret_val = list_contains_AST_block(new_ast, second_ast);
```
MaxineK36, line 206: Note that this only works because you passed in the same *pointer* you added to the list

tests/custom-actions/test_ast_block.c
```diff
@@ -136,7 +133,386 @@ Test(AST_block_t, init_CONDITIONAL)
...
+    AST_block_t* second_ast = AST_block_new(brnc, second_block_type);
+    cr_assert_not_null(second_ast, "AST_block_new failed to create the second AST_block");
+
+    block_t *act = malloc(sizeof(action_block_t));
+    block_type_t third_block_type = ACTION;
+     
+    AST_block_t* third_ast = AST_block_new(act, third_block_type);
+    cr_assert_not_null(third_ast, "AST_block_new failed to create the third AST_block");
+
+    first_ast->next = second_ast;
```
MaxineK36, line 293: Again, please don't link manually

MaxineK36 (changes requested):
Made a couple of comments/asked clarification questions about your list operations; in general, you should be using the utlist macros, and not manually modifying `next`. Please also make sure that your search function behavior is what you intended.

src/custom-actions/src/ast_block.c
LapsededPacifist: This was the intended behavior because there isn't a key to search by. The only thing to search is the pointer, and the goal was to make sure exact duplicates were not added to the list. If something was added to the list twice, all functionality would break because it would become a loop.
We were initially trying to search by something different, but unlike other structs there aren't any defining features of AST_blocks to search by,

src/custom-actions/src/ast_block.c
LapsededPacifist: The reason you would need to add AST_blocks at a specific location would be because the order of actions in the list affects is the order that they are executed currently. The thought was that modifications to the custom action after it was created would need to be able to place new actions at specific points in the list.  I don't think that there are any plans for this kind of modification in the future, but it was something I thought would be likely to happen at some point.
I don't believe that LL_INSERT_INORDER could be used for this because (I think) it is just a compare function to put this in the correct location of the list. Our compare function, as pointed out earlier, wouldn't be very useful for this.

src/custom-actions/src/ast_block.c
LapsededPacifist: As mentioned before, this is intentional. The goal was to check if the that exact block was already in the list to avoid infinite loops.  In hindsight, the LL functions might have protections against that.

MaxineK36: In what situation do you imagine this coming up?

src/custom-actions/src/ast_block.c
MaxineK36: Given that custom actions (like everything else) will be read from the game developer's wdl specification, how would you imagine this coming up?

src/custom-actions/src/ast_block.c
LapsededPacifist: Well, it came up when Jose and I were writing tests, but yeah it will probably never come up when actually running chiventure.

MaxineK36: Ah, got it. Given that this isn't an expected issue, I think it would be reasonable to simply include a note in the function headers for the `add` functions stating that the function assumes that the user won't add duplicates to the list (as this could result in infinite loops)

src/custom-actions/src/ast_block.c
LapsededPacifist: They could specify changes to an existing actions at some trigger point. However, this could also be done by simply creating a new custom action at that point and removing the old one. 
I don't think there is anything in the near (or really distant) future, so yeah it might make more sense to remove this if it interferes with utlist.

src/custom-actions/src/ast_block.c
jasulcaf: LL_DELETE doesn't bother to change the pointers. So imagine if we have A->B->C and we LL_DELETE B. Then all that will do is set A->C, but we haven't changed the pointer to B yet. If we don't set next pointer to NULL then we get double free() errors.

src/custom-actions/include/ast_block.h
```diff
@@ -73,4 +72,48 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type);
  */
```
MaxineK36, line 72: It may be helpful to make this into two functions -- one that frees a single block, and one that frees a block and all the other blocks in the list.

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,98 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
-            {
-                conditional_block_free(ast->block->conditional_block);
-            }
-            break; 
-        case BRANCH:
-            if (ast->block->branch_block != NULL)
-            {
-                branch_block_free(ast->block->branch_block);
-            }
+        switch(ast->block_type)
```
MaxineK36, line 55: This switch statement would go in the normal free function

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,98 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
+                if (ast->block->branch_block != NULL)
+                {
+                    branch_block_free(ast->block->branch_block);
+                }
+                free(ast->block);
+                ast->block == NULL;
+        }
     }
+
     if (ast->next != NULL)
```
MaxineK36, line 93: This would go in the list free function, and would also call LL_DELETE

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,98 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
 }
+
+//---------------To add List Functionality for AST_blocks----------------------
+
+/* See ast_block.h */
+int list_how_many_AST_block(AST_block_t* head)
+{
+    assert(head != NULL);
+
+    int ret_val;
```
MaxineK36, line 110: style: call this `count` or something similar but more descriptive than `ret_val`

src/custom-actions/src/ast_block.c
```diff
@@ -50,41 +49,98 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
+/* See ast_block.h */
+int list_remove_AST_block(AST_block_t* head, AST_block_t* del)
+{
+    assert(head != NULL);
+    assert(del != NULL);
+    
+    LL_DELETE(head, del);
+
+    /* Set 'next' pointer to NULL do avoid recursive freeing from AST_block_free */
+    del->next = NULL;
```
MaxineK36, line 142: If the changes are made to the `free` functions as suggested above, this becomes unnecessary

MaxineK36 (changes requested):
A few more fixes to work better with utlist

src/custom-actions/include/ast_block.h
jasulcaf: That seems like a good abstraction idea, will implement this as `AST_free` freeing the resources for _one_ block, and `AST_block_free` will free _all_ blocks

src/custom-actions/src/ast_block.c
jasulcaf: This probably slipped by when we were busy writing tests, changed variable name to `count`!

src/custom-actions/src/ast_block.c
jasulcaf: Yes. This was no longer necessary with the abstraction of free functions

src/custom-actions/src/ast_block.c
```diff
@@ -41,50 +41,116 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
```
MaxineK36, line 152: Please now go ahead and remove this line

src/custom-actions/src/ast_block.c
```diff
@@ -41,50 +41,116 @@ int AST_block_init(AST_block_t *ast, block_t *block, block_type_t block_type)
...
+int AST_block_free(AST_block_t *ast)
+{
+    assert(ast != NULL);
+
+    AST_block_t *elt, *tmp;
+    LL_FOREACH_SAFE(ast, elt, tmp)
     {
-        AST_block_free(ast->prev);
+        LL_DELETE(ast, elt);
+	    AST_free(elt);
```
MaxineK36, line 107: nit: line 107 seems over-indented

MaxineK36 (changes requested):
2 tiny changes!

MaxineK36 (approved):
Looks great!

jasulcaf:
As a summary:

This Pull Request provides list functionality to exist for the AST_block_t struct. Through the process of the Pull Request Reviews however, we decided to eliminate some of the functions we were planning to add. As such, the functions that _have_ been added are the following:

- `append_list_AST_block` : Function for adding a new AST_block to the end of a linked list
- `prepend_list_AST_block` : Function for adding a new AST_block to the start of a linked list
- `list_how_many_AST_block` : Function that checks how long an AST_block_t linked list is
- `list_remove_AST_block` : Function to remove a specific AST_block_t from a linked list (if it is found). Note, this function _does_ free the removed AST_block_t

To be explicit on what functions didn’t go through (and why):

- `list_contains_AST_block:` At this moment in time we don’t have a specific way of identifying an AST_block_t so this function wouldn’t be too useful (would just be checking if a specific pointer is in the list)
- `list_add_AST_block:` The functionality of being able to specify in what spot of the linked list a new AST_block_t should be added to isn’t required as of this Pull Request. If in the future the order of AST_blocks does play an integral role in Chiventure, then there will be a case for adding this function again

Also, through the review phase, we decided to abstract the way that `AST_block_free` functioned. Before, we could only free the _entire_ list of blocks. However, we have added `AST_free` which will only free a _singular_ block. This abstraction helps utilize LL functions and helps with making code cleaner. `AST_block_free` functionality was not changed (still frees the _entire_ list)

MaxineK36:
Issue Score: ✔️++

Comments:
Excellent work! Your opening and closing summaries are particularly well-written and makes it easy for readers to understand the intent and outcome of this PR.

PR #1083, opened 2021-05-23
Added the ability to load NPCs from WDL files; added multiple hash table support for items

### LINKED ISSUE #1055: Create a load_npc module to load NPCs from a wdl_ctx struct
Currently, Chiventure is able to load rooms and items from a wdl_ctx struct. By extending this to NPCs, we can add NPCs to the game from a WDL file.

### LINKED ISSUE #1116: Fix a bug where trying to add an item to two hash tables causes errors
Trying to add an item to two hash tables causes a bug because it requires two hash handles, when the item struct currently only has 1.

In order to fix this issue, I will add a second hash handle to the item struct and modify related functions accordingly.

jeyeong:
In preparation for our team's game demo, I have made the following additions:

**Loading NPCs from WDL files** `wdl/load_npc.c`
- NPCs can now be loaded into Chiventure using the following format:
image
- This creates the NPC struct and adds the NPC to the room indicated by "in"

**Loading basic dialogue, inventories from WDL files** `wdl/load_npc.c`
- Dialogue and inventories can be loaded in the following manner:
image

**Conditional dialogue & Actions** `wdl/load_npc.c` `wdl/load_condition.c`
- Both can now be loaded into the game in the following manner:
image

**Multiple hash table support for items** `game-state/game.c` `game-state/item.c`
- Since while implementing the loading of NPCs from WDL files I regularly needed to get items from game->all_items, I thought I would try to introduce multiple hash table support for items.
- Items can now belong to two hash tables. This was achieved by adding a second hash handle.
- I did not change the interfaces of add_item_to_game(), add_item_to_room(), add_item_to_player(), etc., so no prior code should be broken.

Closes #1055, Closes #1116


jeyeong:
I would like to put this pull request on pause for now.

Given the merging of conditional dialogue + actions into the game today, I would like to make some changes to the load_npc module.

src/game-state/src/item.c
```diff
@@ -608,7 +688,6 @@ int delete_all_items(item_hash_t** items)
             iter = current_item->next;
 
             remove_item_from_hash(items, current_item); /* deletes (items advances to next) */
-            item_free(current_item);             /* free it */
```
jeyeong, line 611: Since items now belong to two hash tables, namely, game->all_items and player->inventory (for example), the freeing process is as follows:
1. Items are removed from the player->inventory, npc->inventory, room->items hash tables, but **not freed**.
2. Items are then removed from the game->all_items hash table, and freed at the same time.

jeyeong:
I've completed the implementation. The summary above has been updated.

src/game-state/src/item.c
```diff
@@ -177,6 +177,86 @@ int remove_item_from_hash(item_hash_t **ht, item_t *old_item)
     return SUCCESS;
 }
 
+/* See item.h */
+int add_item_to_all_items_hash(item_hash_t **all_items, item_t *item)
```
pkasner, line 181: Is there a reason that this takes an item_hash_t** and not just an item_hash_t*? Or is that just a carryover from previous implementations of similar functions? Just wondering if you figured out why that was

src/game-state/src/item.c
pkasner: If the delete_all_items() function is only removing all items from the hash table without freeing them, I believe the HASH_CLEAR macro could be used to simplify that function significantly.

src/npc/src/dialogue.c
```diff
@@ -133,7 +134,14 @@ int do_node_actions(node_t *n, game_t *game)
...
 
             case GIVE_ITEM: ;
-                npc_t *npc = get_npc(game, game->mode->mode_ctx);
+                // Ideally, this function would look like:
+                //   npc_t *npc = get_npc(game, game->mode->mode_ctx);
+                // That is, we get the NPC struct from game->all_items rather
+                // than game->curr_room->npcs->npc_list. However, as NPC
+                // structs cannot currently belong to two hash tables, we will
+                // have to use this workaround.
+                npc_t *npc = get_npc_in_room(game->curr_room,
```
pkasner, line 143: You say that this would be better if using the get_npc function, but I think using get_npc_in_room makes perfect sense here. In conversation mode, the mode_ctx is necessarily assigned to an npc in the current room via the talk_operation, so the two functions will act equivalently here.

src/npc/src/dialogue.c
```diff
@@ -143,7 +151,10 @@ int do_node_actions(node_t *n, game_t *game)
                 break;
 
             case TAKE_ITEM:
-                npc = get_npc(game, game->mode->mode_ctx);
+                // Ideally, this function would look like:
+                //   npc = get_npc(game, game->mode->mode_ctx);
+                // See GIVE_ITEM
+                npc = get_npc_in_room(game->curr_room, game->mode->mode_ctx);
```
pkasner, line 157: Again, I think using get_npc_in_room seems totally fine here.

src/npc/src/dialogue.c
```diff
@@ -337,19 +348,27 @@ char *run_conversation_step(convo_t *c, int input, int *rc, game_t *game)
 
     c->cur_node = cur_edge->edge->to;
 
-    // Step 2: Execute actions (item, quest, battle, etc.), if any
+    // Step 2: Disable edge if the node it leads to has an action
+    // NOTE: This is a temporary solution that prevents issues like being able
+    //       to receive multiple copies of items, starting the same quest twice.
+    //       This SHOULD be changed / made more complex in the future.
+    if (c->cur_node->actions != NULL) {
+        cur_edge->availability = EDGE_DISABLED;
```
pkasner, line 356: This temporary solution makes a lot of sense to prevent node actions from being taken multiple times. Because currently if node actions fails, for example because the item the NPC was supposed to give is no longer in their inventory, the run_conversation_step function gives a return code of -1, which will cause a crash via assertion. Perhaps a more graceful solution would involve some manner of specifying if an action failed without treating that case as a full blown error, but for now I agree that this solution is sufficient to prevent those action failures.

src/wdl/src/load_game.c
```diff
@@ -15,16 +16,20 @@ game_t *load_game(obj_t *obj_store)
     int rc;
     game_t *game = create_game(obj_store);
 
-    // call functions that parse items, actions, rooms, and game attributes
-    // into a game pointer
+    // player
+    player_t *p = player_new("player");
+    add_player_to_game(game, p);
+    set_curr_player(game, p);
```
pkasner, line 22: Why was this added? I believe this is done in chiventure_ctx_init().

src/wdl/src/load_npc.c
```diff
@@ -0,0 +1,267 @@
...
+                fprintf(stderr, "Dialogue was not loaded properly. NPC: %s\n",
+                        id);
+                return FAILURE;
+            }
+        }
+
+        // add NPC to the game
+        // NOTE: this is commented out because the NPC struct cannot currently
+        //       support belonging to two hash tables
+        // add_npc_to_game(g, npc);
```
pkasner, line 253: Perhaps it is best to wait until the talk_operation can be merged first (which allows npcs to be stored in a second hash table) before this PR. That way all the commented out code can be cleaned up.

pkasner (changes requested):
Most of my comments shouldn't require changes for this PR, and overall this looks really good. I think it is best to wait for the talk_operation PR to be merged, so that the commented code relying on second hashtable support can be implemented.

src/game-state/src/item.c
jeyeong: I saw that you used that in your implementation of multiple hash handles for NPCs, and it was brilliant! In the case of items, however, I don't think we can use HASH_CLEAR, because items are implemented to also be linked lists (to accommodate multiple items with same ID), so there is a more complex hash deletion process.

src/game-state/src/item.c
jeyeong: The function, for some reason, doesn't seem to work properly if we pass an item_hash_t*. I tried using item_hash_t* at first, but I encountered issues adding the item to the hash table correctly. After changing it to item_hash_t**, it worked fine. I have no idea why that is.

src/wdl/src/load_npc.c
jeyeong: Agreed

src/wdl/src/load_game.c
jeyeong: Oh is it? I did not realize that. That actually is a problem, though. In order for conditions (`inventory_condition_new()`) to be created, they require the player to be passed as a parameter. But since `chiventure_ctx_init()` happens after the condition creation process, that will be a problem.

src/game-state/src/item.c
pkasner: Correct me if I'm wrong, but as I understand it, the reason it's more complicated is because if you delete the head, then you need to add the next element of the list to the hash table. And if you delete an item that's not the head, you just splice it out of the list. And if it's just one element you delete it normally. It seems to me then that you could still use HASH_CLEAR, and the only difference is that the items would remain "linked" together, and we still delete all the buckets. I don't think this interferes with their freeing via delete_all_items_from_game because that function still does go through and free each element in every linked list.

lilyehsani (changes requested):
Great job overall! Added a few comments. I also did not see anything about loading NPC struct fields like class, movement, or battle (which makes sense as these are new or complex additions). However, it might be good to add some comments about how we would like to support loading these fields in the future.

src/game-state/src/item.c
jeyeong: You know what, you're right. It actually makes perfect sense to use HASH_CLEAR. I've updated the code to use it.

src/game-state/src/item.c
jeyeong: Somehow, the item wasn't being added to the hash table. I'm guessing that's the reason why the previous years' team decided to use item_hash_t ** too

include/game-state/game.h
```diff
@@ -5,6 +5,7 @@
 #include "player.h"
 #include "room.h"
 #include "item.h"
+#include "npc/npc.h"
```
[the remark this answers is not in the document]
jeyeong: Done

include/wdl/load_condition.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * Header file for loading condition objects into game struct
+ */
+
+#ifndef INCLUDE_LOAD_CONDITION_H
+#define INCLUDE_LOAD_CONDITION_H
+
+#include <stdio.h>
```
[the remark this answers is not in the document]
jeyeong: Done

include/wdl/load_npc.h
```diff
@@ -0,0 +1,35 @@
+/*
+ * Header file for loading NPC objects into game struct
+ */
+
+#ifndef INCLUDE_LOAD_NPC_H
+#define INCLUDE_LOAD_NPC_H
+
+#include <stdio.h>
```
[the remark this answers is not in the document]
jeyeong: Done

src/game-state/src/game.c
```diff
@@ -272,14 +269,43 @@ int move_room(game_t *game, room_t *new_room)
...
+ *
+ * Returns:
+ *  SUCCESS if successful, FAILURE if failed
+ */
+int delete_all_items_from_game(item_hash_t* all_items)
+{
+    item_t *current_item, *tmp;
+    HASH_ITER(hh_all_items, all_items, current_item, tmp)
+    {
+        item_t* iter = current_item;
```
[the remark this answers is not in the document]
jeyeong: Done

include/wdl/load_npc.h
```diff
@@ -0,0 +1,35 @@
...
+/* load_npcs
+ * accesses the document object and obtain list of NPCs to parse into the
+ * game struct
+ * 
+ * parameters:
+ * - obj: the document object
+ * - g: a game pointer as a return parameter
+ *
+ * side effects:
+ * - adds NPCs into the inputted game pointer (currently not possible) and
```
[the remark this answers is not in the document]
jeyeong: Revised

src/wdl/src/load_game.c
jeyeong: So conditions (and by extension conditional dialogue) cannot be created unless a player struct has been initialized. Thus, in order for us to develop conditional dialogue, we require the player struct to be created prior to dialogue loading. Right now, however, the player struct is created after the game has been loaded, since in `chiventure.c`, `chiventure_ctx_new` is called after `load_game`.

I imagine this would result in issues for other teams too. For example, if we want to assign default classes to players, or if we want to create conditions for quests, we would need the player struct to be initialized first.

A simple solution to this would simply be to move these lines of code:
```
player_t *player1 = player_new("player1");
add_player_to_game(game, player1);
set_curr_player(game, player1);
```
from chiventure_ctx_init (in ctx.c) to game_new (in game.c). Since every call to chiventure_ctx_new (in all the game demos) is preceded by game_new, this should not result in any previous code being broken. It would, however, allow us (and future teams) to work with conditions in Chiventure.

jeyeong: I've just tested it out, and the Chiventure (together with the demos) compiles fine with the change. Furthermore, conditional dialogue works great. If this is okay with you, I'll make the change.

jeyeong: To clarify, I will also remove the three lines above (20-22)

jeyeong: Hey Lily, I've gone ahead and made the change so that we can start working on our game demo tomorrow! If this is not ideal for any reason, please let me know, and I will change it back to what it previously was.

pkasner (approved):
This LGTM! There was a small section in test_dialogue.c that hadn't been updated for npcs in two hash tables, so I went in fixed that.

src/wdl/src/load_game.c
lilyehsani: I think this is fine.

jeyeong:
I edited the HELP command to include "talk to" as a possible command

src/npc/src/dialogue.c
lilyehsani: I agree with Peter, so you should probably delete this in-code comment and the following one.

src/wdl/src/load_npc.c
lilyehsani: Talk operation has been merged, so you are now able to make these changes.

lilyehsani (changes requested):
Great changes! Just a few more, especially now that talk operation has been merged.

src/wdl/src/load_npc.c
jeyeong: Done!

src/npc/src/dialogue.c
jeyeong: Done!

lilyehsani (approved):
LGTM! Approved.

jeyeong:
**Summary of changes made:**

- Introduced the ability to load NPCs from WDL files. The format for this is specified in images above.
- Modified the items struct, and associated functions, such that items can now belong to two hash tables. Practically, the first hash table an item will belong to is `game->items`, and the second `player->inventory`, etc.

**Next steps:**

- Design our team's game demo.

lilyehsani:
Issue Score: ✔️++

Comments:
Great work overall! The summary and closing of this PR are very thorough.

Issue #1026, opened 2021-05-18
Write tests for npc/battle integration


amymorrill:
Over the past couple of sprints (and during this sprint) we've created many functions to help integrate the npc and battle modules. For more information about these functions see #883, #992, and #1025. 

Now we need to implement both unit tests for our functions and integration tests between npc and battle. Note that the integration tests will rely on the battle team incorporating our functions into their module, which they are working on this sprint. We will discuss this with them and may defer integration testing to them or leave it to the backlog if we run out of time.

amymorrill:
I've written tests for the npc_battle struct's new, init, and free functions. The next step will be to write tests for the additional functions outlined in #1025.

northofsophia:
I have added multiple tests for the transfer_all_npc_items() function that will help in looting dead bodies. I will now issue a pull request to push these tests to dev.

lilyehsani:
Issue Score: ✔️++

Comments:
Great work overall! The summary of this issue is very thorough.

PR #1090, opened 2021-05-24
New npc_battle functions

amymorrill:
After creating the npc_battle struct and its basic functions (new/init/free) in a previous pull request, I now have implemented more functions relating to npc_battle. Hopefully, these functions will help other developers who want to use the npc_battle field in the npc struct. 

Here are a list of the changes in this pull request:
- Created a check_npc_battle() function which takes an npc and returns true if its will_fight and npc_battle fields are compatible
- Created a get_npc_battle() function which takes an npc and returns a pointer to its npc_battle struct or NULL if that npc doesn't have an npc_battle struct
- Moved the get_npc_health() function to be under the get_npc_battle function because health is in npc_battle now, so this order makes sense
- Created an add_battle_to_npc() function which takes an npc and the parameters of an npc_battle struct and adds an npc_battle struct with those parameters to the npc
- Moved change_npc_health() to be under add_battle_to_npc() because health is in npc_battle now, so this order makes sense


src/npc/src/npc.c
```diff
@@ -172,6 +180,39 @@ int add_convo_to_npc(npc_t *npc, convo_t *c)
...
+/* See npc.h */
+int change_npc_health(npc_t *npc, int change, int max)
+{
+    assert(npc->npc_battle != NULL);
+
+    if ((npc->npc_battle->health + change) < 0)
+    {
+        npc->npc_battle->health = 0;
+    }
+    if ((npc->npc_battle->health + change) < max)
```
northofsophia, line 205: This line should be an else if instead of an if or else the previous if statement "if ((npc->npc_battle->health + change) < 0)" will get overwritten.

northofsophia (changes requested):
Everything looks good except for line 205 in change_npc_health() where the if statement should be changed to an else if. Other than that, the functions look clean.

src/npc/src/npc.c
amymorrill: Whoops, good catch, I'll change that!

northofsophia (approved):
All the functions look good.

include/npc/npc.h
```diff
@@ -221,6 +231,42 @@ int remove_item_from_npc(npc_t *npc, item_t *item);
...
+ *  change: the positive or negative change to be made to the health points
+ *  max: the maximum health the npc can achieve
+ *
+ *  The change has a minimum value of 0
+ *  The change has a maximum value of max
+ *
+ * Returns:
+ *  the npc's new health
+ */
+int change_npc_health(npc_t *npc, int change, int max);
```
lilyehsani, line 270: I know we talked about moving these to npc_battle.h, but I actually think they make the most sense here.

src/npc/src/npc.c
```diff
@@ -121,27 +129,28 @@ item_list_t *get_npc_inv_list(npc_t *npc)
...
-        npc->npc_battle->health = max;
+    return npc->npc_battle;
+}
+
+/* See npc.h */
+int get_npc_health(npc_t *npc)
+{
+    assert(npc != NULL);
+
+    if (npc->npc_battle == NULL) {
```
lilyehsani, line 145: Be consistent with placement of curly braces. You mostly use Allman style, see: https://uchicago-cs.github.io/dev-guide/style_guide_c.html.

src/npc/src/npc.c
```diff
@@ -73,6 +73,20 @@ int npc_free(npc_t *npc)
...
 }
 
+// "CHECK" FUNCTIONS ----------------------------------------------------------
+
+/* See npc.h */
+bool check_npc_battle(npc_t *npc)
+{
+    assert(npc != NULL);
+
+    if (npc->will_fight == true && npc->npc_battle == NULL) {
```
lilyehsani, line 83: Be consistent with placement of curly braces. You mostly use Allman style, see: https://uchicago-cs.github.io/dev-guide/style_guide_c.html.

lilyehsani (changes requested):
Great job! Added a few comments.

include/npc/npc.h
amymorrill: Okay, I'll leave them as is.

lilyehsani (approved):
Looks great! Approved.

amymorrill:
Closing statement: This pull request has now been merged, with no major changes other than those described in its opening statement. The next step will be to write tests for each of these new functions, as described in #1026.

lilyehsani:
Issue Score: ✔️+

Comments:
Great work overall! I have one comment:
- The title of a PR has to convey what the PR will be producing. In the future, try using simple titles with a "Verb-Object" structure that clearly states what you will be doing in the issue (e.g., “Design and implement a ‘circle’ module”, “Research libraries for drawing simple geometric shapes”, etc.)

PR #1091, opened 2021-05-24
Removing an item from player struct once used, and having an item count as the player's turn

lczielinski:
I created a function, remove_battle_item(), which removes an item from the doubly linked list of battle_items from the player struct. It is called after the player uses a battle_item, and I modified the function use_battle_item() to check if the item's quantity is 0 and if so, it removes the item from the list. This ensures that items with a quantity of 0 do not stay in the player's list of items. I also created tests for the function.


lczielinski:
I have also added changes to allow a player's use of an item to count as their move (e.g. after the player uses an item, the enemy makes a move). This involved splitting part the battle_flow() function into a new function, enemy_make_move(), which handles the enemy choosing and using a move. Now, enemy_make_move() and corresponding print functions are used in battle_command_example.c after an item is used.

jintae412 (approved):
Seems like all of the code should work quite well. We'll probably need to mimic what you've done in the battle_command_example with the actual battle_flow function.

firatciftci (approved):
Everything looks good to me!

firatciftci:
Issue Score: ✔️++

Comments:
Great work!

PR #1093, opened 2021-05-24
Loot dead npc bodies

northofsophia:
After discussing with the other NPC team of how to approach the looting of dead npc bodies after a won battle, we came to the conclusion to transfer all npc items to the room struct so that the player can take items from the room instead of directly off the npc. This pull request includes the function transfer_all_npc_items() that deletes all items from the npc inventory and adds them to the room struct. It returns SUCCESS if the operation is successful.


amymorrill (approved):
Looks good to me!

lilyehsani (approved):
Looks great! Approved.

lilyehsani:
Issue Score: ✔️

Comments:
Great work overall! I have two comments:
- The title of a PR has to convey what the PR will be producing. In the future, try using simple titles with a "Verb-Object" structure that clearly states what you will be doing in the PR (e.g., “Design and implement a ‘circle’ module”, “Research libraries for drawing simple geometric shapes”, etc.)
- Like issues, PRs require closing comments. When an PR is done, you should include a final comment summarizing what was merged to `dev` and what work remains to be done, if any.

Issue #998, opened 2021-05-13
Add variables to grammar


r-e-stern:
We'll work on assigning variables to the DSL grammar (per the DSL specifications on the wiki) such that users can declare and reuse a variable within a DSL world description file.

jacknugent1529:
Right now we have a simple implementation of variables, where we can define a variable as a string and reuse it in different locations. Here is an example of how to define variables
```
$Room = """
ROOM room
    short desc: "this is a room"
"""
```
Note that the name of the variable `room` in this case must be a is similar in syntax to a variable name in c (or any other common language). It must be a single word and can include numbers. To separate between multiple words, we recommend using snake case when the variable represents simply a string such as a description or a property-value pair. We recommend using CamelCase for a variable represents a room, item, or any other compound objects which contains its own properties.

Additionally, the syntax has been changed to triple quotes (from single quotes). This allows us to place single quotes inside the variable declaration, which is necessary for more complicated structures.

We still need to add the ability to substitute strings into variables such as
```
$greeting = """hello {name}"""
```

jacknugent1529:
Now we also support simple escaped strings as variables. For example instead of using triple quotes, we can write
```
$intro = "introduction to the game"
```

jacknugent1529:
Variables are now fully implemented including support for template strings (similar to python f-strings). In summary, there remain two types of variable declarations:
```
$var1 = """
multi-line
replacement
"""

$var2 = "escaped string replacement"
```
Both of these variables can be referenced with the `$varname` syntax. For the first type, the text within the quotes is inserted directly into the document, so the reference to `$var1` evaluates to
```
multi-line
replacement
```
Notice that this does not include quotes. Therefore, this variable type is better to create objects such as items or rooms which can be reused. 

The second declaration evaluates to an escaped string *surrounded by quotes*. Specifically, `$var2` would evaluate to
```
"escaped string replacement"
```

We have just added support for template strings, where for either type of declaration, we can include unspecified inner variables such as
```
$intro = "Hello {name}"
```
Then to reference this variable, we would write
```
GAME ...
    intro: $intro { name: "John" }
```
This would evaluate to
```
GAME ...
    intro: "Hello John"
```
Note that we don't need to include the quotes around "John", and we could instead reference the variable as `$intro { name: John}`. However, the quotes would be necessary if we wanted to include escaped characters in one of the replacements.

[...]

PR #1095, opened 2021-05-24
Implement Variables in the DSL

jacknugent1529:
Variables are implemented as a pre-parsing addition to the language, so they are evaluated before the language itself is parsed. This means that variables will not conflict with the grammar of the language. The exact specifications are described in issue #998, but the basic features are the following
- multi-line variables declared with triple quotes (e.g. `$var = """<content>"""`
- escaped-string variables declared with single quotes (e.g. `$var = "This will evaluate to a string surrounded by quotes"`)
- Support for template strings similar to python f-strings

The main files added in this pull request are `vars_parser.py` and `vars.lark` which evaluate the  variables in the dsl file. Additionally, there is an example file using variables in several different manners, and the `vars_test.py` file for testing purposes. This pull request will close issue #998.


john-hahn (commented):
The changes made to the files all pertain to the goal of implementing variables in the DSL. First off, the `src/dsl/src/vars.lark` file which sets the rules for the variables, nicely and efficiently denotes what is recognized as triple quotes, escaped strings, key values. `src/dsl/src/vars_parser.py` takes the variables and manipulates it using the functions defined. They essentially make the tree more readable, and Jack has done a nice job with the coding of these functions. Lastly, there is a file for testing the variables out. The `src/dsl/src/dsl_parser.py` file has minimal change, which is the inclusion of the `evalVars` library.

Overall, all these changes are necessary for the implementation of variables, and there isn't any problem I see with the code.

john-hahn (approved):

src/dsl/examples/vars.dsl
```diff
@@ -0,0 +1,25 @@
+# Define variables
+$intro = "hello {firstname} {lastname}"
+$fork = """
+ITEM Fork
+    long desc: "a fork. theres \"nothing\" remarkably special about it."
+"""
+
+
+GAME START Room A END room C
+  intro: $intro {firstname: John, lastname: "Smith"}
```
ecoble, line 10: Is it intentional that Smith is in quotations and John is not? If so, it might be good to leave a note pointing out that you are showing the flexibility of DSL

src/dsl/src/vars.lark
```diff
@@ -0,0 +1,40 @@
+// ?start: assignment
```
ecoble, line 1: This comment looks like it should be removed, or turned into prose

src/dsl/src/vars_test.py
```diff
@@ -0,0 +1,17 @@
+# simple script to test the functionality of the evalVars function
```
ecoble, line 1: This and any other tests file should go in their own tests directory, for now I think you can just nest the tests directory within your src directory, I will check about whether you should put it in the main tests directory since you are using python.

ecoble (changes requested):
This looks great! Just a few small things.

src/dsl/examples/vars.dsl
jacknugent1529: Yes, it is intentional. The quotes are only necessary if escape characters are included. I included a comment to that effect, and it will be mentioned in the documentation.

src/dsl/src/vars.lark
jacknugent1529: Yeah, it was not necessary, so I removed it

src/dsl/src/vars_test.py
jacknugent1529: This script only printed out the file after the variables had been replaced by the appropriate strings. Since it wasn't a true test file, I instead included the functionality in the `dsl_parser.py` file with the `--debug` flag.

ecoble (approved):

jacknugent1529:
Closing statement: We added support for variables in the dsl. This was implemented as a pre-parsing transformation that replaces references to variables with the variables itself. The main features involve escaped string variables, multi-line variables, and template strings similar to python f-strings. Since opening the pull request, we simplified debugging functionality to use the `--debug` flag with the `dsl_parser.py` instead of needing another script. Since these changes have been successfully merged, this closes issue #998.

ecoble:
Issue Score: ✔️++

Comments:
Great work adding variables to the DSL, and documenting progress on this PR!

PR #1101, opened 2021-05-25
Regression

m-ghosal:
Implemented leveling algorithm and implemented tests for the leveling algorithm.


tests/skilltrees/skill_test.c
```diff
@@ -76,35 +76,33 @@ Test(skill_tests, skill_execute_test)
...
 /* Tests skill_level_up when return value is 0 */
-
-Test(skill_tests, skill_level_up_0) {
-    skill_t* skill = skill_new(1000, ACTIVE, "defuse bomb", "defuses a bomb",
-      2, 5, effect_defuse_bomb);
-    check_level_up(skill, 0);
+Test(skill_tests, skill_level_up_0)
+{
+  skill_t* skill = skill_new(1000, ACTIVE, "defuse bomb", "defuses a bomb",
+    2, 5, effect_defuse_bomb);
```
Joshuashou, line 90: are these the same change?

rohangupta2806 (approved):
I went through the changes and they seem to make sense.  I don't think anything needs to be changed.  Approved.

tests/skilltrees/skill_test.c
m-ghosal: no this was what i added

firatciftci (approved):
Everything looks good to me!

firatciftci:
Issue Score: ✔️-

Comments:
Good work overall! I have a few things to comment on:
- The title of a PR has to convey what the issue will be producing. In the future, try using simple titles with a "Verb-Object" structure that clearly states what you was done in the PR (e.g., “Design and implementation of ‘circle’ module”, etc.).
- Even though you can add labels to PRs to categorize them, you should avoid titles that could plausibly apply to multiple teams.
- The summary for this PR is too terse. A good rule of thumb is that you should be writing at least one full paragraph in your summary (see https://uchicago-cs.github.io/cmsc22000/project/policies.html#tasks-issues for examples).
- When an PR is done, you should include a final comment summarizing what was merged to `dev` and what work remains to be done, if any.

PR #1102, opened 2021-05-25
Restructing of branch_block_t and implementation of do_branch_block 

LapsededPacifist:
Restructures branch blocks to the format seen on the [wiki](https://github.com/uchicago-cs/chiventure/wiki/Custom-Actions-~-Current-State-and-Future-Plan-of-Custom-Actions).
Restructures all the related functions and their corresponding tests.
- Resulted in the deletion of 3 quarters of the existing tests as they were not relevant (Testing for all the different enum values)
Added do_branch_block to branch_block.c/.h
- Executes a branch block for if and while loops
- - For loops are executed the same way as while loops. Creating a separate issue for easier creation of for loops.
Added do_branch_block to the implementation of run_ast_block, so that ast_blocks can now run branch_blocks.


src/custom-actions/include/branch_block.h
```diff
@@ -86,4 +86,15 @@ int branch_block_init(branch_block_t *branch, int num_conditionals, conditional_
...
 int branch_block_free(branch_block_t *branch);
 
+/* Given an branch block and its corresponding arguments,
+ * attempt to execute the given block.
+ *
+ * Parameters:
+ * - block: A pointer to the branch block to be executed
+ *
+ * Returns:
+ * - Integer of if the block being run was successful
```
jasulcaf, line 96: Usually I have seen in the rest of the Chiventure code that people are explicit on the return value. Usually something like the following: "SUCCESS if success, FAILURE if error". I think it'd be good to maintain this trend to make sure that the style of the code is consistent.

src/custom-actions/src/ast_block.c
```diff
@@ -96,20 +96,22 @@ int run_ast_block(AST_block_t *block)
...
-  int rc;
   switch(block->block_type)
     {
     case(CONTROL):
       return FAILURE;
       break;
     case(BRANCH):
-      //To be implemented
-      return FAILURE;
+        if (do_branch_block(block->block->branch_block) == FAILURE)
```
jasulcaf, line 105: Nit: Indentation of if() seems odd (as if it has 2 extra spaces).

src/custom-actions/src/branch_block.c
```diff
@@ -105,15 +105,76 @@ int branch_block_free(branch_block_t *branch)
...
     
     return SUCCESS; 
 } 
+
+/* See branch_block.h */
+int do_branch_block(branch_block_t *block)
+{
+    if (block == NULL)
+    {
+        return SUCCESS;
```
jasulcaf, line 126: I had a question about the decision of returning SUCCESS if block == NULL. This seems like it can give the impression that a branch_block was successfully ran when in fact perhaps nothing happened. Below you return FAILURE for when it's not a valid branch (if either num_conditionals or num_actions <= 0), so I was wondering about how you made the decision to make this return SUCCESS instead of FAILURE if we pass in a NULL branch_block_t

src/custom-actions/src/branch_block.c
```diff
@@ -105,15 +105,76 @@ int branch_block_free(branch_block_t *branch)
...
+	    {
+	        return FAILURE; //All conditions must have actions
+	    }
+	    for (int i = 0; i < block->num_conditionals; i++)
+	    {
+	        if (eval_conditional_block(block->conditionals[i]) == 0)
+                {
+		    return run_ast_block(block->actions[i]);
+                }
+            }
```
jasulcaf, line 145: Nit: The indentation of the closing "}" for the for() loop seems to have an extra space

src/custom-actions/src/branch_block.c
```diff
@@ -105,15 +105,76 @@ int branch_block_free(branch_block_t *branch)
...
+	    if(block->num_conditionals > 2 || block->num_actions != 1)
+	    {
+	        return FAILURE; //Invalid while syntax
+	    }
+	    conditional_block_t *loop = block->conditionals[0];
+	    if (block->num_conditionals == 2)
+	    {
+	      /*
+               * If the condition to end the loop is different than starting the loop
+	       */
```
jasulcaf, line 162: Nit: This could be just a single line comment instead of 3

jasulcaf (changes requested):
Aside from Nits about spacing, there is one question that I had about `do_branch_block` and the decision to return SUCCESS if block == NULL. Aside from this, looks good and great job on the tests, especially on collapsing some of the tests.

src/custom-actions/src/branch_block.c
LapsededPacifist: I changed it. It looked fine in emacs, but I deleted the spacing and readded it.

jasulcaf (approved):
Spacing was corrected and comment about `do_branch_block` was resolved by changing the return to FAILURE if passed in a NULL pointer, so looks good!

src/custom-actions/src/branch_block.c
```diff
@@ -105,15 +105,74 @@ int branch_block_free(branch_block_t *branch)
...
+    switch (block->control_type)
+    {
+        case IFELSE:
+            if (block->num_conditionals > block->num_actions || block->num_conditionals + 1 < block->num_actions)
+            {
+                return FAILURE; //All conditions must have actions
+            }
+            for (int i = 0; i < block->num_conditionals; i++)
+            {
+                if (eval_conditional_block(block->conditionals[i]) == 0)
```
MaxineK36, line 141: Can you add a comment explaining what's happening here?

src/custom-actions/src/branch_block.c
```diff
@@ -105,15 +105,74 @@ int branch_block_free(branch_block_t *branch)
...
+                {
+                    return run_ast_block(block->actions[i]);
+                }
+            }
+            if (block->num_actions > block->num_conditionals)
```
MaxineK36, line 146: Here as well please

src/custom-actions/src/branch_block.c
```diff
@@ -105,15 +105,74 @@ int branch_block_free(branch_block_t *branch)
...
+        case FORENDFOR:
+        case WHILEENDWHILE:
+	    if(block->num_conditionals > 2 || block->num_actions != 1)
+	    {
+	        return FAILURE; //Invalid while syntax
+	    }
+	    conditional_block_t *loop = block->conditionals[0];
+	    if (block->num_conditionals == 2)
+	    {
+	      // If the condition to end the loop is different than starting the loop
```
MaxineK36, line 160: This should be a block comment (`/* ... */`)

src/custom-actions/src/branch_block.c
```diff
@@ -105,15 +105,74 @@ int branch_block_free(branch_block_t *branch)
...
+	        loop = block->conditionals[1];
+	    }
+	    if (eval_conditional_block(block->conditionals[0]) == 0)
+	    {
+	        //No detection or protection for infinite looping
```
MaxineK36, line 165: Should also be a block comment

MaxineK36 (changes requested):
Looks good overall! A couple of places could use some clarification

MaxineK36 (commented):
Why is 0 true? In C, generally true is 1 and false is 0; it would be helpful to maintain this convention in the control blocks unless there's a particular design reason for doing otherwise.

LapsededPacifist:
> Why is 0 true? In C, generally true is 1 and false is 0; it would be helpful to maintain this convention in the control blocks unless there's a particular design reason for doing otherwise.

This is actually a larger problem with custom actions that we just uncovered. TRUE is defined as 0 and FALSE is defined in 1 across all of custom actions.
I believe that the reason for this is match with strcmp, so that it can be used for the list search functions.
This is a significant problem because CLI currently defines TRUE as 1 and FALSE as 0, which means that there are definition conflicts when trying to make custom-actions with the rest of chiventure.

MaxineK36 (approved):
Looks good!

MaxineK36:
Issue Score: ✔️++

Comments:
Great work! Your closing statement is particularly well-written, especially as it explains an unanticipated change that resulted from this PR (changing the macro T/F values).

PR #1104, opened 2021-05-26
Integrate npc_battle struct with battle code

lczielinski:
In this pull request, we replaced the old npc_enemy_t struct with the new struct implemented by the npc team, npc_battle_t. This new struct is part of the more general npc_t struct. We modified functions especially in battle_flow, since many functions involved taking in an npc_enemy_t and turning it into a combatant_t. One difference with an npc_t and npc_enemy_t is that npc_t (and npc_battle_t) requires moves, stats, and descriptions to not be null. This caused some changes in the way enemies were created and their moves created with build_moves() (the move had to be created with the enemy, rather than the enemy being created then the move being added). Furthermore, the npc_t struct does not contain next and prev fields like the old npc_enemy_t struct did, so for now, multiple enemies in a single battle is not implemented. We updated all tests and examples, which function just as before.


firatciftci:
@lczielinski It seems the automated tests for this PR's branch is currently failing due to other PRs being merged to `dev`. You should check out the CI log and try to fix the issue before the reviewers can submit their reviews. Let me know if you have any questions/issues with this.

lczielinski:
The issue was with linking the npc and game-state libraries. The fix was to add the line "target_link_libraries(game-state npc)" to the game-state CMake file.

adinilfeld (approved):
Looks great!

sonnykaiyuenlee (approved):
looks like you guys changed a lot, good work!

firatciftci (approved):
Everything looks good to me, great work!

lczielinski:
To summarize this pull request, the NPC team integrated our npc_enemy_t struct into their main npc_t struct. We refactored functions, especially in battle_flow.c, tests, and examples to use this new struct, and removed the old struct from our code. The npc_battle_t struct (the new struct within npc_t) cannot be created with null stats and moves, and the npc_t cannot be created with null descriptions. In the future, we will add more functionality for the npc_t struct to hold battle_item_t that enemies can use in battle.

firatciftci:
Issue Score: ✔️++

Comments:
Great work on this!

Issue #1060, opened 2021-05-19
Create new CLI context struct (pending history command merge)


dchen0258:
Currently the command_history is set to be in the chiventure_ctx struct, however, if more cli specific features are to be added, it makes sense to create a new cli_ctx struct to contain this information. The command_history and lookup table should both be placed in this new struct, and any future cli specific information.

dchen0258:
This issue should consist of: 
- Creating new struct in a new header file (similar to how the ui context struct is constructed)
- Putting the look up table and command history
- Refactoring all code that references ctx->lookup or ctx->command_history to reference the cli_ctx struct

Beza20:
The creation of a new module: cli_ctx.c to handle the new struct was a good idea and is now in the cli/ctx branch. The module now has 2 functions to allocate and free the struct. I am trying to see whether or not to add a particular function to initialize the struct like we have almost all other structs. We will address this issue after further correspondence with @dchen0258 and maybe @ecoble

dchen0258:
Pull request submitted.

dchen0258:
Closing statement:
This struct was aimed at making it easier for future cli specific components to be added, so hopefully future classes will find some cool enhancements!

ecoble:
Issue Score: ✔️++

Comments:
Excellent work making the new context struct!

PR #1108, opened 2021-05-26
Implementing the new cli context struct

dchen0258:
Linked to issue #1060 

Goal: Currently the command_history is set to be in the chiventure_ctx struct, however, if more cli specific features are to be added, it makes sense to create a new cli_ctx struct to contain this information. The command_history and lookup table should both be placed in this new struct, and any future cli specific information.

Changes: Big changes were new ctx header and c file for the cli_ctx struct. The rest of the changes was a lot of refactoring of code.


src/cli/src/cli_ctx.c
```diff
@@ -0,0 +1,32 @@
...
+}
+
+
+int free_cli_ctx(cli_ctx_t *cli_ctx)
+{
+    if(cli_ctx == NULL || cli_ctx->command_history == NULL)
+    {
+        return SUCCESS;
+    }
+    free(cli_ctx);
```
ecoble, line 28: You'll probably need to iteratively free each element in the command_history here, something like this (taken from the utlist docs, so you will use LL instead of DL): 

DL_FOREACH_SAFE(head,elt,tmp) {
      DL_DELETE(head,elt);
      free(elt);
}

src/common/src/ctx.c
```diff
@@ -18,6 +16,8 @@ chiventure_ctx_t* chiventure_ctx_new(game_t *game)
         return NULL;
     }
 
+    ctx->cli_ctx = cli_ctx_new();
```
ecoble, line 19: This should be moved into the init function, where the initialization of the table the command_history list used to be.

ecoble (changes requested):
Great work! A few small changes.

borjasotomayor (approved):
The changes to the non-CLI code look good to me!

mjgoodman321 (approved):
Looks solid to me!

ecoble (approved):

dchen0258:
Closing statement:
This struct was aimed at making it easier for future cli specific components to be added, so hopefully future classes will find some cool enhancements!

ecoble:
Issue Score: ✔️++

Comments:
Great job making the new context struct! I think this will definitely help future teams as well as they expand the CLI.

Issue #1058, opened 2021-05-19
Implement the code associated with the quests section in the Player Interface Implementation Documentation


afranzese99:
As part of implementing the player interface, we reached out to multiple teams (playerclasss, skilltrees, quests, and battles) to best understand how their respective modules would interact with the player module. We took note of these discussions and created a [wiki page](https://github.com/uchicago-cs/chiventure/wiki/Player-~-Interface-Implementation-Plan) to list and describe the necessary code to implement to best serve their needs (Task to document needs: #944). We have implemented the code for skilltress and stats and effects sections of the wiki which is documented in task #952 (this task also includes updates made to the other sections of the wiki after further discussing the needs of certain teams). This task concerns specifically implementing the quests section.

We will have to wait until the quests team implements hash table functionality while handling the quests struct in all of their functions until we can start to implement the code from the plan. (The quest struct has the functionality to be treated as a hash table, but none of the functions in the quest module treat it as a hash, so we decided it would be best to wait for them to update their module before starting to work on their section of the plan.)

For quests we will need to do the following:

- `player_assign_quest()` to add a quest to the quests field (calling `start_quest()` in `quests_state.c`)
- `player_quest_status()`/`player_is_quest_completed()` to get the status of given quest (calling `get_quest_status()` or `is_quest_completed()` in `quests_state.c`)
- `player_complete_achievement()` to mark given achievement for given quest as completed (calling `complete_achievement()` in `quests_state.c` )
- `player_complete_quest()` to mark given quest as complete and give reward item to player (calling `complete_quest()` in `quests_state.c` and the existing `add_item_to_player()` in the player module, as well as `player_is_quest_completed()`)

afranzese99:
After discussion with the quests team, they said they expect to be done implementing the hash functionality into their module over the next couple days (including an `all_quests` hash stored in the game state and an add quest to hash and a get quest from hash function). We expect to be able to implement the stated functions into the player interface soon (hopefully most if not all of them before the code freeze this Friday, May 28). I will reach out again soon to get more updates on timing and any issues they may be running into.

afranzese99:
After further discussion with the quests team, there was a pull-request (pull-request #1092) made 3 days ago with the functionality required to begin implementing the relevant section of the player interface for quests. However, it is still under review, and likely will not be approved with enough time before the code freeze (Friday, May 28th) to implement the functions and code noted in this issue. 

[...]

Issue #1057, opened 2021-05-19
Implement the code associated with the playerclass section in the Player Interface Implementation Documentation


afranzese99:
As part of implementing the player interface, we reached out to multiple teams (playerclasss, skilltrees, quests, and battles) to best understand how their respective modules would interact with the player module. We took note of these discussions and create a [wiki page](https://github.com/uchicago-cs/chiventure/wiki/Player-~-Interface-Implementation-Plan) with the necessary code to implement to best serve their needs (Task to document needs: #944). We have implemented the code for skilltress and stats and effects sections of the wiki which is documented in task #952 (this task also includes updates made to the other sections of the wiki after further discussing the needs of certain teams). This task concerns specifically implementing the playerclass section.

For the playerclass we will need to do the following:

- update `player_set_class() `such that it initializes the `player_combat_skills`, `player_noncombat_skills`, and `player_stats` fields of the player struct.
- Create functionality for deep copying skill inventories

jasulcaf:
In the process of creating the functionality for deep copying skill_inventory_t (which will be called `copy_inventory`), I had to fix some memory leakage that was happening in two other function in skilltrees/inventory.c: 

- `inventory_free`
- `inventory_skill_add`

As a summary of the problem:

1. `inventory_free` had a condition that only free’d the inventory->active whenever inventory->active was greater than zero. However, inventory_new always makes an initial malloc for skills so these two arrays where simply being lost if the inventory was freed with zero skills.

      - Solution: Get rid of the if statement and _always_ free the skill arrays.

2. inventory_skill_add realloced the wrong space. This is easiest to explain with example. Before, if a new inventory had 0 active skills then when we added a skill, the function would change the number of active skills to 1 and realloc the array as: `sizeof(skill_t*) * 1` (aka we still only had an array of size 1!) This was _not right_ because if we tried to add another skill, then the function would first try to place the new skill in the second array spot which doesn’t exist. 

      - Solution: Malloc enough space so that another skill can be properly added without reaching outside of the array

WPDolan:
@afranzese99 and I have updated the `player_set_class()` function to also set the player's `player_stats`, `player_skills`, and `player_effects` fields when setting a player's class. These fields are deep copied using the `copy_inventory()`, `copy_stat()` and `copy_effect()` functions implemented by @jasulcaf. We have also updated the corresponding `player_set_class()` test to reflect our changes.

[...]

Issue #1053, opened 2021-05-19
Allow player_t to store and gain moves


adinilfeld:
Currently, there is no way for a player to unlock moves for future use in battle. To pave the way for future implementation of this functionality, we will add a linked list of `move`s to the `player_t` struct, and refactor the `add_move` function to take in a `player_t` instead of a `combatant_t`.

adinilfeld:
So far, I added a "moves" field to the player_t struct, and transferred the `add_move` function from battle_move_maker to player. Next steps will be figuring out how best to integrate the `execute_move_effect` function, as well as figuring out how to transfer the unit tests for `add_move` (which are currently inside of `tests/battle/`).

adinilfeld:
After moving the tests over to `test_player`, there seems to be some issue with the tests crashing, which I'll have to look into further.

adinilfeld:
Managed to fix the unit tests in `test_player` -- one of them had been adding an item to the player before adding the move, but it was a `battle_item_t`, which `player_t` wasn't meant to hold. So now those work, and @tgao564 to finish up the integration with skilltrees.

adinilfeld:
This issue has now been completed. The `player_t` struct now has a field to store a linked list of moves; the `add_move` function has been refactored to take in a `player_t` instead of a `combatant_t`; and `add_move` has been moved from battle to player. Lastly, the unit tests associated with `add_move` were moved from battle to player as well, and now pass successfully.

This issue will be closed when the associated pull request in approved.

tgao564:
The abovementioned pull request Skilltrees/effect_v2 #1110 has been initiated. Due to its size, it may not go through before the code soft-lock goes into place. Will finish the tests for move-based skills if the PR goes through in time. Otherwise, will leave a backlog issue for next year to finish my implementation.

adinilfeld:
A pull request has been opened which closes this issue. See the summary message from above: 

> This issue has now been completed. The `player_t` struct now has a field to store a linked list of moves; the `add_move` function has been refactored to take in a `player_t` instead of a `combatant_t`; and `add_move` has been moved from battle to player. Lastly, the unit tests associated with `add_move` were moved from battle to player as well, and now pass successfully.

firatciftci:
Issue Score: ✔️++

Comments:
Great work on this! The updates that you have provided give a really detailed overview of your work.

PR #1123, opened 2021-05-27
Implemented the Playerclass section of the Player Interface Plan with functionality for deep copying skill inventories

WPDolan:
We have implemented our planned functionalities described under the "Playerclass" section of our player interface implementation [plan](https://github.com/uchicago-cs/chiventure/wiki/Player-~-Interface-Implementation-Plan) as outlined in Issue #1057.

Changes Include:
- The addition of the  `copy_inventory` function to deep copy `skill_inventory_t` structs
-  Updates to the `player_set_class()` function and its corresponding test to also set the player's `player_stats`, `player_skills`, and `player_effects` fields when setting a player's class. These fields are deep copied with `copy_inventory()`, `copy_stat()` and `copy_effect()`
- Updates to `player_free()` to also free the newly deep copied `player_stats`, `player_skills`, and `player_effects` fields

Deep copying fields from the given `class_t` to the `player_t` when setting the class will allow for multiple players to be the same class without the risk of unintentionally modifying every player's stats/effects/skills when modifying the fields of a single player.

The remaining sections of the interface plan will be implemented into the player module in Issues #1053 and #1058.


src/game-state/src/player.c
```diff
@@ -75,12 +107,22 @@ int player_free(player_t* plyr)
         free(plyr->player_race);
     }
 
-    if (plyr->player_class != NULL)
+    delete_all_items(&plyr->inventory);
+
+    if (plyr->player_skills != NULL)
     {
-        class_free(plyr->player_class);
```
LapsededPacifist, line 80: Not really certain if this is your job, but now that player isn't freeing their own class, should the list of all classes be freed in game? This is the only place that I know of where class_free gets called, so removing it might mean that every class struct is a memory leak.

LapsededPacifist (commented):
I think it looks good. I had the one comment about when classes get freed. Everything else in game-state looked good.
I didn't really review for the changes made to the skill trees section. I briefly looked over the code, and it looked OK. However, I don't really understand many of the changes there.

src/skilltrees/src/inventory.c
```diff
@@ -70,14 +67,43 @@ int inventory_skill_add(skill_inventory_t* inventory, skill_t* skill) {
...
+
+
+/* See inventory.h */
+skill_inventory_t* copy_inventory(skill_inventory_t* original)
+{
+    assert(original != NULL);
+
+    skill_inventory_t* copy = inventory_new(original->max_active, original->max_passive);
+    if (copy == NULL)
+        return NULL;
```
rohangupta2806, line 87: I would add an error message out here if the copy is NULL

tests/skilltrees/inventory_test.c
```diff
@@ -30,6 +30,108 @@ Test(inventory_tests, inventory_free_test)
...
+}
+
+Test(inventory_test, deep_copy_inventory_with_skills)
+{
+    int ret_val;
+    skill_inventory_t* inventory = inventory_new(3,4);
+    cr_assert_not_null(inventory, "inventory_new() failed to create inventory struct");
+
+    /* Add two skills for both Passive and Active field */
+    skill_t* activ1 = (skill_t*)malloc(sizeof(skill_t));
```
rohangupta2806, line 74: You should be able to just use skill_new here

rohangupta2806 (changes requested):
Looks good!  I just suggested a couple of changes that are pretty minor.

src/skilltrees/src/inventory.c
jasulcaf: Good catch, I was not being consistent with the rest of the code returning error messages in this type of situation. Added this

tests/skilltrees/inventory_test.c
jasulcaf: I didn't think that I needed all the specific details of a skill to test my function which is why I just malloc'ed, but using skill_new would make the code easier to read, so I am also incorporating this change.

rohangupta2806 (approved):
All seems good now!

src/game-state/src/player.c
WPDolan: We believe that classes will be allocated at the beginning of the game, and will provide the starting skills/stats for players that choose the class. Some games may allow players to change classes throughout the course of a play session, so we believe that classes will be allocated at game creation, and freed at the end of the game.

LapsededPacifist (approved):
My only question was answered. I think the code looks good.

jasulcaf:
This is a general comment regarding a bug with the Pull Request Logs. I am an assignee to this Pull Request, yet in the process of resolving a comment from @rohangupta2806 I was added as a reviewer to this Pull Request. As such, this is just to clarify that I am _not_ a reviewer and this Pull Request does not require any approval from me to be merged.

include/game-state/player.h
```diff
@@ -89,16 +89,19 @@ int delete_all_players(player_hash_t* players);
 
 /*
  * Sets an allocated player_t object's player_class field to given class_t class
+ *          Makes deep copies of the given class's "starting_skills",
+ *          "base_stats", and "effects fields" and updates the corresponding
+ *          fields in the player struct
  *
  * Parameters:
- *  plyr: A player. Must point to already allocated memory.
+ *  player: A player. Must point to already allocated memory.
```
MaxineK36, line 97: Looks like `plyr` is used pretty consistently as a parameter in this file; please either change all instances or no instances (unless you have a particular reason for just doing this one)

src/game-state/src/player.c
MaxineK36: This makes sense to me! In general, frees should happen in parallel to allocations (so if a player doesn't allocate its own class, it shouldn't free it either)

MaxineK36: Feel free to make a backlog issue (bug) for classes being freed if you think this should be happening but isn't

src/skilltrees/src/inventory.c
```diff
@@ -36,14 +36,11 @@ skill_inventory_t* inventory_new(unsigned int max_active,
...
 
-    if (inventory->num_active > 0) {
-        free(inventory->active);
-    }
-
-    if (inventory->num_passive > 0) {
-        free(inventory->passive);
-    }
+    /* Free the arrays of skills */
+    free(inventory->active);
```
MaxineK36, line 40: Are these guaranteed to be non-null?

src/skilltrees/src/inventory.c
```diff
@@ -70,14 +67,45 @@ int inventory_skill_add(skill_inventory_t* inventory, skill_t* skill) {
...
+    assert(original != NULL);
+
+    skill_inventory_t* copy = inventory_new(original->max_active, original->max_passive);
+    if (copy == NULL)
+    {
+        fprintf(stderr, "Failed to allocate memory for skill inventory copy");
+        return NULL;
+    }
+
+    /* Fill Array of Active Skills */
```
MaxineK36, line 92: This seems to work with the existing functions, but it seems like these sets skills should really be linked lists (one for active, one for passive). Do you have any reason to believe this is not the case?

Otherwise, it would be helpful if you would make a backlog issue for this, as it would be a good task for future teams.

MaxineK36 (changes requested):
A few small fixes/clarifications!

include/game-state/player.h
WPDolan: I have replaced all instances of `plyr` with `player` in both player.h and player.c

src/skilltrees/src/inventory.c
jasulcaf: Yes. If we take a look at chiventure/src/skilltrees/src/inventory.c, the `inventory_new()` function is always preventively mallocing this memory.

src/skilltrees/src/inventory.c
jasulcaf: I believe that the reason this is not the case is because the skilltrees team did not think there'd be a ton of skills added at once, and thus preferred the quick nature of being able to call any specific skill at once (vs. having to go through a lot of linked skills).

rohangupta2806: Yeah I think the previous years did decide on arrays, and if an array is full, they would make a new array and copy everything over.  We never got around to changing it this year.  We might make a backlog issue on this now that you've pointed it out.

src/game-state/src/player.c
WPDolan: I believe that no current games in chiventure start off with allocating classes, so hopefully this shouldn't be a bug as long as the playerclass team addresses freeing when classes are integrated into playable games.

MaxineK36 (approved):
Looks good!

WPDolan:
Closing statement:

The pull request has been merged, and the playerclass section of the [interface plan](https://github.com/uchicago-cs/chiventure/wiki/Player-~-Interface-Implementation-Plan) is now complete, along with additional required functionality for creating deep copies of `skill_inventory` structs.

The core changes and additional functions listed in the opening statement of this pull request remain unchanged. They include:
- The addition of the  `copy_inventory` function to deep copy `skill_inventory_t` structs
-  Updates to the `player_set_class()` function and its corresponding test to also set the player's `player_stats`, `player_skills`, and `player_effects` fields when setting a player's class. These fields are deep copied with `copy_inventory()`, `copy_stat()` and `copy_effect()`
- Updates to `player_free()` to also free the newly deep copied `player_stats`, `player_skills`, and `player_effects` fields

Future pull requests will aim to complete additional sections listed under the interface plan.

MaxineK36:
Issue Score: ✔️++

Comments:
Great work! Your closing statement is particularly well-done.

PR #1124, opened 2021-05-28
Implementing lowercase case insensitivity

mjgoodman321:
Case insensitivity is currently unoptimally implemented in Chiventure by uppercasing the vast majority of inputs. The goal of this branch was to streamline case insensitivity by creating a helper function and implementing across as much of chiventure as was necessary to make the whole game case insensitive. Also, case insensitivity was changed to lowercasing the input instead of uppercasing to match industry standards

Linked to Issue #392


include/game-state/room.h
```diff
@@ -3,6 +3,8 @@
 
 #include "game_state_common.h"
 #include "item.h"
+#include "cli/cli-utility.h"
+
```
LapsededPacifist, line 7: There shouldn't be an extra space here.

include/game-state/game_state_common.h
```diff
@@ -36,4 +36,8 @@
 #define ROOM_SRC_NULL 2
 #define ROOM_DEST_NULL 3
 
-#endif
+
+
+
+
+#endif
```
LapsededPacifist, line 43: This isn't really a problem, but I am very confused by this. The white space was nice before and I am very confused as to why it was changed.

LapsededPacifist (changes requested):
I didn't have any problems with the code modifications. Assuming that case_insensitize2() just changes it lower case (Which it appears to), I don't think there will be any functional problems.
The only change that I want made, is the extra line in the includes in room.h.

mjgoodman321:
Made one last change to the code because I was reviewing some monkeypatched code, and realized that because everybody uses the add_entry function with capitalized strings that I was about to break a lot of code. Decided it was easier to make sure that you could call add_entry with any case of string than to require people to only use lowercase strings

include/game-state/game_state_common.h
mjgoodman321: That's where I put the case insensitivity functions there to begin with, didn't realize I left that space there when I removed them. Sorry!

LapsededPacifist (approved):
The minor changes to the spacing that I requested look good. 
I am only approving for game-state and action-management, since I only looked at the game-state and action-management code.

src/action_management/src/actionmanagement.c
```diff
@@ -126,13 +126,13 @@ int do_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *i, char **ret_
     }
 
     /* use representative c_name for action synonyms */
-    if(strncmp(a->c_name, "PICKUP", BUFFER_SIZE) == 0) 
+    if(strncmp(a->c_name, "pickup", BUFFER_SIZE) == 0)
```
Beza20, line 129: Why did you need to changed these hard coded strings? shouldn't the fact that you have now made chiventure case insensitive have no effect on how we write our hard coded strings at least that's what I was hoping for when putting in the delimiter for the parse_r function

Beza20:
Beautiful! great code to read and follow through and a very valuable functionality for chiventure.

Beza20 (approved):

src/action_management/src/actionmanagement.c
mjgoodman321: Well because case insensitivity is being implemented as making everything lowercase, so some hard-coded strings (specifically those that aren't called by a function that lowercases them) have to be made lowercase to work in the framework of the game. There is a version of strcmp that doesn't care about case, but I'm not sure it can be used here because I don't know how important it is to use strncmp here as opposed to a version of the generic strcmp.

include/cli/cli-utility.h
```diff
@@ -0,0 +1,34 @@
...
+
+
+/*
+ * Function used to make struct IDs and other identifiers
+ *      case insensitive
+ * Parameters:
+ *   -  input: the string being lower-cased
+ * 
+ */
+void case_insensitize1(char *input);
```
ecoble, line 22: I think it would be better to name these functions something more descriptive than case_insensitize1 and case_insensitize2, since those do not describe the differences between these functions. You could maybe name the first one case_insensitize, and the other case_insensitize_alloc, or something along those lines. I realize this is inconvenient since the function is called in so many places, but using find and replace should help.

src/wdl/src/load_item.c
```diff
@@ -30,6 +30,8 @@ action_type_t *get_game_action(char *action, list_action_type_t *valid)
         curr = curr->next;
     }
 
+    if(curr == NULL)
```
ecoble, line 33: Space after if

ecoble (changes requested):

src/game-state/src/game_action.c
```diff
@@ -28,7 +28,7 @@ game_action_t *game_action_new(char *action_name, char *success_str, char *fail_
     new_action->success_str = calloc(1, MAX_MSG_LEN * sizeof(char));
     new_action->fail_str = calloc(1, MAX_MSG_LEN * sizeof(char));
 
-    int check = game_action_init(new_action, action_name, success_str, fail_str);
+    int check = game_action_init(new_action, case_insensitized_string(action_name), success_str, fail_str);
```
borjasotomayor, line 31: This results in a memory leak, because the string returned by `case_insensitized_string` is copied into `new_action->action_name` inside `game_action_init`, but the original string is never freed. In general, the function would have to be called like this:

```
char *action_name_case = case_insensitized_string(action_name);
int check = game_action_init(new_action, action_name_case, success_str, fail_str);
free(action_name_case);
```

However, in this particular case, it may simply make more sense to call `case_insensitized_string` inside `game_action_init`: after action_name is copied into `new_action->action_name`, just case-insensitize that string (which avoids having to allocate a separate case-insensitized string)

src/game-state/src/game_action.c
```diff
@@ -51,6 +51,7 @@ game_action_t *game_action_new(char *action_name, char *success_str, char *fail_
 /* see game_action.h */
 game_action_t *get_action(item_t *item, char *action_name)
 {
+    action_name = case_insensitized_string(action_name);
```
borjasotomayor, line 54: Avoid altering the content of the parameter in this way. Instead, do something like this:

```
char *action_name_case = case_insensitized_string(action_name);
```

And then remember to free `action_name_case` at the end of the function.

src/game-state/src/item.c
```diff
@@ -25,6 +26,9 @@ item_t *item_new(char *item_id, char *short_desc, char *long_desc)
 {
     item_t *new_item = malloc(sizeof(item_t));
     memset(new_item, 0, sizeof(item_t));
+
+    item_id = case_insensitized_string(item_id);
```
borjasotomayor, line 30: Same as the comment in `src/game-state/src/game_action.c:31`. `item_id` needs to be freed, but a better solution may be to case-insensitize `new_item->item_id` inside `item_init` after `item_id` has been copied into it.

src/game-state/src/path.c
```diff
@@ -2,10 +2,13 @@
 #include "game-state/game_action.h"
 #include "game-state/room.h"
 
+
 /* See room.h */
 path_t *path_new(room_t *dest, char *direction)
 {
 
+    direction = case_insensitized_string(direction);
```
borjasotomayor, line 10: Instead of allocating a new string, just case-insensitize `path->direction` at the end of the function.

src/game-state/src/room.c
```diff
@@ -22,6 +23,8 @@ int room_init(room_t *new_room, char *room_id, char *short_desc,
 room_t *room_new(char *room_id, char *short_desc, char *long_desc)
 {
 
+    room_id = case_insensitized_string(room_id);
```
borjasotomayor, line 26: Instead of allocating a new string here, case-insensitize `new_room->room_id` in `room_init`.

src/game-state/src/room.c
```diff
@@ -127,6 +130,8 @@ path_t *path_search(room_t *room, char* direction)
         return NULL; //cannot search path in NULL room
     }
 
+    direction = case_insensitized_string(direction);
```
borjasotomayor, line 133: Same as comment in `src/game-state/src/game_action.c:54`. Avoid overwriting the value of the parameter. Create a new variable for the case-insensitized string, and remember to free it once you are done with it.

src/game-state/src/room.c
```diff
@@ -177,6 +182,8 @@ path_t *list_paths(room_t *room)
 /* see room.h */
 item_t* get_item_in_room(room_t* room, char* item_id)
 {
+    item_id = case_insensitized_string(item_id);
```
borjasotomayor, line 185: Same as comment in `src/game-state/src/game_action.c:54`. Avoid overwriting the value of the parameter. Create a new variable for the case-insensitized string, and remember to free it once you are done with it.

src/game-state/src/room.c
```diff
@@ -197,6 +204,8 @@ room_t *find_room_from_path(path_t *path)
 /* See room.h */
 room_t *find_room_from_dir(room_t *curr, char* direction)
 {
+    direction = case_insensitized_string(direction);
```
borjasotomayor, line 207: Same as comment in `src/game-state/src/game_action.c:54`. Avoid overwriting the value of the parameter. Create a new variable for the case-insensitized string, and remember to free it once you are done with it.

src/wdl/src/load_item.c
```diff
@@ -30,6 +30,8 @@ action_type_t *get_game_action(char *action, list_action_type_t *valid)
         curr = curr->next;
     }
 
+    if (curr == NULL)
+        return NULL;
```
borjasotomayor, line 34: Add blank line after this one.

src/wdl/src/load_item.c
```diff
@@ -57,19 +59,19 @@ int load_actions(obj_t *item_obj, item_t *i)
     obj_t *curr;
     DL_FOREACH(action_ls->data.lst, curr)
     {
-        temp = get_game_action(obj_get_str(curr, "action"), val_actions);
+        temp = get_game_action(case_insensitized_string(obj_get_str(curr, "action")), val_actions);
 
         if (obj_get_str(curr, "text_success") != NULL && obj_get_str(curr, "text_fail") != NULL)
         {
-            add_action(i, obj_get_str(curr, "action"), obj_get_str(curr, "text_success"), obj_get_str(curr, "text_fail"));
+            add_action(i, case_insensitized_string(obj_get_str(curr, "action")), obj_get_str(curr, "text_success"), obj_get_str(curr, "text_fail"));
```
borjasotomayor, line 66: Avoid calling `case_insensitized_string(obj_get_str(curr, "action")` twice to compute the same value. Instead, compute it once at the top of the loop:

```
char *action = case_insensitized_string(obj_get_str(curr, "action"));
```

And remember to free it at the end of the loop.

src/wdl/src/load_item.c
```diff
@@ -116,7 +118,7 @@ int load_items(obj_t *doc, game_t *g)
         }
 
         //retrieve the pointer for the room that the item is located in
-        room_t *item_room = find_room_from_game(g, in);
+        room_t *item_room = find_room_from_game(g, case_insensitized_string(in));
```
borjasotomayor, line 121: The string allocated by `case_insensitized_string(in)` is never freed. That said, calling `case_insensitized_string` here is redundant since `find_room_from_game` already case-insensitizes the `room_id` parameter.

src/game-state/src/game.c
```diff
@@ -225,6 +225,7 @@ player_t *get_player(game_t *game, char *player_id)
 /* See game.h */
 room_t *find_room_from_game(game_t *game, char* room_id)
 {
+    room_id = case_insensitized_string(room_id);
```
borjasotomayor, line 228: Same as comment in `src/game-state/src/game_action.c:54`. Avoid overwriting the value of the parameter. Create a new variable for the case-insensitized string, and remember to free it once you are done with it.

src/cli/src/cmd.c
```diff
@@ -12,6 +12,7 @@
 /* === hashtable helper constructors === */
 void add_entry(char *command_name, operation *associated_operation, action_type_t *action, lookup_t **table)
 {
+    command_name = case_insensitized_string(command_name);
```
borjasotomayor, line 15: Instead of allocating a new string, just case-insensitize `t->name`

src/wdl/src/load_room.c
```diff
@@ -57,7 +58,7 @@ int add_connections_to_rooms(obj_t *doc, game_t *g)
     HASH_ITER(hh, rooms_obj->data.obj.attr, curr, tmp)
     {
         // obtain room id
-        char *id = curr->id;
+        char *id = case_insensitized_string(curr->id);
```
borjasotomayor, line 61: Case-insensitizing `curr->id` is redundant since that value is only used to call `create_connection`, which in turn calls `find_room_from_game`, which already case-insensitizes the `room_id` parameter.

src/wdl/src/load_room.c
```diff
@@ -73,7 +74,9 @@ int add_connections_to_rooms(obj_t *doc, game_t *g)
         {
             // get id of room we are going to and direction
             char *to = obj_get_str(conn_curr, "to");
```
borjasotomayor, line 76: Same issue as `curr->id` above.

src/wdl/src/load_room.c
```diff
@@ -73,7 +74,9 @@ int add_connections_to_rooms(obj_t *doc, game_t *g)
...
+            to = case_insensitized_string(to);
             char *direction = obj_get_str(conn_curr, "direction");
+            direction = case_insensitized_string(direction);
```
borjasotomayor, line 79: Similar to `curr->id` and `to`: `direction` eventually makes its way to `path_new`, which already case-insensitizes the `direction` parameter.

borjasotomayor (changes requested):
The changes to the non-CLI part of the code look overall good, but the use of the `case_insensitized_string` function has to be reviewed carefully: in many cases, the string allocated by this function is not being freed, or the use of `case_insensitized_string` is not necessary at all, since the string that is being case-insensitized is being passed to a function that already case-insensitizes its inputs.

Justin08784 (changes requested):
I checked out your most recent commit ee646f7ea5cc5286d8e18623e471edfd5062ddfc.

A massive number of tests are failing and/or crashing in several modules, including openworld, game-state, action-management, and quests.

Justin08784 (approved):
My previous comment was based on a complete misunderstanding. Changes in the openworld module look fine.

include/game-state/item.h
```diff
@@ -4,6 +4,7 @@
 #include "game_state_common.h"
 #include "action_management/action_structs.h"
 #include "game-state/stats.h"
+#include "cli/cli-utility.h"
```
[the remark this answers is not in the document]
mjgoodman321: Does this apply to other cli files like the example above, or should it only be included in .c files for other modules?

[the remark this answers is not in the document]
borjasotomayor: Since this header files does not include any struct declarations, it should not appear in any `.h` files. Instead, `#include` it only in the C files that use the case-insensitize functions.

src/game-state/src/game_action.c
mjgoodman321: So I know you said I should use `case_insensitized_string` inside of `game_aciton_init`, but could I try using the regular `case_insensitize` since the `new_action->action_name` is already calloced? Since that way there's no chance of producing a string that isn't freed elsewhere.

borjasotomayor: Yes, sorry, that's what I meant: after `action_name` is copied into `new_action->action_name`, you can just case-insensitize it in-place (using the `case_insensitize` function)

src/cli/src/cmd.c
mjgoodman321: Would it be better to case-insensitize newname as it is assigned to t->name like so:
`t->name = case_insensitize(newname);`
Or to do it separately on the line after?

mjgoodman321: Wait nevermind that was a dumb question, case_insensitize returns void

src/game-state/src/game_action.c
mjgoodman321: For some reason after implementing this, the two tests in test_item.c have started failing due to `possible_actions` returning a failure, which I'm really confused by since both the modifications to this function and to the load_actions function (the only other function I could find that would affect those tests) shouldn't affect their behavior. I'll go ahead and push all the changes I've made to address your comments so you can see it yourself.

borjasotomayor: This turned out to be a very subtle bug in how actions were being added to items (which, as far as I can tell, has been in the repository for two years, and I'm surprised it took so long to surface). In a nutshell, whenever you add a struct to a uthash hash table, the key must be in the struct you are adding (it can't be an arbitrary string, separate from the struct). `add_action` was using the `action_name` parameter as the key, when it should've used the `action_name` in the `action` struct.

Anyway, I've pushed the fix, which now makes all the tests pass: https://github.com/uchicago-cs/chiventure/commit/4bfb908133de45c3c81400387fab052cddd213e1

borjasotomayor (approved):

ecoble (approved):

mjgoodman321:
Closing Statement: case insensitivity has been implemented through the use of two helper functions, case_insensitize and case_insensitized_string, which respectively change a string at its source into its lowercase version and produce a lowercase version of a string out of a given input. Various other changes have been made to hardcoded strings throughout chiventure in order to properly implement case insensitivity (most notably in the actionmanagement module).

ecoble:
Issue Score: ✔️++

Comments:
Great work on a tough issue! Nice job getting all the proper reviewers, and keeping this PR up to date.

Issue #1054, opened 2021-05-19
Create sandbox example for demo of the DSL feature


tessied:
Our goal is to create a sandbox with example files to demonstrate the features of the DSL we have been working on this quarter. This demo will be presented during the final presentation.

Note: while adding this issue to the backlog, we considered the fact that this issue would depend on how far we get in terms of our implementation. 

Assignees: John and Tessie
team:dsl

tessied:
We had a team meeting where we discussed what we wanted to include in the demo. We decided that we wanted to have a standard DSL file that has the main features of the grammar. One thing we noted was that we cannot use custom actions as they are not currently supported in chiventure. Once we finish up the DSL file, we will test it with the parser and attempt to load the game in chiventure.

tessied:
I created a new branch `dsl/demo` and made a new directory `demo_example` where you can find the `demo.dsl`. I was able to successfully load it into chiventure after fixing the errors, but actions currently do not work (commit [`48e77f0`](https://github.com/uchicago-cs/chiventure/commit/48e77f07ed1203cc08b541b912c27994c7b4976d)) For example, when I specify the action `TAKE` under an object, in the game it says the action is not supported for the object.

tessied:
I received help from @r-e-stern and fixed the issues regarding actions (commit [`3daa30a`](https://github.com/uchicago-cs/chiventure/commit/3daa30ae1978f0eb432d666d85ab5548d9b09f3e)). I changed all calls to `USE`, `DRINK`, and `EAT` to `CONSUME`, and now the actions can be run in chiventure. We will add a backlog issue specifying that in the future, the parser should be able to convert all calls to USE`, `DRINK`, and `EAT` to `CONSUME` on its own.

tessied:
The pull request linked to this issue reflects the changes made, that is, a demo file `demo.dsl` is now included in `src/dsl/examples/dsl` which has our standard example of a game written in our DSL that we will present. Pull request has been merged, closing the issue.

tessied:
The pull request linked to this issue reflects the changes made, that is, a demo file `demo.dsl` is now included in `src/dsl/examples/dsl` which has our standard example of a game written in our DSL that we will present. Pull request has been merged, closing the issue.

ecoble:
Issue Score: ✔️++

Comments:
Great work creating this demo, and updating this issue with your progress!

PR #1134, opened 2021-05-28
Created DSL Demo File 

tessied:
We have added the demo file named `demo.dsl` to the `examples` directory in `src/dsl`. This file contains an example that we wrote to showcase the features of our grammar and parser. The file contains features such as creating a game, creating rooms, specifying connections, and adding properties. We hope to present this demo during the final presentation.

This pull request reflects the implementation done in issues #1054.


src/dsl/examples/dsl/demo.dsl
```diff
@@ -0,0 +1,73 @@
...
+ROOM bathroom 
+    short desc: "This is the bathroom"
+    long desc: "The toilet is broken. You can only use the faucet to wash your hands in this bathroom."
+    connections: NORTH TO lobby
+
+    ITEM faucet
+       short desc: "A faucet in the bathroom"
+       long desc: "Use the faucet to wash your hands"
+
+    action: CONSUME
```
jacknugent1529, line 26: Is `CONSUME` the most descriptive action here?

src/dsl/examples/dsl/demo.dsl
```diff
@@ -0,0 +1,73 @@
...
+ROOM storage room
+    short desc: "A storage room."
+    long desc: "There are various items, including a toolbox, all caked with dust. The door on the west is locked but looks like the lock can be picked."
+    connections: WEST TO computer lab
+                 DOWN TO kitchen
+
+    ITEM toolbox IN storage room
+        short desc: "A toolbox caked in dust."
+        long desc: "The toolbox contains a very rusty hammer and screwdriver. There are a lot of tools missing."
+    action: CONSUME
```
jacknugent1529, line 59: It might make sense to also have a different action here.

jacknugent1529 (commented):
The demo file looks great! I was able to test that it is parsed properly and all the actions in chiventure function as expected. It's a fun game to play and it really demonstrates the conciseness and various features of the language. One small issue has to do with the `CONSUME` action being used in cases where different actions might make more sense.

src/dsl/examples/dsl/demo.dsl
ecoble, line 59: Should this action be "TAKE"?

src/dsl/examples/dsl/demo.dsl
ecoble, line 26: Should this be "USE"?

ecoble (commented):
Looks good but I just have a couple questions about some of the actions on the items.

src/dsl/examples/dsl/demo.dsl
tessied: From what we gathered, when a user specifies the action to `USE`, `EAT` or `DRINK`, they are calling the synonymous action `CONSUME`. When specifying `USE` for the `toolbox` object in the grammar, chiventure does not allow the action because it does not recognize the object with the 'CONSUME' action, which is the one that it is calling when typing `USE` into the command line. I originally had `USE` as the action but after realizing this, I changed it to `CONSUME` and now one can call `USE` in the command line with `toolbox` and the action works. We will add a backlog issue specifying that in the future, the parser should make the conversion from `USE`, `EAT` or `DRINK` to `CONSUME` on its own so the author of the dsl can specify these more appropriate actions directly.

src/dsl/examples/dsl/demo.dsl
tessied: This is the same problem as the `toolbox` object above. By specifying the action `CONSUME`, the action `USE` can be called in the command line by the user. The current parser does not support this conversion from `USE`, `EAT` or `DRINK` to `CONSUME` on its own, and we will add it to a backlog issue.

tessied: When specifying USE for the `faucet` object in the grammar, chiventure does not allow the action because it does not recognize the object with the 'CONSUME' action, which is the one that it is calling when typing USE into the command line. We will add a backlog issue specifying that in the future, the parser should make the conversion from USE, EAT or DRINK to CONSUME on its own so the author of the dsl can specify these more appropriate actions directly.

src/dsl/examples/dsl/demo.dsl
tessied: This is also the case that `USE` needs to be specified as `CONSUME` for chiventure to be able to associate `USE` with the object.

jacknugent1529:
Regarding the `USE` and `CONSUME` issue: it takes less than 10 lines of code to create aliases to fix that problem. Should we include that code here, or does it need its own pull request?

tessied:
> Regarding the `USE` and `CONSUME` issue: it takes less than 10 lines of code to create aliases to fix that problem. Should we include that code here, or does it need its own pull request?

@ecoble what would be the best course of action here?

src/dsl/examples/dsl/demo.dsl
ecoble: I see! That is a bit strange, making a note of it sounds good, we can let the action management team know.

ecoble (approved):
Looks good! As for the aliasing, I would not fix that in this pull request. You can make a note of it and open an issue, and maybe let another team know since I am not sure it is a DSL problem, but rather an action management or WDL problem.

jacknugent1529:
Ok, besides that issue, everything looks good! The user can still "USE" the items in the chiventure game itself, and everything else works as expected.

jacknugent1529 (approved):

tessied:
Closing statement: We added a DSL demo file in `src/dsl/examples/dsl`. Since opening the pull request, we discussed the issue of synonymous actions, specifically that `USE`, `DRINK`, and `EAT` all need to be specified as `CONSUME` in the dsl file for chiventure to run the actions; we made note of this issue and will communicate with action management about this. This demo file, which we will use as our standard demo of the dsl file during the presentation, includes an example of a game written in our DSL, including the core features (declaring a game, a room (with connections, properties, items), an item (with properties, actions), adding escape characters for strings). The changes have been merged and the `dev` branch now has the DSL demo file, closing the pull request.

ecoble:
Issue Score: ✔️++

Comments:
Excellent work! The demo file is great, and you both did a great job keeping this PR up to date and doing a thorough review!

PR #1136, opened 2021-05-28
Created a sample game to display new features for GUI

perenewang:
We created a sample game, gui_sample_game.c, that uses GUI for split screen where the image changes when the user moves into a new room and there's a map displayed in the top left corner which also updates when the user moves into a new room.


perenewang:
This PR is only for the gui_sample_game which includes the map in the top left corner and this sample game does include changing the room image based on the player moving.

src/ui/examples/gui_sample_game.c
```diff
@@ -0,0 +1,296 @@
...
+
+        if (mouseOnText) {
+            /* Get pressed key (character) on the queue */
+            int key = GetKeyPressed();
+
+            /* Check if more characters have been pressed on the same frame */
+            while (key > 0) {
+                /* NOTE: Only allow keys in range [32..125}
+                as these are the printable characters */
+                if ((key >= 0) && (key <= 127) && (letterCount < MAX_INPUT_CHARS))
```
xWngl, line 76: Just curious about the 0-127 range when the note says 32-125 are the printable characters?

xWngl (changes requested):
I noticed that while playing the game, as you progress through the rooms, they just kind of disappear. I guess this is because you are unable to go backwards through the map, only forwards. I believe this is fixable by using create_connection() to build the path backwards in addition to forwards

xWngl (approved):
Code and example look good to me!

MaxineK36 (commented):
It looks like most of this is copied from `gui.c`, and you haven't called any of your `gui` functions. Can you help me understand why it's set up this way? Would you be able to do it using `gui` if you could add something to the context (e.g. "char *room_filepath")?

It would be added here: https://github.com/uchicago-cs/chiventure/blob/dev/include/game-state/game.h

perenewang:
We did not call run_gui() fromm gui.c because it does not have the capability to change room image based on the player moving as we don't know the paths to which the room images will be held. Thus, our code in gui_sample_game.c is mostly copied over from gui.c. We decided not to go further with the paths for images as we needed to get this PR merged in time to use gui_sample_game.c in our final presentation demo.

MaxineK36 (approved):
Looks good! Feel free to also use part of your demo time to showcase the chiventure executable with the `--gui` flag, and/or either of the other samples you made this quarter!

grkapoor17:
In this pull request, we have successfully merged code from the gui/sample_game into dev. This involved changes to the ui example CMakeLists, but mostly revolved around creating a new module called gui_sample_game.c. Particularly, we have created a demonstration of all the work that the graphics team has done this year. This includes, but is not limited to, a graphical split-screen interface with a game-related map and game-related images drawn on the upper half of the split screen. In this module we had to create a sample chiventure context game and had to include image files for the rooms associated with the game in src/ui/examples. We then ran our code to implement the graphical drawings in the main() function. We could not directly call run_gui() from gui.c, as this function directly hard codes the path for a singular chiventure image that does not change and we wanted to demonstrate that an image could change as a user entered a new room. We will be using this module to aid our demo for the final presentation.

MaxineK36:
Issue Score: ✔️++

Comments:
Great work overall! Don't forget to make sure all issues/PRs are assigned to a milestone.
Also, the closing statement is particularly well-written for this PR.

PR #1145, opened 2021-05-29
Added npc_one_move function and npcs_in_room_operation

mbutera17:
This pull request contains steps towards our eventual goal of player action leading to an NPC with definite movement moving (we were unable to complete this goal). We did create a new function called `npc_one_move` which will handle the entirety of movement of one NPC. Previously, we had functions to add/remove an NPC from an `npcs_in_room` struct and we had a way to move an NPC down its path in its `npc_mov` struct (`movement` field in NPC struct), but we did not have one function to do both of these and thus completely "move" an NPC. 

Summary of major changes:
Added `npc_one_move` and `npc_one_move_helper` functions
Added `npcs_in_room_operation` function (CLI)
Added tests for `npc_one_move`

PR #1012, opened 2021-05-17
Added new functions in the npc_mov module and updated relevant tests

torresszy:
The two major functions left unfinished from last year, `move_npc_indefinite` and `auto_gen_movement` have been implemented under the `npc_move` module in this pull request. `move_npc_indefinite` will move an existing NPC with an indefinite move type from one room to another. `auto_gen_movement` will randomly generate a list of rooms and append the list to a given NPC's path. In addition, we have implemented two new functions, `get_npc_num_rooms` and `get_num_rooms`. `get_npc_num_rooms` takes a `npc_mov` struct and returns the number of rooms a NPC has in its list of rooms to visit. `get_num_rooms` is a helper function for `auto_gen_movement` that returns the total number of rooms in a given game. In addition, new tests are created for these functions in `test_npc_move.c`. One thing to be noted is that in both `move_npc_indefinite` and `auto_gen_movement`, we have inherited the concept of time from last year NPC team's implementation. However, this concept is not yet well-established in Chiventure at this point. Therefore, Updates on these functions may be needed when a more matured system on "time" is developed in the future.

Summary of major changes:

Added new functions, `move_npc_indefinite`, `get_npc_num_rooms`, `get_num_rooms`,and `auto_gen_movement`, in `npc_move.c` and their headers in `npc_move.h`.
Added new tests for these functions in `test_npc_move.c`

PR #1147, opened 2021-05-29
Demo for NPC_1 2021

torresszy:
This pull request mainly revolves around the file `battle_move_example_2021.c` for our end-of-the-quarter demo. It showcases our works in  `npc_battle`, `npc_move`, and `rooms-npc` modules. Specifically, it shows how NPC movement and battle features can be integrated into Chiventure and used in games at this point. We will demonstrate these new features by moving NPCs between rooms and letting the player interact and battle with the NPCs. The player is also able to loot the dead NPCs. 

Notes:
- One new function, `npcs_in_room_operation`, which prints out all the NPCs in the current room to CLI, is a new feature and monkey-patched to `battle_move_example_2021.c`. However, this function can and should be added to the CLI module in the future officially.
- This branch is branched out of `npc/mov-integration`, therefore, it relies on pull request #1012 and #1145 to be merged successfully.


src/npc/examples/battle_move_example_2021.c
```diff
@@ -0,0 +1,339 @@
+/*
+ * This example program runs a full instance of chiventure with an in-memory game.
+ * The CLI is monkey-patched to accept functions that serve to showcase the movement
+ * and battle functionalities of the NPC.
```
amymorrill, line 4: We should add in-depth comments to explain our monkey-patched functions at the top of the file (see npc_example.c for how last year's npc team did this). I can add a comment for the attack CLI function.

src/npc/examples/battle_move_example_2021.c
```diff
@@ -0,0 +1,339 @@
...
+	return head;
+}
+
+/* Makes sure the game is loaded */
+char *check_game(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
+{
+    game_t *game = ctx->game;
+    if (game == NULL || game->curr_room == NULL)
+    {
+        return "Room not found! Error! Look for Jim if you're not in a room!\n";
```
amymorrill, line 110: We should probably change this because our example no longer includes "Jim"

src/npc/examples/battle_move_example_2021.c
```diff
@@ -0,0 +1,339 @@
...
+{
+    game_t *game = ctx->game;
+    if(game == NULL || game->curr_room == NULL)
+    {
+        print_to_cli(ctx, tokens[0]);
+        return "Error! We need a loaded room to move.\n";
+    }
+
+    move_room(game, arena);
+    add_npc_to_room(arena->npcs, friendly_fiona);
```
amymorrill, line 161: For the purpose of the example, we may not need to change this, but if you try to do the following sequence of commands: 
    "arena", "lobby", "arena"
you get a segmentation fault, which may be because the second "arena" tries to delete the npcs from the lobby, but they were already deleted by the first "arena". It might be worth it to add an if statement for adding the npcs to the arena and deleting them from the lobby so we don't try to move them multiple times.

src/npc/examples/battle_move_example_2021.c
```diff
@@ -0,0 +1,339 @@
...
+{
+    chiventure_ctx_t *ctx = chiventure_ctx_new(NULL);
+
+    game_t *game = game_new("Welcome to Chiventure!");
+
+    load_normal_mode(game);
+
+    /* Initialize the lobby room */
+    lobby = room_new("lobby", "This is lobby", 
+                     "Fiona and Harry are in the lobby, "
```
amymorrill, line 242: I don't know if this matters because this message may not be displayed, but should we change it so it doesn't say "try to talk to them" since we don't have any dialogue prepared? We also don't have a "request a battle" function, so maybe we should say instead: "Fiona and Harry are in the lobby. You can attack them if you go to the arena."

src/npc/examples/battle_move_example_2021.c
```diff
@@ -0,0 +1,339 @@
...
+        return "Error! We need a loaded room to move.\n";
+    }
+
+    move_room(game, lobby);
+    return "You are back to the lobby";
+}
+
+/* Defines a new CLI operation that removes 1 HP from the specified npc if they
+ * are in the room, and their health is greater than their surrender_level */
+char *attack_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
```
amymorrill, line 186: Should we make it so you can only attack in the arena and not the lobby? I didn't think about that functionality before, but it could be cool.

src/npc/examples/battle_move_example_2021.c
```diff
@@ -0,0 +1,339 @@
...
+    move_t *moves2 = create_enemy_moves();
+    add_battle_to_npc(hostile_harry, 5, stats2, moves2, BATTLE_AI_GREEDY,
+                      HOSTILE, 0);
+
+    /* Add items to hostile npc */
+    item_t *potion = item_new("POTION","This is a health potion.",
+                              "This potion will increase your health. Feel"
+			                  " free to take it.");
+    add_item_to_npc(hostile_harry, potion);
+    item_t *elixir = item_new("ELIXIR","This is an elixir.",
```
amymorrill, line 300: I noticed that when hostile_harry is killed, only the potion shows up in the room's items and not the elixir. That may be due to a problem with adding the elixir to hostile_harry here, or a problem with transfer_all_npc_items (called at line 205).

src/npc/examples/battle_move_example_2021.c
```diff
@@ -0,0 +1,339 @@
...
+    return ctx;
+}
+
+
+int main(int argc, char **argv)
+{
+    chiventure_ctx_t *ctx = create_sample_ctx();
+
+    /* Monkeypatch the CLI to add the new operations
+     * (not handled by action management, as that code
```
amymorrill, line 326: Is this left over from last year that action management only supports items? We should remove it if it's not still true.

amymorrill (changes requested):
I think our demo code looks good! Just a couple of comments about small bugs I've noticed / things we could incorporate.

src/npc/examples/battle_move_example_2021.c
amymorrill: @northofsophia Did you notice this problem with your tests of transfer_all_npc_items? If not, then there might be a problem with adding the elixir to hostile_harry.

src/npc/examples/battle_move_example_2021.c
torresszy: I have added comments explaining each of the monkey-patched functions in our example.

src/npc/examples/battle_move_example_2021.c
torresszy: This bug is fixed now, and now we can move between the lobby and arena multiple times without any problem.

src/npc/examples/battle_move_example_2021.c
torresszy: I have removed these comments.

src/npc/examples/battle_move_example_2021.c
torresszy: This update has been made, thanks so much!

src/npc/examples/battle_move_example_2021.c
torresszy: I have added dialogues for Fiona and Harry and now we can talk to them through the `talk to fiona` or `talk to harry` command, and I have changed the message back to "try to talk to them" since it's available now.

torresszy:
A few updates since the initial pull request:
- A few simple dialogues are added to the example and the player is able to talk to both NPCs now. Chiventure already supports talking with NPCs through the `TALK` command, but it only supports single word npc_id, so I changed `friendly_fiona` and `hostile_harry` to `FIONA` and `HARRY`. Now we can type in `talk to fiona` or `talk to harry` in the CLI to start a conversation with them.
- The `segfault` bug when calling `lobby` and `arena` more than once is solved, now we can move between lobby and arena multiple times without a problem
- I have updated the comments explaining each of our monkey-patched functions
- When printing out a list of NPCs in the room to the CLI, it will check their healths and only prints those whose health is above 0
- The `ATTACK` command is only allowed in the arena room now, since that is set to be the designated battle room.

src/npc/examples/battle_move_example_2021.c
northofsophia: I have fixed transfer_all_npc_items so that it properly iterates through the hash table and the demo prints all items correctly.

northofsophia:
The issue with transfer_all_npc_items not properly iterating through all the items in the hash table and the demo not properly printing out all the items has been resolved.

amymorrill (approved):

xlewellen (approved):
Everything looks good!

mbutera17 (approved):
Looks good! Many of the changes are from other pull requests so the main changes are just in the battle_move_example_2021.c file. This file looks good though!

lilyehsani (approved):
Looks great! Can't wait to see this in action!

torresszy:
This is the closing comment for this pull request; it has been merged successfully and all goals have been achieved. One change since the beginning of this pull request is that `npcs_in_room_operation` is not monkey-patched in `battle_move_example_2021.c` anymore since it was added to `operation.h` in PR #1145. Other than that the core features listed in the opening statement for this pull request and the demo remained the same, which include:
- Moving the NPCs between rooms
- Printing out all the NPCs who are alive in the current room
- Having battles with NPCs through the `ATTACK` command
- The NPCs can be set to either fight till death or surrender after a certain threshold.

We also hope this demo, `battle_move_example_2021.c` to be a helpful reference for further implementations and developments on NPC movement and battle features in the future.

lilyehsani:
Issue Score: ✔️++

Comments:
Great work overall! The summary, updates, and closing of this PR are very thorough.

Issue #1039, opened 2021-05-18
Create openworld demo "game" (Spring 2021 team)


Justin08784:
Create a sample "game" that demonstrates the work that the openworld team has done since sprint 1–– level-oriented generation and recursive generation. Since the openworld module is not properly integrated into chiventure, monkeypatching will be necessary. We expect to add more features and do some refactoring in the current sprint so production of the demo game will need to be delayed/adjusted accordingly.

Concurrent issues:
Refactoring: #1033, #1034, 
New: #1032, #1035 

This demo will be utilized in our final presentation. We will create a new branch (openworld/demo-spring-2021) for the demo and merge it into `dev` using a PR.

ahmed-sumaiya:
The three of us met today over zoom and decided to work on having the following features displayed in our demo game:
1. Autogeneration of rooms within a radius (n) where n number of rooms are generated randomly around the current room in each direction (north, south, east and west)
2. Level oriented room generation where a room is generated randomly but according to the player's level
3. Item generation with a random probability of appearing in a certain room at a random quantity

Justin08784:
Discovered a bug where filter_speclist_with_difficulty() removes nodes from the original, unfiltered speclist and inserts them into the filtered speclist output.

Steps for reproduction:
In c62c1cb, I added a test to check filter_speclist_with_difficulty() does not alter the original, unfiltered speclist. Simply compile tests and confirm that test unfiltered_unchanged fails.

Justin08784:
Summary of 487abb2:

- hardcoded rspecs, speclist, context, and levelspec 'placeholder variables' that are initialized and referred to by functions
- introduced initialize_sample_structs() to initialize placeholder variables before game execution

_New/Modified CLI commands_
- LEVELNEW <int>, updates player level to given integer
- NUKE, removes and frees all rooms in the game hash except for the original room, "room1". Resets num_built of all rspecs to 0 as well.
- LEVELCHECK, outputs the current player level and the corresponding difficulty level of rooms that will be generated
- COUNT, outputs the number of rooms (total and per difficulty level)
- Changed LEVELGEN so that it takes an integer token to specify number of rooms to be loaded (4 is maximum because it only generates in the compass directions)

Justin08784:
53d669f:
- renamed openworld_example.c (and executable) to openworld-demo-2021.c
- Bug: discovered that 'nuke' CLI command doesn't reset game->curr_room to original room. Fixed to do so.

[...]

PR #1148, opened 2021-05-29
Created an openworld demo game

wmingyan:
This PR is related to issue #1039 for creating a monkey-patched demo game for openworld that showcases the following features we implemented this quarter:
- Level-oriented room generation, where a room is generated according to the player's level
- Recursive generation of rooms within a radius (n), where n layers of rooms are generated around the pivot room in the specified direction (north, south, east, and west in the demo game)

During the implementation of the demo game, a bug in roomspec_to_room that could cause segfault when calling game_free() is called was discovered and fixed.


src/openworld/examples/openworld_example.c
```diff
@@ -0,0 +1,157 @@
...
+char *level_gen(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
+{
+    /* This operation has to be called without parameter */
+    if(tokens[1] != NULL)
+    {
+        return "I do not know what you mean.";
+    }
+    
+    /* Create speclist */
+    roomspec_t *roomspec0 = roomspec_new("level_0_room", "A level-0 room.", "A room with difficulty level 0.", NULL);
```
Justin08784, line 55: roomspecs, gencontexts, speclists etc. can be hardcoded as global struct variables outside of the main function. level_gen and recursive_gen can simply use and update these variables instead of having to redefine them with every call. This prevents the addition of duplicate rooms and allows us to call level_gen and recursive_gen multiple times.

Justin08784 (changes requested):
Hardcoded gencontexts will enable us to preserve the current player_level between function calls. This means that implementing a "LEVELUP" CLI command to update player level (as we originally intended) should be possible. I will be making these changes.

Justin08784 (approved):
I have addressed my own changes requested above. See [my comment](https://github.com/uchicago-cs/chiventure/issues/1039#issuecomment-851478701) on the issue to get a more detailed summary of changes (I also fixed a [bug](https://github.com/uchicago-cs/chiventure/issues/1039#issuecomment-851188026) for filter_speclist_with_difficulty()). 

Each CLI command can now be called arbitrarily many times. LEVELGEN now takes a num_rooms token/parameter.

ahmed-sumaiya (approved):
Looks good to me!

src/openworld/examples/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+               openworld_example.c)
+
+# Link with chiventure libraries
+foreach(example ${OPENWORLD_EXAMPLES})
+    foreach(module ${CHIVENTURE_MODULES})
+        target_link_libraries(${example} ${module})
+    endforeach(module)
+endforeach(example)
+
+add_custom_target(openworld-examples
```
MaxineK36, line 13: nit: might be better to call this something like `openworld-demo-2021`

src/openworld/src/autogenerate.c
```diff
@@ -54,8 +54,12 @@ room_t* roomspec_to_room(roomspec_t *roomspec)
 
     /* we use buff for the room name instead */
     room_t *res = room_new(buff, roomspec->short_desc, roomspec->long_desc);
-    /* instead of taking all the items, just take a few of them */
-    res->items = random_items(roomspec);
+    
+    item_hash_t *items = random_items(roomspec);
```
MaxineK36, line 58: Can you say more about the change being made here?

src/openworld/src/autogenerate.c
```diff
@@ -261,8 +265,9 @@ speclist_t* filter_speclist_with_difficulty(speclist_t *speclist,
         int is_given_difficulty = roomspec_is_given_difficulty(roomlevels, 
                                                                curr->spec, 
                                                                difficulty_level);
-        if (is_given_difficulty == SUCCESS) { 
-            DL_APPEND(filtered_speclist, curr);    
+        if (is_given_difficulty == SUCCESS) {
+            speclist_t *curr_copy = speclist_new(curr->spec);
```
MaxineK36, line 269: & here?

src/openworld/examples/openworld_example.c
```diff
@@ -0,0 +1,300 @@
...
+
+
+const char *banner = "THIS IS AN OPENWORLD EXAMPLE PROGRAM";
+
+
+/* ==== Placeholder sample struct variables ====
+These will be initialized in initialize_sample_structs() */
+
+/* ROOMSPECS
+   Pre-declaring roomspecs like this is important as we need to preserve
```
MaxineK36, line 31: In a full integration, would these go in the chiventure `ctx` struct?

src/openworld/examples/openworld_example.c
```diff
@@ -0,0 +1,300 @@
...
+These will be initialized in initialize_sample_structs() */
+
+/* ROOMSPECS
+   Pre-declaring roomspecs like this is important as we need to preserve 
+   num_built between function calls in order to generate unique rooms. */
+roomspec_t *rspec_lvl0 = NULL;
+roomspec_t *rspec_lvl1 = NULL;
+roomspec_t *rspec_lvl2 = NULL;
+roomspec_t *rspec_lvl3 = NULL;
+
```
MaxineK36, line 37: Would these go in `ctx` as well?

src/openworld/examples/openworld_example.c
```diff
@@ -0,0 +1,300 @@
...
+    }
+    
+    room_t *orig_room = ctx->game->curr_room;
+
+    /* Remove paths from original room. */
+    delete_all_paths(ctx->game->curr_room->paths); 
+    orig_room->paths = NULL;
+
+    /* Temporarily remove original room from hash. */
+    HASH_DELETE(hh, ctx->game->all_rooms, orig_room);
```
MaxineK36, line 118: Are there existing functions for deleting rooms or does this need to be done manually?

src/openworld/examples/openworld_example.c
```diff
@@ -0,0 +1,300 @@
...
+
+    room_t *curr, *tmp;
+    /* Remove all other rooms from hash. */
+    HASH_ITER(hh, ctx->game->all_rooms, curr, tmp) {
+        HASH_DELETE(hh, ctx->game->all_rooms, curr);
+        room_free(curr);
+    }
+
+    /* Reinsert original room into hash. */
+    add_room_to_game(ctx->game, orig_room);
```
MaxineK36, line 128: Given that you're adding this one back, there may be some `free_rooms` or similar function; if so, please use that

MaxineK36 (changes requested):
This looks great! Just a couple of clarifications/things to change to make this example useful for future integration efforts

src/openworld/src/autogenerate.c
Justin08784: I don't think this change is necessary. Will undo.

src/openworld/src/autogenerate.c
Justin08784: Creates a copy of the speclist node to be added to the filtered speclist (output). Previously it was just yoinking nodes out of the unfiltered speclist (input), which caused some problems.

src/openworld/examples/openworld_example.c
Justin08784: I think so, but am not 100% sure.

src/openworld/examples/openworld_example.c
Justin08784: Yes. But instead of having multiple fields for roomspecs, levelspecs etc. I think it would be best to wrap them under gencontext as a sort of neat package for all generation-related info.

src/openworld/examples/openworld_example.c
Justin08784: Yes, there is `delete_all_rooms(room_hash_t *rooms)`–– found [here](https://github.com/uchicago-cs/chiventure/blob/83f706a60e5d100082cb5438de4c4da4f0997631/src/game-state/src/room.c).

As you can see, this function's code is nearly identical to that of mine. However, if I just call `delete_all_rooms(ctx->game->all_rooms)` it crashes. 

Strangely, if I define an identical function `delete_all_rooms2(room_hash_t **rooms` <note the different parameter type>, it works just fine. Thoughts?

src/openworld/examples/CMakeLists.txt
Justin08784: Should I change the name of the .c file as well?

MaxineK36: Yes

src/openworld/examples/openworld_example.c
MaxineK36: Ok, makes sense! Please note that in a comment in the demo file, and (if you're feeling up to it) it would be awesome to open up a backlog issue for this

src/openworld/examples/openworld_example.c
MaxineK36: Hm, that's very odd. Have you figured out at all why it's happening/where the crash is coming from?

src/openworld/src/autogenerate.c
MaxineK36: Ok, sounds good

src/openworld/examples/openworld_example.c
MaxineK36: If they need to be declared before any open world functions are called, they'd need to be in the context struct, which you should note here.

src/openworld/examples/openworld_example.c
MaxineK36: Ok, makes sense! Please note that in a comment in the demo file, and (if you're feeling up to it) it would be awesome to open up a backlog issue for this

MaxineK36 (commented):
Responded to comments

src/openworld/examples/CMakeLists.txt
Justin08784: UPDATE: 53d669f

Misc.:
While tinkering around, discovered that 'nuke' CLI command doesn't reset `game->curr_room` to original room. Fixed.

src/openworld/examples/openworld_example.c
Justin08784: Yes, it's `delete_all_rooms()`. See 868f13a.

To reproduce bug, 'make' debug2 in the openworld examples directory. The `debug2` program simply add 4 rooms to a hash and runs delete_all_rooms() to delete the rooms.

Running the core delete_all_rooms() results in segfault. Running my modified version frees all rooms without leaks (you can check using `valgrind ./debug2` as well).

In my experience, if we want to define a function that adds or deletes elements from a hash using standard uthash functions, the hash table parameter needs to be dereferenced by 1 (so taking a `struct_hash_t**` instead of `struct_hash_t*`). e.g.
```
add_roomlevel_to_hash(&roomlevels, lvl0_roomname, 0); // makes use of HASH_ADD_STR
int add_item_to_hash(item_hash_t **ht, item_t *new_item); // makes use of HASH_ADD_STR
```
In fact, if there are other functions like delete_all_rooms(), I don't how they are functioning properly.

src/openworld/examples/openworld_example.c
Justin08784: Noted: 0edd6b4

src/openworld/src/autogenerate.c
Justin08784: Noted: b4ca0d9

src/openworld/examples/openworld_example.c
MaxineK36: @Justin08784 which other functions are you referring to?

Justin08784: This is the only other example I am aware of so far. 
```
int delete_all_paths(path_hash_t *hash);
```
Every other function that uses uthash macros directly within its function scope always takes the hash table dereferenced by 1 (`struct_hash_t **`). I've added even more examples to prove my point in the above comment.

MaxineK36: @Justin08784 Ok, thanks for tracking that down!
It looks like this function is only called one other place; if you can switch the `delete_all_rooms` function to match the `delete_all_rooms2` function you wrote, and it allows both your demo and the current chiventure to work without crashing, please go ahead and do that! 

If you encounter issues/it won't work with both, let me know, and in that case you'd leave what you have in your demo and we'd make a bug issue for the backlog so that this can be taken care of in future years.

Justin08784: UPDATE: 82afba1

Changes summarized [here](https://github.com/uchicago-cs/chiventure/issues/1039#issuecomment-852776759).

MaxineK36 (approved):
Looks great!

Justin08784:
Closing statement:
Comments by reviewers have been addressed and branch `openworld/add-itemspec` is now merged into dev. The sample game `openworld-demo-spr21` can now be made and run in the `build/src/openworld/examples` directory. This demo game demonstrates level-oriented generation and recursive generation.

Misc Changes:
- filter_speclist_with_difficulty() no longer alters the unfiltered speclist – [(see post)](https://github.com/uchicago-cs/chiventure/issues/1039#issuecomment-851188026)
- delete_all_rooms() now takes a `room_hash_t**` parameter [(see post)](https://github.com/uchicago-cs/chiventure/issues/1039#issuecomment-852776759)–– similar changes should be made to other functions

MaxineK36:
Issue Score: ✔️++

Comments:
Excellent work! Your links in the closing statement are particularly useful.

PR #1151, opened 2021-05-31
NPC2 game demo bug fixes related to recent merges

jeyeong:
Recent merges (specifically, case insensitivity) resulted in issues within the NPC and dialogue modules. This pull request fixes those issues. Specifically:
- The issue of items not being able to be found from game->all_items
- The issue of NPCs not being able to be talked to
- The issue of dialogue actions resulting in the game crashing
- Certain tests (Criterion assert_eq's) not passing due to remedies to the above issues
- Chiventure was unable to load rooms if there connection formed a cycle.


matiasp-10 (approved):
Looks good to me!

src/game-state/src/game.c
```diff
@@ -237,7 +237,7 @@ room_t *find_room_from_game(game_t *game, char* room_id)
 item_t *get_item_from_game(game_t *game, char *item_id)
 {
     item_t *i;
-    HASH_FIND(hh_all_items, game->all_items, item_id,
+    HASH_FIND(hh_all_items, game->all_items, case_insensitized_string(item_id),
```
pkasner, line 240: The case_insensitized_string() function mallocs a new string and needs to be freed.

src/game-state/src/game.c
```diff
@@ -246,7 +246,7 @@ item_t *get_item_from_game(game_t *game, char *item_id)
 npc_t *get_npc(game_t *game, char *npc_id)
 {
     npc_t *n;
-    HASH_FIND(hh, game->all_npcs, npc_id,
+    HASH_FIND(hh, game->all_npcs, case_insensitized_string(npc_id),
```
pkasner, line 249: Need to be able to free the case_insensitized_string. Alternatively, does this become redundant because we changed the npc's npc_id to be lowercase when we call npc_new?

src/npc/src/dialogue.c
```diff
@@ -615,6 +615,8 @@ node_action_t *node_action_new(node_action_type action, char *action_id)
     node_action_t *n_a;
     if ((n_a = malloc(sizeof(node_action_t))) == NULL) return NULL;
 
+    action_id = case_insensitized_string(action_id);
```
pkasner, line 618: I don't think you're supposed to overwrite a parameter like this. I think it's best to make a new char *action_id_case. Additionally, the output of case_insensitized_string should probably be freed after it is strdup/strcpy'd into the struct

src/npc/src/npc.c
```diff
@@ -34,6 +34,8 @@ npc_t *npc_new(char *npc_id, char *short_desc, char *long_desc,
     npc->class = malloc(sizeof(class_t));
     npc->movement = malloc(sizeof(npc_mov_t));
 
+    npc_id = case_insensitized_string(npc_id);
```
pkasner, line 37: See my comment about action_id

src/npc/src/npc.c
```diff
@@ -136,7 +138,8 @@ item_list_t *get_npc_inv_list(npc_t *npc)
 bool item_in_npc_inventory(npc_t *npc, char *item_id)
 {
     item_t *check;
-    HASH_FIND(hh, npc->inventory, item_id, strnlen(item_id, MAX_ID_LEN), check);
+    HASH_FIND(hh, npc->inventory, case_insensitized_string(item_id),
```
pkasner, line 141: Output of the case function needs to be freed.

pkasner (changes requested):
I was looking at the closed PR for the CLI's case insensitivity issue, and it looks like we made a lot of mistakes similar to what they did initially. Also it's possible that some of the functions that some of the calls to case_insensitize_string() would be better done if we make sure everything's lowercase right from when they're loaded from the WDL. For example if we insensitized the calls to get_item_from_game when loading the npc's inventory from wdl, then our hash find doesn't need to be insensitized every other time chiventure looks for an item.

src/game-state/src/game.c
jeyeong: Changed!

src/game-state/src/game.c
jeyeong: Changed!

jeyeong: The npc_id provided here may not always be lower-case (this was causing a certain test to fail), unless we convert everything to lower-case at the WDL-to-game translation stage (which I think is good to do for the long term), I thought it would be a good safeguard to convert all strings passed into this function into lower-case.

src/npc/src/dialogue.c
jeyeong: Changed!

src/npc/src/npc.c
jeyeong: Changed!

src/npc/src/npc.c
jeyeong: Changed!

jeyeong:
I definitely agree. A more efficient implementation would be to convert everything to lower-case right from the start, i.e. the WDL-to-game translation stage. If you have the time to work on that, please go ahead!

MaDDylan (changes requested):
Great job tracking down the source of the errors! Everything looks good to me, just did a complete run-through of the demo again. Although there was one thing I was wondering, in the mom's dialogue do we want the second quip "Finding wild pokemon!" to be conditional on talking to oak/ after receiving a pokeball?

pkasner (approved):
Everything looks good now!

lilyehsani (approved):
LGTM! Approved.

MaDDylan (approved):

lilyehsani:
Issue Score: ✔️+

Comments:
Great work overall! I have one comment:
- Like issues, PRs require closing comments. When an PR is done, you should include a final comment summarizing what was merged to `dev` and what work remains to be done, if any.

Issue #1088, opened 2021-05-23
Create quests demo with Monkeypatching/running chiventure


ddelgado60:
In the final demo of quests, we hope to:

Show incorporation of quests into chiventure
Demonstrate new achievement tree data structure to hold achievements
Show the different passive and active missions that can make up an achievement tree
Show new reward types and stats requirements for quests

assignees: Gabi, Olivia, Devon
team:quest

oliviajerdee:
After I fixed the errors that were causing the "build chiventure" step of our action to fail, the action now fails on "build examples".  To fix those errors, we should start working on our monkeypatching. Since the old examples ran with the linked list implementation for achievements, this should entail significant refactoring in addition to demonstrating our new features.

gabigarc03:
As detailed in Issue #957 and PR #1080, I refactored the example so that chiventure now compiles.

ddelgado60:
The branch with some addons for mission types, reward types, and stat requirements compiles and the examples runs, we just need to create a battle instance and demonstrate passive missions for leveling up, getting health, or getting xp. PR is here: https://github.com/uchicago-cs/chiventure/pull/1096

oliviajerdee:
I designed the plan for the achievement tree we will use in our demo. We will feature the new tree structure, the new tasks of killing something and entering a room, and the new stat checks in order to start a quest. The conversion from Word to a Google Doc made one of the text boxes a little messed up, but the plan is still clear. Once we make the current examples run without errors, we will start to create this version of the demo. The Google Doc with our design is attached. [https://drive.google.com/file/d/1z2edKntdcca9Ci5Ijeq2ZycP0clCLUX6/view?usp=sharing](url) Depending on whether all branches are merged to dev this design might change to only include the tasks of getting an item and talking to an NPC. For now, I will work in the implement-trees branch and focus on the tasks of getting an item and talking to an NPC, then add the additional tasks after/if the other branch is merged. I will push the changed example code after the merges are resolved for implement-trees.

oliviajerdee:
We've decided to change our tree from the original plan as it seems that the tasks of killing NPCs, entering rooms, and stat checks won't be possible in our demo. After finishing the demo, I will create a new tree diagram to show our design.

lilyehsani:
Issue Score: ✔️+

Comments:
Great work overall! I have two comments:
- When an issue is done, you should include a final comment summarizing what was done, and why the issue is being closed.
[...]

PR #1154, opened 2021-06-03
Quests/implement trees

oliviajerdee:
Fixed a bug in the demo. #1088


zsp255 (approved):

zsp255 (commented):
Our demo works on a previous version of code without the implement-new-quest-fcts being merged. Not sure if the code will run with this in it.

PR #1213, opened 2022-04-26
Dsl: option to disable default properties generation

awliu2:
As of before, the dsl generator would always include default values if room/item/action descriptions, etc were missing. However, it would be beneficial to have the option to disable this default property generation with a flag passed to the main function. It would also be useful to have the option to only generate the required properties, and not generate anything for optional descriptions.

As such, we modified parse.py to look for the flags "--no-defaults" and "--some_defaults." These flags get passed to the classes, which we slightly modified, adding an instance variable that stores the default flag, if any. The implementation of the generate_default methods were changed to support the two new flags.

In the situation where the parser encounters a missing value and does not generate any default properties, a warning is printed to console. 

Everything works as expected. No bugs were experienced in our testing.

Note that enabling --no-defaults may generate wdl files that will not run in chiventure due to missing required descriptions. However this is the intent of adding this function to make it easier for game authors to create WDL files completely under their control, without any generated components.


alexlee0422:
This looks good. It looks like you guys took care of what this features does by successfully running the test and provided detailed documentation and addressed the objective of why this flag was needed to be added.

alexlee0422 (approved):
This looks good. It looks like you guys took care of what this features does by successfully running the test and provided detailed documentation and addressed the objective of why this flag was needed to be added.

src/dsl/src/to_wdl.py
```diff
@@ -187,27 +198,32 @@ def generate_defaults(self):
...
-
-        # generate default interaction text for actions
-        for i in self.wdl_contents.get('actions', []):
-            if 'text_success' not in i:
-                i['text_success'] = f"You {i['action'].lower()} the {self.id}."
-                warn(f'''missing: success text for action {i['action']} for item {self.id}, generated default: {i['text_success']}''')
-            if 'text_fail' not in i:
-                i['text_fail'] = f"You cannot {i['action'].lower()} the {self.id}."
-                warn(f'''missing: failure text for action {i['action']} for item {self.id}, generated default: {i['text_fail']}''')
+        if self.default =="no-defaults":
```
jacknugent1529: Minor style issue: add a space after `==`

src/dsl/src/to_wdl.py
```diff
@@ -187,27 +198,32 @@ def generate_defaults(self):
...
+           warn(f'''warning: no default values generated for item, wdl file may not run''')
+           return
+        
+        if self.default == "some_defaults" or "":
```
jacknugent1529: What is the purpose of `or ""`? To check if `self.default` is `some_defaults` or `""` I think you need to write it as `self.default == "some_defaults" or self.default == ""`.

src/dsl/src/to_wdl.py
```diff
@@ -187,27 +198,32 @@ def generate_defaults(self):
...
+                default = f"This is a {self.id}. {short_desc}"
+                self.wdl_contents['long_desc'] = f"{default}"
+                warn(f'''missing: long description for {self.id}, generated default: {self.wdl_contents['long_desc']}''')
+                    
+            # generate default for short description
+            if 'short_desc' not in self.wdl_contents:
+                self.wdl_contents['short_desc'] = f"{self.id}"
+                warn(f'''missing: short description for {self.id}, generated default: {self.wdl_contents['short_desc']}''')
+        
+        if self.default =="":
```
jacknugent1529: It is more consistent style to add a space after `==`. Also, is there a reason this is written as an if statement rather than if/elif/else?

src/dsl/src/to_wdl.py
```diff
@@ -265,10 +283,13 @@ def generate_defaults(self):
...
         """
-
-        # generate default for introduction
-        if 'intro' not in self.wdl_contents:
-            default = self.wdl_contents.get('start') or "room"
-            self.wdl_contents['intro'] = f"Welcome! You're in a {default}"
-            warn(f'''missing: introduction for game, generated default: {self.wdl_contents['intro']}''')
+        if self.default == "no-defaults":
+            warn(f'''warning: no default values generated for game, wdl file may not run''')
+            return
```
jacknugent1529: The `return` isn't necessary here, and is inconsistent with the else branch, which doesn't have a return

jacknugent1529 (commented):
It looks good overall, and adding the warnings was a nice touch. I commented on a few things to be fixed or clarified. Also, be sure to add the documentation at the beginning of the "main" function describing what your flags do. If there's any confusion about style, see the pep 8 standard (https://peps.python.org/pep-0008/#other-recommendations).

src/dsl/src/to_wdl.py
awliu2: fixed

src/dsl/src/to_wdl.py
awliu2: fixed

src/dsl/src/to_wdl.py
awliu2: Changed to an if/else with a nested conditional statement

awliu2:
Made revisions based on peer reviews, and added default flag description to main function documentation.

src/dsl/src/to_wdl.py
```diff
@@ -91,6 +93,8 @@ def to_wdl_structure(self) -> dict:
             else:
                 self.wdl_contents[k] = v
 
+        if self.default == "no-defaults":
+            warn(f'''warning: no default values generated for {id}, wdl file may not run''')
```
jacknugent1529: This refers to the `id` function in python. I assume you want `self.id`

jacknugent1529 (commented):
The changes look good, however see the comment for another possible issue

src/dsl/src/to_wdl.py
awliu2: Good catch. I think this was from last year because I didn't touch any of this code but I'll go ahead and fix it right now!

awliu2: fixed

jacknugent1529: Ah, well thanks for fixing it anyway!

jacknugent1529 (approved):
Looks good to me!

jacknugent1529:
PR Score: Excellent

Comments: Excellent work! The summary of changes is very detailed and all the minor changes were made quickly.

Issue #1165, opened 2021-06-05
Strengthen quest tests to better test tree functionality


gabigarc03:
The tests for quests (in `test_quests_state.c`) were really only refactored when non-binary achievement trees were implemented. They don't fully test whether `add_achievement_to_tree`, `complete_achievement`, and `is_quest_completed` actually use tree functionality effectively. See the [current implementation wiki](https://github.com/uchicago-cs/chiventure/wiki/Quests-~-Current-Implementation) for more details.

JSerf02:
When updating these tests, we should also add tests against the updated quest id system [(Pull Request #1237)](https://github.com/uchicago-cs/chiventure/pull/1237) to make sure it works

jakeziqiwei:
We have added extra tests to add coverage to the above functions, but new functions we have written(in the ctx branch). Closing this issue now since it has already been merged into dev

philipchenaux:
Issue Score: Excellent

Comments:
Great job! My only comment here is that you did not include an update when you linked the Quests/add ctx PR to this issue. You should explain the significance of this PR in this issue.

PR #1237, opened 2022-05-01
Quests/change id format

### LINKED ISSUE #1158: Convert quest_id from long id to char* and refactor functions accordingly
Quests currently use a long int to store IDs, which can be confusing. String identifiers would be much more user-friendly, hence this change. See the [current implementation wiki](https://github.com/uchicago-cs/chiventure/wiki/Quests-~-Current-Implementation) for more details.

Castel-Jaloux:
Changed the type of quest_id from long int to char* to improve usability. Changed the appropriate values in all functions and function calls, as well as tests, to reflect this alteration.


JSerf02 (approved):
Changes look good and pass all tests, so not really much else to say. We should take note, though, that when we eventually improve our tests [(Issue #1165)](https://github.com/uchicago-cs/chiventure/issues/1165), we should add tests for multiple different ids instead of using the same id for every test.

src/quests/src/quests_state.c
```diff
@@ -171,13 +171,13 @@ int achievement_init(achievement_t *achievement, mission_t *mission, char *id)
...
 /* Refer to quests_state.h */
-int quest_init(quest_t *q, long quest_id, achievement_tree_t *achievement_tree,
+int quest_init(quest_t *q, char *quest_id, achievement_tree_t *achievement_tree,
                 reward_t *reward, stat_req_t *stat_req, int status)
 
 {
     assert(q != NULL);
 
-    q->quest_id = quest_id;
+    q->quest_id = strdup(quest_id);
```
philipchenaux: It is good practice to use the string-safe version of strdup which is strndup.

philipchenaux (changes requested):
Generally looks good, one very minor comment. But if we can get that fixed quickly, then it should likely be good to merge and will count for the second sprint.

src/quests/src/quests_state.c
Castel-Jaloux: The suggested edit has been updated on the branch and pushed. Additionally, some of the edits to quest-example.c that ended up being made in the other pull request were moved here as well for consistency (since this is the pull request that those changes are related to).

philipchenaux (approved):
Looks good to me, happy to merge this into the code base. Just awaiting @JSerf02 approval of this PR.

JSerf02 (approved):
Looks good. Nice job using a #define to store the max number of characters in the id!

philipchenaux:
Issue Score: Excellent

Comments:
Another great pull request, good job!

PR #1268, opened 2022-05-02
Fixing a bug in NPC movement when a specific room is in multiple steps of its path

### LINKED ISSUE #1240: Fix bug with having repeated rooms in an npc's movement path
members: William

bvvill:


bvvill:
> changed reverse_path so that it supports NPCs with both a definite and indefinite movement path. I also edited move_npc_indefinite so that the movement path is reversed once the end is reached, but the move backward is not followed through, staying true to the original documentation of the function in npc_move.h

I decided to to commit these changes to this branch because they are changes I would've made with more time with this issue

bvvill:
summary of changes:

- addition of `unsigned int npc_path_pos` into the `npc_mov_t` struct, which keeps track of the current location of the NPC within its movement path

- a function to get this value, `track_npc_path_pos`, has also been implemented
- addition of `uint8_t npc_path_reversed` into the `npc_mov_t` struct, which is 0 if the movement path is in the original order/direction, or 1 if the path is in the reverse/opposite order/direction
   
- a function to get this value, `track_npc_path_reversed`, has also been implemented
- for NPCs with definite paths, the `reverse_path` function was edited to flip this member between 0 and 1 whenever called.
- `npc_mov_init` was edited to assign the initial values of the newly added members to be 0
- To fix the bug described earlier, functions using the `char *room_id` to search through the movement path (which is in a linked list), such as `mov_npc_definite` and `mov_npc_indefinite`, were edited to go through the linked list using the values of the `npc_path_pos` and `npc_path_reversed` members, and to (in/de)crement an NPC's `npc_path_pos` accordingly.
- `get_npc_num_rooms` now uses the efficient `LL_COUNT` macro instead of using `LL_FOREACH` and explicitly defining the counting

Commentary made during the progress leading to these changes can be found in https://github.com/uchicago-cs/chiventure/issues/1240

emilybae1705 (commented):
npc_path_reversed is tracking whether an NPC is moving along its path in the original or reversed direction where 0 indicates original and 1 indicates reversed direction. To make this idea more concrete, maybe you could make the struct into an enum where ORIGINAL=0, REVERSED=1. This could possibly help avoid confusion down the line for future developers of the NPC movement module as well.

emilybae1705 (commented):
Additionally, in `npc_move.c` _lines 176-180_, `head` is assigned to different `npc_path` values depending on whether the `mov_type` is DEFINITE or INDEFINITE. The original implementation only accounted for the DEFINITE case as an input to the LL_FOREACH_SAFE function. I wasn't sure whether this was intentional, but after talking to @bvvill I've confirmed that he intentionally made this change because the previous implementation did not account for the INDEFINITE case (npcs move to a different room after a certain time interval passes). Accounting for the indefinite case resolves any bug in npc movement where an npc should move to a different location despite having no definite choice of movement made by a player in the CLI.

emilybae1705 (commented):
`LL_COUNT()`'s implementation uses `LL_FOREACH()` and not `LL_FOREACH_SAFE()`, which to my knowledge is a safer version of `LL_FOREACH()` as it accounts for the case when the number of elements in a list changes. Would this be a problem for your `LL_COUNT()` implementation in _line 223_ of `npc_move.c`? I don't know the details of the npc movement module so I'm not entirely sure, but it might be a consideration you need to make?

bvvill:
> `LL_COUNT()`'s implementation uses `LL_FOREACH()` and not `LL_FOREACH_SAFE()`, which to my knowledge is a safer version of `LL_FOREACH()` as it accounts for the case when the number of elements in a list changes. Would this be a problem for your `LL_COUNT()` implementation in _line 223_ of `npc_move.c`? I don't know the details of the npc movement module so I'm not entirely sure, but it might be a consideration you need to make?

`LL_FOREACH_SAFE` is the same as `LL_FOREACH`, it just has an extra parameter in case the pointer to the next step in the linked list struct is named something other than `*next`, which is not the case here, so `LL_COUNT` is fine.

bvvill:
> npc_path_reversed is tracking whether an NPC is moving along its path in the original or reversed direction where 0 indicates original and 1 indicates reversed direction. To make this idea more concrete, maybe you could make the struct into an enum where ORIGINAL=0, REVERSED=1. This could possibly help avoid confusion down the line for future developers of the NPC movement module as well.

I think I agree with using an enum here to avoid confusion in the future. I know other teams in chiventure have a defined range of enums they use so that everyone's not using enums that equal zero or one, but I don't think our team has a defined range yet, I'll ask Lily and then I'll implement an enum for this.

bvvill:
> I know other teams in chiventure have a defined range of enums they use so that everyone's not using enums that equal zero or one, but I don't think our team has a defined range yet, I'll ask Lily and then I'll implement an enum for this.

To correct/clarify myself, other teams making enums for use within the `tid_t` struct were given a range of values to work with so there wasn't any overlap, but that is not a concern here. Also, an enum (`npc_path_direction_t`) for npc path direction has been implemented, as discussed above.

emilybae1705 (approved):
I have reviewed the pull request and it looks good! @bvvill has incorporated the feedback I gave previously and all the necessary changes have been made. Overall, I see no error in the code that would cause the program to crash. Moreover, changing the npc_path_direction_t struct into an enum has made it easier to follow the implementation details. Looks like the request is ready to go!

src/npc/src/npc_move.c
```diff
@@ -154,22 +156,36 @@ char* track_room(npc_mov_t *npc_mov)
 
 
 /* See npc_move.h */
-int reverse_path(npc_mov_t *npc_mov)
+unsigned int track_npc_path_pos(npc_mov_t *npc_mov)
```
lilyehsani: I think it would fit chiventure conventions more if these were named `get_npc_path_pos` and `get_npc_path_direction`.

src/npc/src/npc_move.c
```diff
@@ -154,22 +156,36 @@ char* track_room(npc_mov_t *npc_mov)
...
 
+/* See npc_move.h */
+int flip_npc_path_direction(npc_mov_t *npc_mov)
+{
+    room_list_t *head = NULL, *reversed_path_head = NULL;
     room_list_t *room_elt, *room_tmp;
-    LL_FOREACH_SAFE(npc_mov->npc_mov_type.npc_mov_definite->npc_path,
-                room_elt,room_tmp)
+
+    if (npc_mov->mov_type == NPC_MOV_DEFINITE) {
```
lilyehsani: Need to be consistent with bracket placement: chiventure uses [Allman style](https://uchicago-cs.github.io/dev-guide/style_guide_c.html#indentation).

src/npc/src/npc_move.c
```diff
@@ -179,36 +195,37 @@ int reverse_path(npc_mov_t *npc_mov)
...
 
-	if(npc_mov->mov_type == NPC_MOV_DEFINITE)
-	{
-        LL_FOREACH(npc_mov->npc_mov_type.npc_mov_definite->npc_path, elt)
-        {
-            count++;
-            curr_room = elt->room;
-        }
+	if (npc_mov->mov_type == NPC_MOV_DEFINITE) {
+        LL_COUNT(npc_mov->npc_mov_type.npc_mov_definite->npc_path, elt, count);
```
lilyehsani, line 240: Nice use of this macro!

src/npc/src/npc_move.c
```diff
@@ -241,6 +271,7 @@ int move_npc_definite(npc_mov_t *npc_mov)
     {
         room_t *next_room = current_room->next->room;
         npc_mov->track = next_room->room_id;
+        (direction)? npc_mov->npc_path_pos-- : npc_mov->npc_path_pos++;
```
lilyehsani: The question mark operator is cool, but for readability, I think we should expand this to a case or if statement checking the enum.

src/npc/src/npc_move.c
```diff
@@ -252,25 +283,39 @@ int move_npc_definite(npc_mov_t *npc_mov)
...
     if(current_room->next == NULL)
     {
+        assert(flip_npc_path_direction(npc_mov) == SUCCESS);
         return 1;
     }
     if((strcmp(current_room->room->room_id,npc_mov->track)) == 0)
     {
         room_t *next_room = current_room->next->room;
         npc_mov->track = next_room->room_id;
+        (direction)? npc_mov->npc_path_pos-- : npc_mov->npc_path_pos++;
```
lilyehsani: Same as above.

lilyehsani (changes requested):
Looks really good! Just a few comments, mostly stylistic.

src/npc/src/npc_move.c
bvvill: Yes I agree

src/npc/src/npc_move.c
bvvill: I used the astyle command for this lol I didn't realize it would change so many files, sorry

src/npc/src/npc_move.c
bvvill: thank u !

src/npc/src/npc_move.c
bvvill: Yes I agree

lilyehsani (approved):
Looks good to me! Approved.

bvvill:
summary of changes:

* addition of `unsigned int npc_path_pos` into the `npc_mov_t` struct, which keeps track of the current location of the NPC within its movement path
* a function to get this value, `get_npc_path_pos`, has also been implemented
* addition of `uint8_t npc_path_direction` into the `npc_mov_t` struct, which is `enum NPC_MOV_ORIGINAL` if the movement path is in the original order/direction, or `enum NPC_MOV_REVERSED` if the path is in the reverse/opposite order/direction
* a function to get this value, `get_npc_path_direction`, has also been implemented
* for NPCs with definite paths, the `flip_npc_path_direction` function was edited to flip this member between 0 and 1 whenever called.
* `npc_mov_init` was edited to assign the initial values of the newly added members to be 0
* To fix the bug described earlier, functions using the `char *room_id` to search through the movement path (which is in a linked list), such as `mov_npc_definite` and `mov_npc_indefinite`, were edited to go through the linked list using the values of the `npc_path_pos` and `npc_path_direction` members, and to (in/de)crement an NPC's `npc_path_pos` accordingly.
* `get_npc_num_rooms` now uses the efficient `LL_COUNT` macro instead of using `LL_FOREACH` and explicitly defining the counting

Commentary made during the progress leading to these changes can be found in #1240 and in this PR

lilyehsani:
Issue Score: Excellent

Comments:
Great work! The summary and updates are very clear and thorough. When you create a PR, you are requesting a review of the code in the branch you want to merge. As such, you should not push any commits to the branch right after creating the PR; you should only do so in response to a code review. Make sure that, when you create a PR, the code is ready for review and passes all the tests.

PR #1304, opened 2022-05-06
RPG Battles stat_t changes

### LINKED ISSUE #1275: Use design document to implement specialized stats in the battle module
Our task is to implement our new specialized stats. This includes adding magical attack/defense, critical damage, accuracy, and skill points. The full struct and functions changes are outlined in this [document](https://github.com/uchicago-cs/chiventure/issues/1233). This will look like not only changing our struct but modifying other areas of the code to accommodate and use these new stats.

Members: Nicholas, Fatimah

nicholaswlee:
We have created this pull request to create our new struct, stat_t. We have added a couple of new fields for this data structure:
`typedef struct stat {`

    int max_hp; //The maximum amount of health this unit can currently have.

    int hp; //The current health of the unit; the battle ends when this hits 0.

    int max_sp; //The maximum amount of skill/special points this unit can currently have.

    int sp; //The current amount of skill/special points of the player; magic and skills deplete this resource.

    int phys_atk; //The physical damage capacity of the unit; used in damage calculation against phys_def.

    int mag_atk; //The magical damage capacity of the unit; used in damage calculation against mag_def.

    int phys_def; //The physical damage resistance of the unit; calculated against phys_atk.

    int mag_def; //The magical damage resistance of the unit; calculated against mag_atk;

    int speed; //How quickly the unit moves; used to determine turn order, with faster units going first.

    int crit; //The default odds of this unit getting bonus damage; the value is out of 100.

    int accuracy; //The default odds of any move used by this unit hitting the target; the value is out of 100.

    int xp; //The current amount of experience points gained by this unit.

    int level; //The current level of the unit.

`} stat_t;`
We have also edited our damage function to accommodate for critical rate in battle_ai.c and also made sure that all of the code in the code base is now aligned with stat_t.c. Additionally we have temporarily commented out all of the battle examples as we will refine them once move_t and battle_items_t have been made. To add on, we have changed our tests to work with the new stat_t.c struct. A couple of the NPC tests/examples had to be changed to work with stat_t. A new issue will have to be created to create further testing, but that should be done once move_t has become better. Information about our [implementation](https://github.com/uchicago-cs/chiventure/wiki/Battles-~-Stat-Changes-Design-Document) and [work](https://github.com/uchicago-cs/chiventure/issues/1275) can be found with these links.

### DIFF
--- a/include/battle/battle_flow.h
+++ b/include/battle/battle_flow.h
@@ -77,7 +77,7 @@ battle_t *set_battle(battle_player_t *ctx_player, npc_t *npc_enemy,
  * - accuracy : the accuracy of the user using the move
  * returns: the 1 or 0 depending on if the move hits
  */
-int calculate_accuracy(int accuracy)
+int calculate_accuracy(int accuracy);
 
 /*
  * Carries out one iteration of the battle flow loop when a move is used
--- a/include/battle/battle_flow.h
+++ b/include/battle/battle_flow.h
@@ -72,7 +72,7 @@ battle_t *set_battle(battle_player_t *ctx_player, npc_t *npc_enemy,
                       environment_t env);
 
 /*
-* Calculates Accuracy of a Move
+* Determines whether a move hits based on accuracy stat.
  * Parameters:
  * - accuracy : the accuracy of the user using the move
  * returns: the 1 or 0 depending on if the move hits
--- a/include/battle/battle_flow.h
+++ b/include/battle/battle_flow.h
@@ -72,6 +72,14 @@ battle_t *set_battle(battle_player_t *ctx_player, npc_t *npc_enemy,
                       environment_t env);
 
 /*
+* Calculates Accuracy of a Move
+ * Parameters:
+ * - accuracy : the accuracy of the user using the move
+ * returns: the 1 or 0 depending on if the move hits
+ */
+int calculate_accuracy(int accuracy)
+
+/*
  * Carries out one iteration of the battle flow loop when a move is used
  *     This includes:
  *         - receiving battle_player's move
@@ -87,6 +95,7 @@ battle_t *set_battle(battle_player_t *ctx_player, npc_t *npc_enemy,
  *  - A string consisting of the output from the turn
  */
 char *battle_flow_move(battle_ctx_t *ctx, move_t *move, char *target);
+
 
 /*
  * Carries out one iteration of the battle flow loop when an item is used
--- a/include/battle/battle_structs.h
+++ b/include/battle/battle_structs.h
@@ -58,9 +58,14 @@ typedef struct stat {
 /* stat changes stub */
 typedef struct stat_changes {
     int speed;
-    int defense;
-    int strength;
-    int dexterity;
+    int max_sp;
+    int sp;
+    int phys_atk;
+    int mag_atk;
+    int phys_def;
+    int mag_def;
+    int crit;
+    int accuracy;
     int hp;
     int max_hp;
     int turns_left;
--- a/include/battle/battle_structs.h
+++ b/include/battle/battle_structs.h
@@ -40,10 +40,17 @@ typedef struct move {
 
 /* stats stub */
 typedef struct stat {
+    int hp;
+    int max_hp;
+    int max_sp;
+    int sp;
     int speed;
-    int defense;
-    int strength;
-    int dexterity;
+    int phys_atk;
+    int mag_atk;
+    int phys_def;
+    int mag_def;
+    int crit;
+    int accuracy;
     int hp;
     int max_hp;
     int xp;
--- a/include/battle/battle_structs.h
+++ b/include/battle/battle_structs.h
@@ -40,8 +40,6 @@ typedef struct move {
 
 /* stats stub */
 typedef struct stat {
-    int hp;
-    int max_hp;
     int max_sp;
     int sp;
     int speed;
--- a/src/battle/examples/battle_cli_example.c
+++ b/src/battle/examples/battle_cli_example.c
@@ -20,21 +20,23 @@
 #include "battle/battle_logic.h"
 #include "battle/battle_default_objects.h"
 #include <time.h>
-
+/*
 const char *banner = "BATTLE MODE DEMO";
 
 class_t *make_bard2()
 {
     return class_new("Bard", "Cool", "Super Duper and Awesome", NULL, NULL, NULL);
 }
-
+*/
 /* initializes a dummy wizard class */
+/*
 class_t *make_wizard2()
 {
     return class_new("Wizard", "Wise", "Old and wise", NULL, NULL, NULL);
 }
-
+*/
 /* Defines an CLI operation for starting a fight */
+/*
 char *fight_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 {
     //int rc;
@@ -86,9 +88,10 @@ char *fight_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
     battle_ctx->game->player = p;
 
     int add_battle_ctx = add_battle_ctx_to_game(ctx->game, battle_ctx);
-
+*/
     /* start_battle begins the battle by finalizing 
        all finishing touches for a battle to begin */
+/*
     int rc = start_battle(battle_ctx, e, ENV_GRASS);
 
     // prints the beginning of the battle 
@@ -106,36 +109,40 @@ char *fight_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
 
     return start;
 }
-
+*/
 /*
  * Creates a chiventure context with a sample game.
  *
  * Returns: a chiventure context with 
  */
+ /*
 chiventure_ctx_t *create_sample_ctx()
 {
     game_t *game = game_new("Welcome to the Battle CLI Integration Demo for Chiventure!");
     room_t *room1 = room_new("room1", "This is The Room", "You are in The Room. You'll will fight a Goblin in The Room.");
     add_room_to_game(game, room1);
     game->curr_room = room1;
-
+*/
     /* Create context */
+/*
     chiventure_ctx_t *ctx = chiventure_ctx_new(game);
 
     return ctx;
 }
-
+*/
 int main(int argc, char **argv)
 {
+    /*
     chiventure_ctx_t *ctx = create_sample_ctx();
-
+*/
     /* Monkeypatching in a fight action to support dialogue */
-    add_entry("FIGHT", fight_operation, NULL, ctx->cli_ctx->table);
+
+    //add_entry("FIGHT", fight_operation, NULL, ctx->cli_ctx->table);
 
     /* Start chiventure */
-    start_ui(ctx, banner);
+    //start_ui(ctx, banner);
 
-    game_free(ctx->game);
+    //game_free(ctx->game);
 
     return 0;
 }
--- a/src/battle/src/battle_ai.c
+++ b/src/battle/src/battle_ai.c
@@ -81,9 +81,13 @@ move_t* find_greedy(combatant_t* player, combatant_t* enemy)
 double calculate_crit(int crit_chance)
 {
     int chance = randnum(1, 100);
-    if(chance <= crit_chance){
+
+    if (chance <= crit_chance)
+    {
         return 1.5;
-    }else{
+    }
+    else
+    {
         return 1;
     }
 }
@@ -95,12 +99,15 @@ double damage(combatant_t* target, move_t* move, combatant_t* source)
     stat_t* src_stats = source->stats;
     stat_t* tar_stats = target->stats;
     
-    if(1){
+    if (1)
+    {
         tar_defense = (double) tar_stats->phys_def;
         power = (double) move->damage;
         src_strength = (double) src_stats->phys_atk;
         src_level = (double) src_stats->level;
-    }else{
+    }
+    else
+    {
         tar_defense = (double) tar_stats->mag_def;
         power = (double) move->damage;
         src_strength = (double) src_stats->mag_atk;
--- a/src/battle/src/battle_ai.c
+++ b/src/battle/src/battle_ai.c
@@ -78,7 +78,7 @@ move_t* find_greedy(combatant_t* player, combatant_t* enemy)
  * - crit_chance : the crit chance of the user using the move
  * returns: the critical damage multiplier
  */
-double calculate_crit(double crit_chance)
+double calculate_crit(int crit_chance)
 {
     int chance = randnum(0, 100);
     if(chance <= crit_chance){
--- a/src/battle/src/battle_ai.c
+++ b/src/battle/src/battle_ai.c
@@ -72,6 +72,21 @@ move_t* find_greedy(combatant_t* player, combatant_t* enemy)
     }
     return strongest_move;
 }
+/*
+* Calculates Critical Damage
+ * Parameters:
+ * - crit_chance : the crit chance of the user using the move
+ * returns: the critical damage multiplier
+ */
+double calculate_crit(double crit_chance)
+{
+    int chance = randnum(0, 100);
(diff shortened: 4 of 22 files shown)


include/battle/battle_flow.h
```diff
@@ -71,6 +71,14 @@ combatant_t *set_enemy(npc_t *npc_enemy);
...
                       environment_t env);
 
+/*
+* Calculates Accuracy of a Move
+ * Parameters:
+ * - accuracy : the accuracy of the user using the move
+ * returns: the 1 or 0 depending on if the move hits
+ */
+int calculate_accuracy(int accuracy);
+
```
aliu65803, line 81: I am a bit confused about this function. Is it calculating the accuracy or is it calculating whether the move will hit based on the accuracy of the user? If the second one, perhaps a better-suited description and name would be "Determines whether a move hits based on accuracy stat" and "hits_target".

include/battle/battle_structs.h
```diff
@@ -40,10 +40,15 @@ typedef struct move {
...
-    int dexterity;
+    int phys_atk;
+    int mag_atk;
+    int phys_def;
+    int mag_def;
+    int crit;
+    int accuracy;
     int hp;
     int max_hp;
     int xp;
```
aliu65803, line 54: I noticed that this does not include the int level that was written in the stat changes design document. What motivated this change?

src/battle/examples/battle_cli_example.c
```diff
@@ -106,36 +109,40 @@ char *fight_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
 
     /* Start chiventure */
-    start_ui(ctx, banner);
+    //start_ui(ctx, banner);
 
-    game_free(ctx->game);
+    //game_free(ctx->game);
 
     return 0;
 }
```
aliu65803, line 148: Is it necessary to code all this out for the example to run?

include/battle/battle_structs.h
nicholaswlee: It is still in it you just need to scroll down. It is on line 55.

src/battle/examples/battle_cli_example.c
nicholaswlee: Yes it is because we will be making a new example at the very end of the sprints anyway.

include/battle/battle_flow.h
nicholaswlee: Given the user's accuracy this function will return 1 or 0 indicating if the move hits or misses. This can be changed later when move_t makes changes to accuracy.

include/battle/battle_structs.h
aliu65803: I see, thank you. It was not shown until I pressed something.

src/battle/src/battle_ai.c
```diff
@@ -72,24 +72,50 @@ move_t* find_greedy(combatant_t* player, combatant_t* enemy)
...
+ */
+double calculate_crit(int crit_chance)
+{
+    int chance = randnum(1, 100);
+    if(chance <= crit_chance){
+        return 1.5;
+    }else{
+        return 1;
+    }
+}
```
aliu65803, line 93: Will the critical damage always be 1.5 times the normal damage? Perhaps there should be a way to allow the game designer to choose how much damage a critical attack does/a way for them to change this percentage each time a move is critical.

src/battle/src/battle_ai.c
```diff
@@ -72,24 +72,50 @@ move_t* find_greedy(combatant_t* player, combatant_t* enemy)
...
 
 /* See battle_ai.h */
 double damage(combatant_t* target, move_t* move, combatant_t* source)
 {
-    double dmg, power, src_strength, tar_defense, src_level;
+    double dmg, power, src_strength, tar_defense, src_level, crit_boost;
     stat_t* src_stats = source->stats;
     stat_t* tar_stats = target->stats;
     
+    if(1){
```
aliu65803: Wouldn't this make it so that this is always true?

src/battle/src/battle_flow.c
```diff
@@ -219,9 +229,16 @@ char *enemy_make_move(battle_ctx_t *ctx)
...
-        string = print_battle_move(b, b->turn, enemy_move);
+        /* Calculates to see if the move will miss */
+        if(!calculate_accuracy(b->enemy->stats->accuracy)){
+            dmg = 0;
+            b->player->stats->hp -= dmg;
+            string = print_battle_miss(b, b->turn, enemy_move);
+        }else{
+            dmg = damage(b->player, enemy_move, b->enemy);
+            b->player->stats->hp -= dmg;
+            string = print_battle_move(b, b->turn, enemy_move);
```
aliu65803, line 241: A minor suggestion, but perhaps you can create a helper function for this that will take in who's turn it is and do this calculation based on that. This way, you don't need two blocks of code that are pretty similar.

src/battle/src/battle_print.c
```diff
@@ -81,6 +81,32 @@ char *print_battle_move(battle_t *b, turn_t turn, move_t *move)
     return string;
 }
 
+/* see battle_print.h */
+char *print_battle_miss(battle_t *b, turn_t turn, move_t *move)
+{
+    char *move_name = move->info;
+    char* enemy_name = b->enemy->name;
+    char* combatant_name;
```
aliu65803: This is a super small detail but I believe these two should be char *

src/npc/examples/battle_move_example_2021.c
```diff
@@ -95,9 +95,10 @@ stat_t *create_enemy_stats()
 	stat_t *test_stats = calloc(1, sizeof(stat_t));
 
 	test_stats->speed = 50;
-	test_stats->defense = 20;
-	test_stats->strength = 150;
-	test_stats->dexterity = 10;
+	test_stats->phys_def = 20;
+	test_stats->phys_atk = 150;
+	test_stats->accuracy = 100;
+    test_stats->crit = 0;
```
aliu65803, line 101: Indentation needed.

tests/battle/test_battle_print.c
```diff
@@ -114,22 +118,126 @@ Test(battle_print, print_player_move)
...
+    cr_assert_not_null(string, "print_start_battle() failed");
+    
+    char *expected_string = "You used Punch! It did 14 damage.\n"
+                            "-- Your HP: 50\n"
+                            "-- bob's HP: 16\n";
+
+    cr_expect_str_eq(string, expected_string, "print_player_move_crit() failed to set string %s\n. We got %s", string, expected_string);
+
+    free(string);
+}
```
aliu65803, line 169: Since the only difference between this test and the previous is that the second one has a move that is critical, would it be easier to make a helper that runs through a test of print battle move and make two tests, one with a player with an extremely low crit rate and one with a player with an extremely high crit rate? This would make it a bit like the homework, where we used a helper to test the functions such as on_segment and created various tests that just passed in different values.

aliu65803 (approved):
Overall, I believe that the code looks great. My comments are only to help the clarity of the code in parts that I felt were confusing or to point out small typos. There were no mistakes that I found that I thought would hinder whether the code would run, and the code is implemented in a way that is easy to follow. One small comment I have that is not in the other comments is for the design document for the stat changes, which says that the battle ends when the hp hits 0; however, there are ways for enemies to surrender once their hp is low enough, so it might be helpful to clarify this in the design document. Overall, I will approve this pull request because I think it runs well and is mostly easy to follow.

src/battle/src/battle_ai.c
```diff
@@ -72,24 +72,50 @@ move_t* find_greedy(combatant_t* player, combatant_t* enemy)
...
+/*
+* Calculates Critical Damage
+ * Parameters:
+ * - crit_chance : the crit chance of the user using the move
+ * returns: the critical damage multiplier
+ */
+double calculate_crit(int crit_chance)
+{
+    int chance = randnum(1, 100);
+    if(chance <= crit_chance){
```
lilyehsani: Need to add some [spaces](https://uchicago-cs.github.io/dev-guide/style_guide_c.html#places-to-use-spaces) here, and switch to [Allman style](https://uchicago-cs.github.io/dev-guide/style_guide_c.html#indentation) for the braces (throughout).

src/battle/src/battle_default_objects.c
```diff
@@ -38,23 +38,67 @@ battle_item_t *get_random_default_weapon()
...
+                                     
     rv_weapon->battle = true;
     rv_weapon->attack = attack_array[rand - 1];
     rv_weapon->defense = defense_array[rand - 1];
     rv_weapon->hp = hp_array[rand - 1];
     rv_weapon->next = NULL;
     rv_weapon->prev = NULL;
+    return rv_weapon;
+
+/*  --BATTLE ITEM TEST CHANGES TO BE IMPLEMENTED WITH BATTLE ITEM CHANGES
```
lilyehsani, line 58: Sounds good. Make sure you've written the issue for this and placed it in the backlog. Same goes for the other examples.

src/battle/src/battle_flow.c
```diff
@@ -113,9 +113,19 @@ char *battle_flow_move(battle_ctx_t *ctx, move_t *move, char* target)
     /* additionally, a check must be performed here to see if player has
        this move, currently not implemented, waiting for player class
        to resolve move_lists() */
-    int dmg = damage(b->enemy, move, b->player);
-    enemy->stats->hp -= dmg;
-    char *string = print_battle_move(b, b->turn, move);
+    int dmg;
+    char *string;
+    /* Calculates to see if the move will miss */
+    if(!calculate_accuracy(b->player->stats->accuracy)){
```
lilyehsani: Add spaces for style (guide linked previously).

src/battle/src/battle_flow.c
```diff
@@ -113,9 +113,19 @@ char *battle_flow_move(battle_ctx_t *ctx, move_t *move, char* target)
...
+        dmg = 0;
+        enemy->stats->hp -= dmg;
```
lilyehsani: This line should not be necessary since `dmg` is set to 0 just before. In fact, it looks like that line is likely not necessary either.

src/battle/src/battle_logic.c
```diff
@@ -86,9 +86,26 @@ battle_item_t *find_battle_item(battle_item_t *inventory, char *input)
 /* see battle_logic.h */
 int consume_battle_item(combatant_t *c, battle_item_t *item)
 {
-    c->stats->hp += item->hp;
-    c->stats->strength += item->attack;
-    c->stats->defense += item->defense;
+    if((c->stats->hp + item->hp) > c->stats->max_hp){
```
lilyehsani: Add spaces for style (guide linked previously).

src/battle/src/battle_print.c
```diff
@@ -81,6 +81,32 @@ char *print_battle_move(battle_t *b, turn_t turn, move_t *move)
...
+
+    if (turn == PLAYER)
+    {
+        combatant_name = "You";
+    } else
```
lilyehsani: `else` should be on its own line.

src/battle/src/battle_print.c
```diff
@@ -81,6 +81,32 @@ char *print_battle_move(battle_t *b, turn_t turn, move_t *move)
...
+    {
+        combatant_name = enemy_name;
+    }
+
+    char *string = calloc(BATTLE_BUFFER_SIZE + 1, sizeof(char));
+
+    snprintf(string, BATTLE_BUFFER_SIZE, "%s used %s! It missed!\n",
```
lilyehsani: This is nit-picky but the phrasing here could potentially be better. How about `"%s used %s, but it missed!\n"` or `"%s tried to use %s, but it missed!\n"`?

lilyehsani (changes requested):
Looks good so far! I left handful of comments, mostly stylistic. I also agree with all of Angela's points, so please address them.

include/battle/battle_structs.h
```diff
@@ -40,10 +40,15 @@ typedef struct move {
 
 /* stats stub */
 typedef struct stat {
```
emilybae1705, line 42: In our most recent meeting, we agreed to include the npc `surrender_level` in the stat_t struct instead of having it as a separate attribute in `npc_battle_t`. Since currently normal players don't have a surrender_level attribute, how would we incorporate surrender_level in stat_t? We could 1) set surrender_level to 0 for players; 2) implement a surrender_level for players, which would require another set of design documentation; 3) have a separate `stat_t` struct for npcs, `npc_stat_t`. I think (1) would be the easiest to adopt but what do you think? We could also just talk this over a really quick meeting if that's better.

emilybae1705 (approved):
The changes made in `stat_t` looks good and I approve of the pull request! I left a comment about incorporating surrender_level in stat_t so take a look and let me know what you think. On a side note, I think the `battle_over` function would be a suitable place to incorporate surrender_level so if an NPC's health level goes below the surrender_level, the battle ends in favor of the player. We can discuss this in more detail in a future meeting.

src/battle/src/battle_ai.c
nicholaswlee: Hi Lily, I think you provided the wrong style link.

src/battle/src/battle_ai.c
nicholaswlee: Yes the point of this is because there is no current "magical" moves.

lilyehsani: Could you add a comment to the code explaining this/what needs to be done when there are magical moves?

src/battle/src/battle_ai.c
lilyehsani: Oops, updated.

src/battle/src/battle_ai.c
nicholaswlee: Right now we have left it as 1.5, but we can discuss this in a later implementation.

tests/battle/test_battle_print.c
nicholaswlee: Having seperate tests makes it clear if there is an issue with a Crit or not a crit.

src/battle/src/battle_ai.c
nicholaswlee: I have fixed the spacing issue and {}

src/battle/src/battle_flow.c
nicholaswlee: I have removed both lines and just called the print statement in this if block.

src/battle/src/battle_flow.c
nicholaswlee: I have fixed the spacing issue and {}

src/battle/src/battle_print.c
nicholaswlee: I have changed the message to ", but it missed!"

src/battle/src/battle_flow.c
nicholaswlee: This will be changed anyway because of move_t, so we can resolve this then.

src/battle/src/battle_ai.c
nicholaswlee: Yes will do now

lilyehsani (approved):
Looks good to me! Approved.

nicholaswlee:
We have successfully merged this PR and it is closed. We have implemented tests and commented out all of our examples and have closed this issue. These changes have allowed battles to become more customizable, including have magical attack/defense, critical damage, accuracy, and skill points! Our current branch passes all the [checks](https://github.com/uchicago-cs/chiventure/actions/runs/2283397844)! We did have to change some NPC tests and examples, but those were only 2-4 lines that is just a field rename. The full outline of this implementation is in this [doc](https://github.com/uchicago-cs/chiventure/wiki/Battles-~-Stat-Changes-Design-Document).

lilyehsani:
Issue Score: Excellent

Comments:
Great work on this PR! One comment:
1. The title of an issue (or PR) has to convey what the issue (or PR) will be producing. In the future, try using simple titles with a "Verb-Object" structure that clearly states what you will be doing in the issue (e.g., “Design and implement a ‘circle’ module”, “Research libraries for drawing simple geometric shapes”, etc.)

PR #1332, opened 2022-05-09
Quest/hash test

### LINKED ISSUE #1259: Write tests for all quest hash table functions
Currently, none of the hash table functions have tests written for them. We need to write tests for get_quest_from_hash, and add_quest_to_hash

jakeziqiwei:
There are two main things for this pull request

- Fixed implementation for add_quest_to_hash and get_quest_from_hash since they did not work previously
- Added tests to strengthen our coverage for some of the hash functions.

### DIFF
--- a/include/quests/quests_structs.h
+++ b/include/quests/quests_structs.h
@@ -130,12 +130,12 @@ typedef struct stat_req {
  *          2: quest has been completed
  */
 typedef struct quest  {
-    UT_hash_handle hh;
     char *quest_id;
     task_tree_t *task_tree;
     reward_t *reward;
     stat_req_t *stat_req;
     int status;  
+    UT_hash_handle hh;
 } quest_t;
 
 /*
--- a/tests/quests/test_quests_state.c
+++ b/tests/quests/test_quests_state.c
@@ -608,7 +608,7 @@ Test(quest,complete_quest2)
 
 
 
-/*see if get_quest_from_hash when there is quest */ 
+/*see if get_quest_from_hash works when there is quest in Hash */ 
 Test(quest,get_quest1)
 { 
     int xp = 50;
@@ -632,11 +632,11 @@ Test(quest,get_quest1)
     int add_quest2 = add_quest_to_hash(quest2, &test_hash_table);
 
     quest_t *answer = get_quest_from_hash(quest1_id,test_hash_table); 
-    cr_assert_eq(answer, quest1, "failed");
+    cr_assert_eq(answer, quest1, "get_queset() did not return the right quest");
 
 }
 
-/*see if get_quest_from_hash when there is no quest */ 
+/*see if get_quest_from_hash work when there is no quest in the hash */ 
 Test(quest,get_quest2)
 { 
     int xp = 50;
@@ -660,10 +660,10 @@ Test(quest,get_quest2)
     int add_quest2 = add_quest_to_hash(quest2, &test_hash_table);
 
     quest_t *answer = get_quest_from_hash("beeppop",test_hash_table); 
-    cr_assert_eq(answer, NULL, "failed");
+    cr_assert_eq(answer, NULL, "There is an quest with ID of beeppop ");
 }
 
-/*test for add quest when there is no same quest_id */ 
+/*test for add quest when 2 unique quest_ID */ 
 Test(test, add_quest_test1)
 {
     int xp = 50;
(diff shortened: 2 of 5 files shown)


jakeziqiwei:
Finally fixed an error where it works on our computer, but not on github. This was the reasons why there were so many git pushes

include/quests/quests_structs.h
```diff
@@ -130,12 +130,12 @@ typedef struct stat_req {
  *          2: quest has been completed
  */
 typedef struct quest  {
-    UT_hash_handle hh;
     char *quest_id;
     task_tree_t *task_tree;
     reward_t *reward;
     stat_req_t *stat_req;
     int status;  
+    UT_hash_handle hh;
```
charlieweiler, line 138: Not sure if you changed the order of initialization in the quest struct to allocate memory more efficiently - if so, clever!

tests/quests/test_quests_state.c
```diff
@@ -585,3 +585,131 @@ Test(quest,complete_quest)
...
+
+    reward_t *res = complete_quest(quest);
+
+    cr_assert_eq(res, NULL,"quest_completed failed to reward the item");
+}
+
+
+
+/*see if get_quest_from_hash when there is quest */ 
+Test(quest,get_quest1)
```
charlieweiler, line 612: I think it'd be better to include more descriptive test names - for example, get_quest1 could be get_quest_is_quest, get_quest2 could be get_quest_no_quest. That way if the tests fail down the line, we can know what's wrong without diving into the code base.

tests/quests/test_quests_state.c
```diff
@@ -585,3 +585,131 @@ Test(quest,complete_quest)
...
+    quest_hash_t *test_hash_table = NULL;
+
+    int add_quest1 = add_quest_to_hash(quest1, &test_hash_table);
+    int add_quest2 = add_quest_to_hash(quest2, &test_hash_table);
+
+    quest_t *answer = get_quest_from_hash("beeppop",test_hash_table); 
+    cr_assert_eq(answer, NULL, "failed");
+}
+
+/*test for add quest when there is no same quest_id */
```
charlieweiler: add_quest_test1 and add_quest_test2 appear, from the comments, to do the same thing, as each comment says:

/*test for add quest when there is no same quest_id*/

However, upon closer inspection, they seem to do different things. add_quest_test1 is adding two different quests to the hash and should succeed in order to pass the test. add_quest_test2 is adding the same quest twice to the hash and should fail in order to pass the test.

tests/quests/test_quests_state.c
```diff
@@ -585,3 +585,131 @@ Test(quest,complete_quest)
...
+    char *quest2_id = "quest two";
+
+    quest_t *quest1 = quest_new(quest1_id, NULL, rewards, stat_req);
+
+    quest_hash_t *test_hash_table = NULL;
+
+    int add_quest1 = add_quest_to_hash(quest1, &test_hash_table);
+    int add_quest2 = add_quest_to_hash(quest1, &test_hash_table); 
+
+    cr_assert_eq(add_quest2, FAILURE, "Add quest did not work");
```
charlieweiler: The error message should maybe be that "add_quest_test2 should not succeed in adding the same quest twice." instead of "Add quest did not work" because in this case, add_quest DID succeed; it just shouldn't have.

tests/quests/test_quests_state.c
```diff
@@ -585,3 +585,131 @@ Test(quest,complete_quest)
...
+
+    quest_t *quest1 = quest_new(quest1_id, NULL, rewards, stat_req);
+
+    quest_hash_t *test_hash_table = NULL;
+
+    int add_quest1 = add_quest_to_hash(quest1, &test_hash_table);
+    int add_quest2 = add_quest_to_hash(quest1, &test_hash_table); 
+
+    cr_assert_eq(add_quest2, FAILURE, "Add quest did not work"); 
+}
```
charlieweiler, line 715: Finally, I get that a lot of the more random commits were meant to just test tiny things and fix the environment issues w/ Github, but it would have been easier to pour through the commit logs with more descriptive names than "poggers" and "messing around", lol. Especially as some of these odd environmental issues can certainly occur again in the future.

Other than those mostly form-concerned comments, I think it looks good!

tests/quests/test_quests_state.c
```diff
@@ -585,3 +585,131 @@ Test(quest,complete_quest)
...
+    quest->status = 0;
+
+    reward_t *res = complete_quest(quest);
+
+    cr_assert_eq(res, NULL,"quest_completed failed to reward the item");
+}
+
+
+
+/*see if get_quest_from_hash works when there is quest in Hash */
```
philipchenaux, line 611: Good to see that you are including proper function header comments.

philipchenaux (approved):
Good coverage of tests, and I am happy with how this looks. Great work.

charlieweiler (approved):
Comments are resolved - good to go!

Issue #1287, opened 2022-05-05
Create Multiclass Module within Playerclass Feature


wellingtj:
Currently, the class.c file contains a lot of code on multi-classing. This includes functions like multiclass_shortdesc, multiclass_longdesc, multiclass, has_component_class. This code could be contained within its own module, multiclass.c. From there, it would be easier and ideal to implement testing for multiclassing

wellingtj:
it appears the code won't compile. I will debug this tomorrow

wellingtj:
I ahce

wellingtj:
I have fixed the issue and submitted a pull request. I believe I accidentally had some code from another issue I worked on which I removed

lilyehsani:
Issue Score: Excellent

Comments:
Great work! One comment:
1. For a task of this size, we would've expected to see more status updates and/or status updates that were more detailed. For example, it would have been nice to get updates about the debugging and what the bug was/how it was fixed.

PR #1333, opened 2022-05-09
Move the multiclassing features into its own module

wellingtj:
Currently, the class.c file contains a lot of code on multi-classing. This includes functions like multiclass_shortdesc, multiclass_longdesc, multiclass, has_component_class. It would therefore make sense to move the multiclassing features into their own module so that we can expand them. This pull request has taken care of that task as well as all of the necessary changes to the cmake files and header files


wellingtj:
I have refactored multiclassing code into its own module

wellingtj:
This is for issue #1287

Vincent-Zhang4135 (approved):
Jake, you did a great job refactoring player class into a class and multi class functional modules. abstracting away these into these two files will come in handy to future developers that need to use these modules, as they will have an easier time digging into the code base to see what functionalities we support. 

I appreciate your carefulness in also making sure to update the CMakeList.txt to include this new header file when we link/compile our code. From what I see, your code passes the build, and so I would like to approve this and move this onto Lily for a final approval.

lilyehsani (approved):
Looks good to me! Approved.

wellingtj:
All the multiclassing features previously in class.c are in their own module multiclassing.c. The functionality remains intact, the linking works, and the modlue is now completely refactored

lilyehsani:
Issue Score: Excellent

Comments:
Great work on this PR!

PR #1358, opened 2022-05-11
Openworld/room autogenerate

williamhu314:
We have created 2 new functions in this pull request:

1. roomspec_autogenerate, which uses an algorithm and the provided adjacency matrix to output a new roomspec given a current roomspec.
2. room_autogenerate, which given a starting room and a direction, generates a new room in the specified direction from the current room, with the associated roomspec being created by room_autogenerate.

Additionally, included in this pull request is the new specgraph struct, which provides an adjacency matrix showing relationships between roomspecs rather than just a list of roomspecs, which was the functionality of the former speclist struct.

Due to the modification of speclist into specgraph, lots of functions associated with specgraph and lots of tests associated with specgraph or functions using specgraphs do not compile and will need to be completely redesigned and reimplemented. Those functions and tests are temporarily commented out in this pull request. We will uncomment them gradually as we update them to accommodate specgraph. This task will be time consuming so we don’t want to wait for that to be completed before working on other functions.

The demo and demo makefile were also causing issues since they are using the speclist struct. The makefile for the demo was deleted (its contents were moved into the demo in case we need it for reference later).


rohandeme:
Looks good to me!

rohandeme (commented):
All looks good! (part 2)

rohandeme (approved):

src/openworld/src/gen_structs.c
```diff
@@ -192,58 +192,57 @@ roomspec_t* roomspec_new(char *room_name, char *short_desc, char *long_desc, ite
...
-    list->next = NULL;
+    specgraph->num_roomspecs = num_roomspecs;
+    specgraph->roomspecs = roomspecs;
+    specgraph->edges = edges;
     return SUCCESS;
 }
 
 /* see gen_structs.h */
-speclist_t* speclist_new(roomspec_t *spec)
+/*specgraph_t* specgraph_new(roomspec_t *spec)
```
philipchenaux, line 207: why have you commented out specgraph_new and specgraph_free functions?

include/openworld/gen_structs.h
```diff
@@ -67,31 +67,35 @@ typedef struct roomspec {
...
-    struct speclist *prev;
-    struct speclist *next;
-} speclist_t;
+/* specgraph_t struct
+ * This struct functions as an adjacency matrix for the relationships between roomspec_t's
+ * The struct contains:
+ * - int num_roomspecs: The number of different roomspecs available in the game
+ * - roomspec_t roomspecs: list of pointers to roomspecs corresponding to each node on specgraph
+ * - int **edges: edges of graph representing the relationship between each roomspec
+ *                the higher the number, the more likely rooms of this roomspec appear
```
philipchenaux, line 76: It might be worth explaining this further. It is not initially clear why a higher number indicates this.

src/openworld/src/autogenerate.c
```diff
@@ -139,36 +183,34 @@ int multi_room_generate(game_t *game, gencontext_t *context, char *room_id, int
...
 
-    DL_COUNT(spec, tmp, count);
-    int idx = rand() % count, i = 0;
-
-    DL_FOREACH(spec, tmp) {
-        if (i == idx) {
-            return tmp->spec;
+    for(int i=0; i<num_rooms; i++){
+        if(i == idx){
+            return roomspecs[i];
```
philipchenaux, line 210: Using a list like this is fine, but in some ways you are re-inventing the wheel when you can just use existing list implementations.

philipchenaux (changes requested):
I have added a few questions/comments about this PR, although you noted that a lot of the tests break. Ideally you should have a few tests for what you have implemented so far. Especially for the basic functions.

philipchenaux:
I just saw that your new PR addresses some of the issues above. I will review your most recent PR instead.

philipchenaux:
This PR will be closed since all edits are part of https://github.com/uchicago-cs/chiventure/pull/1384

PR #1337, opened 2022-05-09
Quests/player module update

Castel-Jaloux:
Updated the player module (player.h and player.c) within the game-state directory in order to properly integrate quests with the player module. Due to circular dependency issues, we cannot directly import quests into the player module. To resolve this issue, we created player_quest and player_task structs to be included within the player struct, as well as corresponding functions (new, init, free) to ensure that they are properly functional.

PR #1362, opened 2022-05-11
Quests/player update 2

JSerf02:
Updates quest-structs.h, quest-state.h, quest-state.c, and the quest tests to work with the player module and with the new player_quest and player_task structs. Contributes to the same redesign as Pull Request #1337.


JSerf02:
Some of these changes are the same as Pull Request #1337 but I'm not sure why it's showing up as new changes. If there are any differences, the changes from the previous PR should be accepted.

jakeziqiwei (approved):
Everything looks. Some concerns were already addressed in slack!

include/quests/quests_structs.h
```diff
@@ -48,14 +47,14 @@ typedef struct active_mission {
...
- * This union represents a mission. Can be used to create a task.
+ * This struct represents a mission. Can be used to create a task.
  * 
  * Components:
  *  a_mission: an active mission
  *  p_mission: a passive mission
  *
  */
-typedef union mission {
+typedef struct mission {
```
philipchenaux, line 57: Does this imply that a mission must be both an active and passive mission?

src/quests/src/quests_state.c
```diff
@@ -372,21 +335,43 @@ int add_task_to_quest(quest_t *quest, task_t *task_to_add, char *parent_id)
...
 /* Refer to quests_state.h */
-int start_quest(quest_t *quest)
+int can_start_quest(quest_t *quest, player_t *player)
+{
+    stats_hash_t *stats_hash = player->player_stats;
+    double health = get_stat_current(stats_hash, "health");
+
+    if (health >= quest->stat_req->hp && 
+        player->level >= quest->stat_req->level){
+            return 1;
```
philipchenaux: You should stick to the convention of returning either SUCCESS or FAILURE instead of 1 or 0.

include/quests/quests_structs.h
```diff
@@ -134,7 +132,6 @@ typedef struct quest  {
     task_tree_t *task_tree;
     reward_t *reward;
     stat_req_t *stat_req;
-    int status;  
     UT_hash_handle hh;
 } quest_t;
```
philipchenaux, line 131: Make sure that you update the appropriate documentation if you change the structs (i.e remove the explanation of status if you remove it from the struct).

include/quests/quests_structs.h
```diff
@@ -85,7 +84,6 @@ typedef struct task {
     mission_t *mission;
     char *id;
     reward_t *reward;
```
philipchenaux, line 85: Make sure that you remove the `completed` explanation in the comment explaining the task struct.

include/quests/quests_structs.h
philipchenaux: One additional note, make sure that you update the new player_quest_t and player_struct_t structs that you have created in the player.h file to include the documentation about the `completed` field as it is currently not clear as to why quests have a `completion` of type int, while tasks have `completed` of type bool.

philipchenaux (changes requested):
Added a few minor comments to change, but once theses are addressed you should be good to go!

include/quests/quests_structs.h
JSerf02: We had to change it from a union to a struct to get it to pass the tests, since it was originally passing only by unknowingly resetting the mission pointer to NULL so the mission wouldn't actually get checked. This is really just a temporary change though since we already created an entire redesign of the mission system that addresses this issue better and we plan to implement that in the next couple of days.

src/quests/src/quests_state.c
JSerf02: Just switched it to bools. Chose bools instead of SUCCESS/FAILURE because SUCCESS and FAILURE typically correspond to whether or not a function actually did what it was supposed to while here, the function is not returning whether or not it succeeded.

include/quests/quests_structs.h
philipchenaux: Okay this makes sense.

philipchenaux (approved):
I am happy with how this looks. Well done guys.

philipchenaux:
Issue Score: Excellent

Comments:
Great job on this PR!

PR #1363, opened 2022-05-11
Quest/delete quest

### LINKED ISSUE #1294: update tests to reflect new changes in quest_Id for delete quest
Last sprint, the team members changed some elements of the quest struct, we need to update the quests to reflect this change

elizabethli31:
We used uthash macros to write two functions to remove quests from a hash table. One of the functions allows the user to remove a single quest and the other removes all quests from the hash table. This pull request incorporates the working add_quest_to_hash and get_quest_from_hash functions as well as the change of quest_ids from long int to char.


src/quests/src/quests_state.c
```diff
@@ -520,3 +520,35 @@ reward_t *complete_quest(quest_t *quest)
...
+    quest_free(check); 
+    if (get_quest_from_hash(quest_id, hash_table) != NULL){
+        return FAILURE;
+    }
+    return SUCCESS;
+
+}
+
+/* refer to quests_state.h */
+int remove_quest_all(quest_hash_t *hash_table)
```
charlieweiler, line 591: Is there any particular reason you guys don't have an error check line in remove_quest_all even though it's in remove_quest_hash? If it's necessary in one, I think it's probably necessary for both

charlieweiler (changes requested):
Looks good, I'd just add an error check in remove_quest_all before merging (unless there's a reason for its absence)

src/quests/src/quests_state.c
charlieweiler: Elizabeth informed me that an error check isn't necessary because it is not present in other team's modules, and also because of how this function links with hash_iter.

charlieweiler (approved):
Functions look good and there are reasons for excluding the error check, so it's good by me.

philipchenaux (changes requested):
You have included all CMake files, please recommit only the files that are relevant to this PR.

elizabethli31:
I removed the cmake files and merged the new code from dev

philipchenaux (approved):
Looks good, great work!

philipchenaux:
Issue Score: Excellent

Comments:
Great job on this PR!

PR #1367, opened 2022-05-12
Updates in sandbox

jiaoboxuan:
I did a slight update in the `sandbox` folder in `src/sound`. I included a few extra functions and also pushed the music files into this branch for the sake of ease of the teams wanting to test it out next year (last year's team had the audio directory in their local files, and they didn't push it up here, so it was kind of a hassle).


jiaoboxuan:
Closing Statement: The sandbox could be readily used for future teams. In the future, it is also possible that they could put more music files in it in the future.

PR #1378, opened 2022-05-13
Cli/ associate words in quotes into a single item (token)

mrestuccia1:
Currently, there is a limit of four tokens per input, which limits the possible inputs. While there is no need for it right now, we want to be able to say, for example, go to "ring of fire", and have that be counted as three tokens instead of five. This will allow for more tokens, but also for games to have special names beyond north, south etc... We edited the current parsing framework such that if the command line input contains something in quotes, whatever is inside is considered its own token, no matter how many words are inside.


src/cli/src/parser.c
```diff
@@ -105,27 +106,80 @@ char **parse(char *input)
...
     //Initializes all words to NULL
     for(int i = 0; i < TOKEN_LIST_SIZE; i++)
     {
         words[i] = NULL;
+        temp[i] = NULL;
+        temp2[i] = NULL;
     }
 
-    char *token = strtok(input, " ");
+    char *token2 = strtok(input, "\"");
```
majorsylvie: tokenizing on quotation mark

src/cli/src/parser.c
```diff
@@ -105,27 +106,80 @@ char **parse(char *input)
...
+    int pos = 0;
 
-    for(int i = 0; i < TOKEN_LIST_SIZE; i++)
+    for(int i = 0; i < 4; i++)
```
majorsylvie: why change this from being the constant to 4? Should this not be linked?

src/cli/src/parser.c
```diff
@@ -105,27 +106,80 @@ char **parse(char *input)
...
+
+        temp[i] = token2;
+        token2 = strtok(NULL, "\"");
+
     }
 
-    //If there are more than 4 words, parser returns NULL and does not attempt
-    //to pass the first four words as tokens
-    if(token != NULL)
+    if (input[0] != '\"')
```
majorsylvie, line 132: should add some kind of description here about what the loop is doing.

src/cli/src/parser.c
```diff
@@ -105,27 +106,80 @@ char **parse(char *input)
...
     {
+        char *token = strtok(input, " ");
+
+        for(int i = 0; i < 4; i++)
```
majorsylvie: Again, why 4? It is not clear to me. 
If this is meant to be the word limit constant, then keep it as such

majorsylvie (changes requested):
Please add some description within your code (particularly at the commented portions) so that it is more clear what you code is meant to be doing.

I already know it is supposed to parse for text within a quote, but it is not obvious when you are looking.

mrestuccia1:
As suggested by Faruk in his review, we added comments and changed the number 4 back to the pre-defined variable TOKEN_LIST_SIZE. The code still passes all the tests and added the functionality described above. We will check with Faruk one more time, but the next step will be to get the code looked over by the TA

src/cli/src/parser.c
```diff
@@ -105,27 +106,93 @@ char **parse(char *input)
 
     char **words;
     words = (char**)malloc(sizeof(char*)*TOKEN_LIST_SIZE);
+    char **temp;
+    temp = (char**)malloc(sizeof(char*)*TOKEN_LIST_SIZE);
+    char **temp2;
```
jacknugent1529: More descriptive names (rather than `temp` and `temp2`) would make this code much easier to read. Also, if names are not enough, give a brief comment explaining why there needs to be 3 of these arrays.

src/cli/src/parser.c
```diff
@@ -105,27 +106,93 @@ char **parse(char *input)
...
+            token3 = strtok(NULL, " ");
+
+        }
+
+        //If there are more than 4 words, parser returns NULL and does not attempt
+        //to pass the first four words as tokens
+
+        if(token3 != NULL)
+        {
+        return NULL;
```
jacknugent1529: style: indentation, and include a space after `if`

src/cli/src/parser.c
jacknugent1529: @fkbad Is correct. Avoid using "[magic numbers](https://uchicago-cs.github.io/dev-guide/style_guide_c.html#avoid-magic-numbers)" because it is often unclear what the numbers mean (especially if someone changes them in the future). Instead use descriptive constants

jacknugent1529 (commented):
The main thing to change is the descriptiveness of the names. Everything else looks good, although I'll do a more thorough review once this is done. Also, since this is an integral part of chiventure, testing is especially important. So, how have you tested this code and can we be confident that it interpreting the quotes in this manner won't break something elsewhere?

milesgardner:
General functionality has been completed.  Unit tests have been implemented, and we believe that tokens surrounded by quotations is ready to be merged into the main branch.

src/cli/src/parser.c
mrestuccia1: This is correct, we switched it initially to test the function inside a sandbox, and never switched it back. This change has been made, and all the for loops now have dynamic variables.

src/cli/src/parser.c
mrestuccia1: We have added descriptive comments for everything we added, so this change should be fixed now.

src/cli/src/parser.c
mrestuccia1: This change has been made, there are descriptions for all our new added code, and we changed the "magic numbers" back to the dynamic variables they had been set to in the previous version.

src/cli/src/parser.c
mrestuccia1: The indentation has been added.

src/cli/src/parser.c
mrestuccia1: We reverted the code to only need two arrays, one named words, which is the array of tokens tokenized by spaces, and one named words_quotes, which is the array of tokens tokenized by quotes.

mrestuccia1:
> Please add some description within your code (particularly at the commented portions) so that it is more clear what you code is meant to be doing.
> 
> I already know it is supposed to parse for text within a quote, but it is not obvious when you are looking.

majorsylvie (approved):

src/cli/src/parser.c
```diff
@@ -105,27 +106,98 @@ char **parse(char *input)
...
 
     //Initializes all words to NULL
     for(int i = 0; i < TOKEN_LIST_SIZE; i++)
     {
         words[i] = NULL;
+        by_quotes[i] = NULL;
     }
 
-    char *token = strtok(input, " ");
+    char *token2 = strtok(input, "\"");
```
jacknugent1529: I still think this naming scheme is a bit confusing. What does "token2" mean in this context?

src/cli/src/parser.c
```diff
@@ -105,27 +106,98 @@ char **parse(char *input)
...
-    //If there are more than 4 words, parser returns NULL and does not attempt
-    //to pass the first four words as tokens
-    if(token != NULL)
+    //If the first character of the input is not "
+    if (input[0] != '\"')
     {
+        char *token = strtok(input, " ");
+
+        //Populates words array with tokens separated by space (" ") character
+        for(int i = 0; i < 4; i++)
```
jacknugent1529: style: space between `for` and `(` (`for (...`). Also, replace 4 with a constant

src/cli/src/parser.c
jacknugent1529: I mean that the `return NULL` line should be indented once more based on its location in the curly braces. And the space between `if` and `(` doesn't seem to have been added.

milesgardner:
Added unit tests as requested through slack.

src/cli/src/parser.c
```diff
@@ -105,27 +106,98 @@ char **parse(char *input)
...
+
+        if(token != NULL)
+        {
         return NULL;
+        }
+
+        return words;
+
+    //If the first character of the input is "
+    }else{
```
jacknugent1529: style: spaces around `else`

jacknugent1529 (commented):
It looks mostly good and the tests are a nice addition. There's just a few style issues remaining

src/cli/src/parser.c
mrestuccia1: Just changed token2 to token_quotes, since it is the token separated by quotes.

src/cli/src/parser.c
mrestuccia1: Added the spaces around else, as suggested.

src/cli/src/parser.c
mrestuccia1: Added the spaced between for and ( in all instances, and changed 4 to the dynamic variable it was previously set to.

src/cli/src/parser.c
mrestuccia1: The indentation should now be added for all instances of return NULL, and spaces between all instances of if and "(" have been added.

mrestuccia1:
All requested edits have been made, we believe the pull request should now be good to go.

jacknugent1529 (commented):
Nice job making the changes! It looks good to me

jacknugent1529 (approved):

jacknugent1529:
PR Score: Excellent

Comments: Excellent work on this PR! The description is good, and you addressed each of the suggestions well. This is not a rubric item, but the commit "discipline" could be improved to give insightful yet concise commit messages and commits in places that logically make sense.

PR #1382, opened 2022-05-14
Implement combined and sequential skills

### LINKED ISSUE #1278: Implement combined and sequential skills based on design documentation
Similarly to the design warmup, in issue #1230 we added a header file for complex skills and added supporting structs as well as functions to be implemented later. In this issue, we will actually implement the functions we prepared by 1) filling out the functions in a .c file and 2) editing some of the functions in the skill module to support the changes to the skill struct we made. This will finish the implementation of combined and sequential skills.

anchristensen:
@nbilaver and I have finished issue #1278 with this pull request; we filled in the body of the functions we wrote as part of the design process, including a new, init, free, xp_up, level_up, and execute function for complex skills. Tests have also been added in a new complex_skill_test file for each of these functions.


KabirDubey (approved):
I approve of this for merging.  The design of combined and sequential skills seems to be in accordance with the wiki explanation and the tests are sufficiently exhaustive (notwithstanding the complex_skill_test.c comment block reserved for next sprint). Other comments:
-The extra "NULL" slot in the skill_new() function in the class_prefabs.c file may be confusing the playerclass team (see slack) because they do not know about the existence of complex skills. 
-The rationale for the tests could be more clear. That sort of explanation may be better implemented in example files, though. Good enough for merging and reserving for next sprint.

src/skilltrees/src/complex_skills.c
```diff
@@ -0,0 +1,136 @@
...
+#include <string.h>
+#include "skilltrees/skill.h"
+#include "skilltrees/complex_skills.h"
+
+/*See complex_skills.h */
+complex_skill_t* complex_skill_new(complex_skill_type_t type, skill_t** skills, int num_skills){
+    complex_skill_t* complex;
+
+    if (num_skills == 0) {
+        fprintf(stderr, "skill_new: max_level is invalid\n");
```
jacknugent1529: Why is the error message "max_level"? Is this intentional?

src/skilltrees/src/complex_skills.c
```diff
@@ -0,0 +1,136 @@
...
+
+/*See complex_skills.h */
+int sequential_complex_skill_execute(complex_skill_t* complex_skill, chiventure_ctx_t* ctx){
+    if(complex_skill->type != SEQUENTIAL){
+        return FAILURE;
+    }
+
+    // Stop execution once a sub_skill fails
+    for(int i = 0; i < complex_skill->num_skills; i++){
+        if (skill_execute(complex_skill->skills[i], ctx) == FAILURE)
```
jacknugent1529: Although technically not necessary, it is still best to include curly braces here in order to prevent confusion or bugs in the future

src/skilltrees/src/complex_skills.c
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/*See complex_skills.h */
+int complex_skill_level_up(complex_skill_t* complex_skill){
+    assert(complex_skill != NULL);
+    
+    int x;
+    for (int i = 0; i < complex_skill->num_skills; i++){
+        x = skill_level_up(complex_skill->skills[i]);
+        if(x == FAILURE){
```
jacknugent1529: style: space after `if`

src/skilltrees/src/complex_skills.c
```diff
@@ -0,0 +1,136 @@
...
+            return FAILURE;
+        }
+        // printf("Skill level up for sub-skill %s returned %d", complex_skill->skills[i]->name, x);
```
jacknugent1529: This should be removed before merging into dev

src/skilltrees/src/complex_skills.c
```diff
@@ -0,0 +1,136 @@
...
+int complex_skill_xp_up(complex_skill_t* complex_skill, unsigned int xp_gained){
+    assert(complex_skill != NULL);
+    int x;
+
+    for (int i = 0; i < complex_skill->num_skills; i++){
+        x = skill_xp_up(complex_skill->skills[i], xp_gained);
+        if(x == FAILURE){
+            return FAILURE;
+        }
+        // printf("Skill xp up for sub-skill %s returned %d", complex_skill->skills[i]->name, x);
```
jacknugent1529: This should be removed before merging into dev

include/skilltrees/skilltrees_common.h
```diff
@@ -71,11 +74,27 @@ typedef struct skill {
     // The minimum number of experience points needed to level up
     unsigned int min_xp;
 
-    // The pointer to the linked list that contains all the effects that a skill can have
+    // The pointer to the linked list that contains all the effects that a skill 
+    // can have
     effect_t* skill_effect;
 
+    //The pointer to information on complex skills
+    complex_skill_t* complex;
```
jacknugent1529, line 82: Can you clarify why this field is here? See review comment.

jacknugent1529 (commented):
This looks really good! The documentation is great and besides a few points the style is very consistent. I'm a bit confused about why there is a `complex_skill_t *complex` field in the `skill_t` type. How would you intend this to be used? My understanding is that the complex skill type would contain a list of `skill_t`'s. Also, this can be something to consider for the future, but what if I wanted to make a complex skill with several other complex skills internally (e.g. a combined skill consisting of multiple sequential skills)?

src/skilltrees/src/complex_skills.c
jacknugent1529: This comment refers to the `break` in the line below

src/skilltrees/src/complex_skills.c
anchristensen: No it wasn't intentional, I think that must be a copy paste error and I will update this to accurately reflect what it should say.

src/skilltrees/src/complex_skills.c
anchristensen: Fixed

src/skilltrees/src/complex_skills.c
anchristensen: Fixed

src/skilltrees/src/complex_skills.c
anchristensen: Fixed

src/skilltrees/src/complex_skills.c
anchristensen: Fixed

include/skilltrees/skilltrees_common.h
nbilaver: This design choice was made with the idea that it's better to extend upon the existing skill struct rather than have a "simple" skill struct and a "complex" skill struct. This new field was made to store all of the extra information needed if a skill is complex, so perhaps naming this struct "complex_skill" might have been a misnomer, I'll redo the comment above it to better describe what it is. 

I assume your question is asking why we have the complex_skill struct be a part of the skill struct rather then letting it be its own independent struct. This design allows for other structs that include skill_t (mainly the skilltrees struct) to use complex skills without adding a new complex_skill field to each of them. It also allows for other parts of the code to more easily use complex skills by using existing functions (skill_execute, skill_level_up) to run them. Does that explanation make sense?

nbilaver: Also, to respond to your question of complex skills including complex skills within them, by adding the complex_skill field to the skill_t struct, it means that the list of skills within complex_skill_t could include a skill_t struct that uses this field to include a sub-complex skill.

jacknugent1529:
@nbilaver @anchristensen It looks like there were some issues when trying to merge it with dev (see the github actions run). Can you just resolve these quickly and then it should be good to go?

nbilaver:
@jacknugent1529 @anchristensen Just fixed it, checks are now passing.

jacknugent1529 (approved):
Looks good to me!

jacknugent1529 (approved):

jacknugent1529:
PR Score: Excellent

Comments: Excellent work! The PR was fantastic, well documented, and well-written code. Before making the PR, it is generally a good idea to review your own code to try to catch, for example, commented out code and other items that shouldn't be in the dev branch.

PR #1393, opened 2022-05-15
Open world/add coords to room struct

### LINKED ISSUE #1326: implement coordinates in the room_t struct
The autogenerate module will run into the situation below:
IMG_1553
where it has generated two rooms which are beside each other from a global view but they are not automatically connected because they were not generated from one to the other. 

In this example, autogenerate algorithm created room 1, then room 2 towards east, then it stopped generating rooms east and started its chain south where it created rooms 3 and 4. 

Being able to know whether rooms not generated right after each other are right next to each other would be easy if the game_t struct had a list of coordinates of generated rooms. 

At this point, the autogenerate algorithm would have to refer to specgraph_t struct to check the relationship between the two rooms (in this case, rooms 2 and 4). Finally, it could autogenerate whether or not a path is generated between the two rooms based on their weighting in the graph (e.g. if their relationship is 0, no path would be generated but if it was 4/5, the likelihood would be high).

helenastc:
Added coordinates struct to room struct to be able to keep track of rooms and relationships. Also created coords_new, coords_init, coords_free, add_coords_to_room, and find_coords_of_room functions.


rohandeme (approved):
Looks good to me! I saw that you included tests in another PR, so I will take a look at those.

src/game-state/src/room.c
```diff
@@ -231,6 +231,52 @@ npc_t *get_npc_in_room(room_t *room, char *npc_id)
...
+    if (check != SUCCESS)
+        return NULL;
+
+    return coords;
+}
+
+/* See room.h */
+int coords_free(coords_t *coords){
+    free(coords);
+        return SUCCESS;
```
philipchenaux: Minor stylistic comment, but try to keep your indentation consistent.

philipchenaux (commented):
Very minor comment, but overall well done. Note that in the future when you submit a PR make sure that you include tests relating to that PR in the same PR as this ensures what you are implementing is actually working.

src/game-state/src/room.c
philipchenaux: Once you address this change, I'll be happy to merge this and move onto merging the other PRs related to this!

helenastc: Fixed the indentation

philipchenaux (approved):
Looks good to merge, well done!

philipchenaux:
Issue Score: Excellent

Comments:
Great job on this PR!

PR #1394, opened 2022-05-15
1375 editing item_t struct

### LINKED ISSUE #1375: GUI: Editing item_t struct
Currently, the item struct looks like this:
`typedef struct item {
    UT_hash_handle hh; //makes this struct hashable for the room struct (objects in rooms) and player struct (inventory)
    char *item_id;
    char *short_desc;
    char *long_desc;
    game_action_hash_t *actions;
    attribute_hash_t *attributes; // a hashtable for all attributes
} item_t;`

But, there is no place for the image of the item that the game developer can input. Because we plan to be able to implement the inventory display by looking at the item struct and finding the respective file path to an image, we want to include another element in the struct, something like `char* item_image;` that would be a file path to a png/jpg.

KendrickXie:
Three new elements have been added to the ```item_t``` struct.

```char* item_image``` is to store a file path to an image of an item.
```unsigned int* inventory_x_pos``` and ```unsigned int* inventory_y_pos``` are to store x and y coordinates of an item within an inventory.

### CODE REVIEW

**Review by HarrisonD123 — APPROVED** [review]
These changes are pretty straightforward and don't look like they'd create any conflicts in the code. Any assignment of these members to values should most likely be left up to the graphics module.

Looks good to merge!

File: include/game-state/item.h [hunk 1]
```diff
@@ -66,6 +66,9 @@ typedef struct item {
     attribute_hash_t *attributes; // a hashtable for all attributes
     effects_hash_t *stat_effects; // hashtable of effects item can have (set to NULL if no effects)
     struct item *next; // points to item w/ identical id, if it exists
+    char* item_image; // a file path to an image of an item
```
ecoble: Please put the asterisk on the other side to be consistent with style, i.e. `char *item_image`, for this line and the next two lines.

ecoble (changes requested):
Looks good! I would also recommend getting a review from a member of the GUI team to confirm that these changes work for their interface.

ecoble (approved):

imadejski (approved):

PR #1399, opened 2022-05-16
Openworld/add tag to room

### LINKED ISSUE #1392: add capability of searching for roomspec_t from room_t
Discussion with Helena made us realize that in order to verify if we can connect two rooms during autogeneration, we need to be able to relate a room_t struct with its roomspec_t struct because the relationships between rooms are stored in specgraph_t which is an adjacency matrix of roomspecs and not rooms.

helenastc:
Added int tag element to the room struct so that we can link rooms to specgraph. i.e. the specgraph will contain room_specs each with a tag to identify them. When we generate a room from this, the room will also have the tag so that we can go back to specgraph to find it.


include/game-state/room.h
```diff
@@ -353,3 +352,43 @@ int add_coords_to_room(coords_t *coords, room_t *room);
 coords_t *find_coords_of_room(room_t *room);
 
 #endif
```
lukasg8: I think this #endif needs to be removed.

lukasg8 (approved):
Functions look good and descriptions are clear and concise.

philipchenaux (approved):
Nice and simple PR, well done!

philipchenaux:
Issue Score: Excellent

Comments:
Great job on this PR!

PR #1401, opened 2022-05-16
Prefab a default NPC Class

### LINKED ISSUE #1288: Prefab an NPC class to represent a "base level" class
It doesn't make conceptual sense to have every npc have some elite sounding fighter class like "warrior" or "wizard". Therefore, it would be ideal to prefab a more vanilla class for NPC's that aren't intended to be something special. This would be a kind of "base" class that modeled after the other prefabs in class_prefabs.c and is prefabricated for NPC's in this issue here.

wellingtj:
This PR has used the preexisting models in class_prefabs.c and repurposed them into making a 7th prefab that functions as a base class and is intended for integration with NPC's. This class is called "basic". At the moment it is just a class that is intended to be initialized with no skill trees, items, or anything that is optional. It does have some pretty basic stats, but they are the lowest of any possible


Vincent-Zhang4135 (commented):
It looks good so far! However, I do want to mention that I noticed that you arbitrarily set the basic stats to be 1 for everything. This is an arbitrary decision, and so it is not a big deal, but what do you think if we change it to 5 instead? I think that it would make more sense in the context of all the other classes, because all the other classes have starting stats of factors of 5. 

Additionally, there is a function called check_and_add_stat in class_prefab.c, and I noticed in this function it makes all stats to be 100, which is what it originally was. I think that implies that it is also initializing the originally game stat to be based off of the basic class. Do you think it would make sense to also make sure to change all these to the basic stats you changed for basic class?

wellingtj:
I have fixed this according to your suggestions vincent. I have set the basic class to have multiples of 5 for their stat, but I believe check add stat adds the stat itself, and creates a maximum for it

src/playerclass/src/class_prefabs.c
```diff
@@ -176,19 +175,19 @@ class_t* class_prefab_new(game_t* game, char* class_name) {
...
+     * 5 Speed
+     * 5 Physical Defense
+     * 5 Physical Attack
+     * 5 Ranged Attack
+     * 5 Magic Defense
+     * 5 Magic Attack
+     * 5 Max Mana 
      * These stats can be off the scale stupid because why not? This is a placeholder class */ 
-    if (!strncmp(temp_name, "basic", MAX_NAME_LEN)) {
+    else if (!strncmp(temp_name, "basic", MAX_NAME_LEN)) {
```
lilyehsani, line 192: Now that this prefab exists, would it also be possible to set this to be the default class for an NPC in npc_new()/npc_init()? Something along the lines of, if NULL is pass to those functions for the NPC class, set it to be this.

lilyehsani (changes requested):
Looks good so far! One comment.

wellingtj:
Hi @lilyehsani, I added some code to to npc new and npc init but for some reason no matter what I did there was no way to get npc.h to recognize the chiventure_ctx_t struct. I asked multiple other people (including a 4th year cs major) and nobody could figure out why that was the case

wellingtj:
I have gotten everything back to its original state from commit [98fef1d] besides that one else statement Lily showed me

src/playerclass/src/class_prefabs.c
wellingtj: I have fixed this

wellingtj:
this PR ultimately creates a new prefab for NPC's to use as a default class. It also fixes a few minor bugs in the codebase

lilyehsani (approved):
LGTM! Approved.

Vincent-Zhang4135 (approved):
Hi Jake, good job with the NPC prefab basic class. I noticed that you also created a separate test for the basic class prefab and i appreciate that as it will make debugging easier for future developers if there is an issue. I am approving this and looking forward to having this get merged!

wellingtj:
A basic class has been completely implemented into the codebase with this PR through class_prefabs.c

lilyehsani:
Issue Score: Excellent

Comments:
Great work on this PR!

PR #1403, opened 2022-05-16
action-management/fixed issue #1072 - updated custom-actions tests to use attribute-specific creator functions

### LINKED ISSUE #1072: Fix attribute_new in custom-action-cond.c
Currently the function: `attribute_t *attribute_new(enum attribute_tag tag, attribute_value_t value)` does not function properly.
It does not have a return value and it does not strdup the attribute key as it is just being set to null. 
A return value needs to be added, so it functions at all. I am not sure how it actually doesn't crash now.
Strduping the attribute key will allow it to be freed using attribute_free, since that attempts to free the attribute_key.

This function should also probably be moved to item.c, since that is where all of the other functions for creating attributes are. 
Since there are an array of other new_attribute functions already in item.c,  attribute_new could also be removed and all the tests updated to use the already existing functions.

Don-Osipov:
Removed the bugged general attribute creator function "attribute_new" in both custom_actions-cond.c and custom-actions-common.h and updated all tests that previously used the function to instead use the attribute-type-specific attribute creator functions found in item.c. These tests were found in test_ca_effect.c and test_ca_cond.c


KendrickXie (approved):
Looks good. Followed the plan laid out in issue #1072.

ecoble (approved):
Looks good! As a note, it might be good to mention in your documentation somewhere that at one point custom actions was trying to make a consolidated attribute_new function, but you removed it in favor of using existing functions. It could be a future task for your team or next year's teams to try and revive this consolidation for the sake of streamlining.

ecoble:
Issue Score: Excellent

Comments:
Nice job diving into the custom actions code to debug this! Please include closing statements on PRs.

PR #1404, opened 2022-05-16
Cli/battle integration

### LINKED ISSUE #1298: cli: implement function to change game mode to battle on user input (battle team request)
Team members: Brianna, Kate, Faruk

Our goal is to meet with the RPG battle team to discuss their requests and implement them as we are able. They want to implement a run_battle function that will run a battle with an NPC from start to finish. Our side is to implement a function that will take the inputted "fight npc_name" command, parsed and put into an array of strings, and subsequently change the game state to BATTLE mode.

majorsylvie:
We've added the framework for the `run_battle` function within operations.c and .h.
It is generally following the framework of the operation functions, such that it checks what the tokenized string input has and follows different control flows based on that. (such as seeing if there is a second argument, and returning an error message when that happens)

We still need to update the framework such that chiventure recognizes the command, as right now just the addition of this framework alone does not ensure the command is recognized


include/cli/operations.h
```diff
@@ -279,4 +279,15 @@ char *palette_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx);
  */
 char *talk_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx);
 
+/* Initiates battle between NPC and Player
+ *
+ * Parameters:
+ *  - tokens: parsed input string (validified)
+ *  - ctx: pointer to a chiventuree context struct
```
mrestuccia1: Just fix Chiventure name to only have one e

src/cli/src/operations.c
```diff
@@ -574,12 +574,14 @@ char *battle_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
 
-    if (npx->will_fight == 0) {
-        return "%s does not want to fight.";
+    if (npc->will_fight == 0) {
+        return "%s does not want to fight.", tokens[1];
     }
 
-    set_game_mode(ctx->game, BATTLE, npc->npc_id);
+    set_game_mode(ctx->game, BATTLE, npc->npc_id); 
+    //Apparently this should automatically call function that initialize and run battle
```
mrestuccia1: I think you should make sure that this is true since if this isn't true you would have to modify the function.

src/cli/src/operations.c
```diff
@@ -574,12 +574,14 @@ char *battle_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
+    
+    return "Beginning battle.";
 
-    /* TODO: do we need to check if npc_battle struct initialized? 
-     *       should we call run_battle or something else? do we need a return value? */
+    /* TODO: do we need to check if npc_battle struct initialized? */
```
mrestuccia1: For the final version (the one that will be merged into the main branch), I don't think we should keep to-dos listed here. Also, it is not clear here which struct you are referring to, as there are many battle structs present, but I was not able to find a struct named npc_battle struct (though I'm sure it's present).

src/cli/src/operations.c
bzeng6: I was concerned about this too but this is specifically how the Battle team told me it would work—they're working on the implementation for this initialization of a battle and they said it should automatically happen when the game mode is switched to battle mode.

I've added a member of the Battle team to the review, I guess they can look it over (if they're allowed) and suggest changes if they don't think it's quite right.

src/cli/src/operations.c
bzeng6: True, I can delete these comments when we finally end up merging. 

If you look in the npc_t struct, the very last item in the struct is another struct, called npc_battle_t (in npc_battle.h), which just contains a bunch of relevant info about the npc regarding battling. The function npc_init, which initializes an npc, actually always sets this npc_battle_t item to NULL, no matter the will_fight boolean item in the npc struct, which tells whether or not the npc is willing to battle. So my question here was whether or not it is necessary to initialize an npc_battle struct and add it to the npc in question (if the will_fight is 1 and npc_battle_t struct is still NULL at time of the user requesting battle), and if so, should CLI do it in this function, or will that be dealt with elsewhere?

I asked Battle team this and I think I got the impression that they were dealing with it, so I didn't add any code to deal with it.

src/cli/src/cmd.c
```diff
@@ -124,6 +124,7 @@ int lookup_t_init(lookup_t **t)
     add_entry("PALETTE", palette_operation, NULL, t);
     add_entry("ITEMS", items_in_room_operation, NULL, t);
     add_entry("TALK", talk_operation, NULL, t);
+    add_entry("FIGHT", battle_operation, NULL, t);
```
emiliolevinspage, line 127: So does this mean that cli will call battle_operation() when the user inputs something like "FIGHT Goblin" in regular mode?

src/cli/src/operations.c
```diff
@@ -558,3 +558,30 @@ char *talk_operation(char *tokens[TOKEN_LIST_SIZE], chiventure_ctx_t *ctx)
...
+    if (npc == NULL) {
+        return "No one by that name want to fight.";
+    }
+
+    if (npc->will_fight == 0) {
+        return "%s does not want to fight.", tokens[1];
+    }
+
+    set_game_mode(ctx->game, BATTLE, npc->npc_id); 
+    //Apparently this should automatically call function that initialize and run battle
```
emiliolevinspage: still trying to figure this out but our team can circle back to this if it turns out we need to do more here.

emiliolevinspage (approved):

src/cli/src/operations.c
jacknugent1529: I agree with @mrestuccia1. This should be resolved before merging

jacknugent1529 (commented):
The code looks good! Once you resolve a few of the comments and questions it should be good to merge.

milesgardner (commented):
This code looks good - I think after cleaning up some of the comments it should be ready to merge!

src/cli/src/cmd.c
bzeng6: yes, if it's working properly that is what it should do

milesgardner (approved):
Nice job!  You made the proper changes and I think it's ready to go!

mrestuccia1 (approved):
The code looks good, I believe it is ready to merge

jacknugent1529 (approved):
Nice work with this! It looks good to me

jacknugent1529:
@bzeng6 @fkbad @katyashchukina There seems to be some issues with the tests. Once these are fixed it should be good

jacknugent1529 (approved):
Looks good to me!

jacknugent1529:
Issue Score: Excellent

Comments: Great work on this PR!

PR #1384, opened 2022-05-14
Created new functions to create and free specgraphs and created tests for those functions

williamhu314:
In this pull request, we made new functions to create and free our new specgraph struct and also made tests for these new functions. Specifically, we created the following new functions:

1. specgraph_new
2. specgraph_init
3. specgraph_free

PR #1435, opened 2022-05-17
Fixed warnings associated with specgraph function from pull request #1384

### LINKED ISSUE #1434: Fix warnings related to new specgraph functions
The goal of this issue is to fix the warnings related to the new specgraph functionalities that have popped up from pull request #1384.

williamhu314:
We fixed the warnings that were caused by our new specgraph functions.

Specifically, we recognized that the edges field in the specgraph struct was causing the warnings. This is because we wanted to make edges a 2D array, but accidently made it an int** instead of an int*. We have now changed edges to an int* in all instances in gen_structs.h and gen_structs.c.


williamhu314:
This pull request was closed because I realized there was still a warning not addressed. Pull request #1441 addresses this warning.

PR #1456, opened 2022-05-18
Dsl/to-wdl-refactor: new to_wdl subdirectory

### LINKED ISSUE #1421: dsl: refactor `dsl_parser.py` and `to_wdl` into multiple modules
Currently, new Chiventure features all must be added to these two files and could potentially make these files very long and convoluted in the future. Therefore, we want to divide these files into modules that are more specific to each feature, and create subdirectories for conversion to intermediate dictionary form, and for conversion from the dictionary to WDL/JSON format. (ie. NPC, rooms, items, etc.) 
team members: Jay, Alex, Andi

awliu2:
We first created a new subdirectory, 'src/to_wdl'. In this subdirectory, we have created three new files, 'wdl_room.py', 'wdl_item.py', and 'wdl_game.py'. We have moved each class definition to each file so that in the future, when DSL implements supports features such as NPCs, Battles, etc, they will be placed in separate files. The goal of this is to separate the high-level parsing/conversion function, 'parsed_dict_to_json()', from the rest of the code (which can be hidden at a lower level). Everything has been tested with the example dsl files and works as expected.


baigh1:
Something that immediately stood out to me was that there is a list of property aliases and action aliases at the beginning of each new file you guys made. I would say to reduce complexity/eliminate the amount of text at the beginning of the file, you could consider making a common.py file that you would then import to each.

Overall, I would say that I like the way you split up the file into three and kept them in a single file. Organizationally, I would say this make sense.

Similarly, other defined functions within each file seem to overlap immensely, such as generate_defaults, to_wdl_structure, and to_json. I'm not entirely sure how feasible this is, but finding a way to import them from a common file again would be super helpful in reducing complexity for outsiders coming in!

I think with a few changes, this is ready to merge.

jacknugent1529:
I'll do a full review once @baigh1's changes are discussed or addressed, but one thing to add is header comments at the top of each file. I know this isn't done right now, but as we break it into more modules it becomes more important to clarify the purpose of each. See https://uchicago-cs.github.io/dev-guide/style_guide_python.html#header-comments

awliu2:
We worked on adding some headers yesterday, and will hopefully have a new commit ready soon!

awliu2:
Made changes based on Hira's peer review. Notably moved aliases into its own file. I don't think we can move most of the methods defined in each class into separate files, even though they look similar, as each unique class has very distinct patterns that are difficult to abstract over.

jacknugent1529 (approved):
Looks good to me! For other issues, be sure to pull from dev as soon as possible so merge conflicts aren't too bad.

src/dsl/src/to_wdl/util.py
```diff
@@ -0,0 +1,20 @@
...
+}
+
+ACTION_ALIASES = {
+    "success": "text_success",
+    "failure": "text_fail",
+    "fail": "text_fail"
+}
+
+def to_json(self) -> str: 
+        """ For internal testing only: converts an item to its JSON format """
```
jacknugent1529: I missed a few things. This indentation should be decreased, and you need `import json`. Also, I'm not quite clear on the purpose of this function here

baigh1:
Looks good! Sorry for the delay.

baigh1 (approved):
Approved.

alexlee0422:
We first created a new subdirectory, 'src/to_wdl' with three new files, 'wdl_room.py', 'wdl_item.py', and 'wdl_game.py'. We successfully moved each class definition to each file so that in the future, when DSL implements supports features they will be placed in separate files and prevent it to get too long and convoluted. Everything has been successfully tested and are approved by reviewers. Ready to merge.

src/dsl/src/to_wdl/util.py
awliu2: good catch, this is something I meant to delete from the file and forgot to delete.

jacknugent1529 (approved):
Looks good to me!

jacknugent1529:
Issue Score: Excellent

Comments: Great work with this PR!

PR #1470, opened 2022-05-19
Implement reader effect

### LINKED ISSUE #1340: Implement reader effect
This task will entail implementation of the reader effect, a supporting feature used by conditional skills (a type of complex skill) to check the state of the game, for instance whether a player has x item or if the opponent is "undead". As described in the wiki page, this implementation will inherently be limited in scope because it is extremely difficult to ensure we will cover every single condition a game developer would possibly want to check to then execute the conditional skill.  Therefore we are aiming for a broad, surface level implementation of useful conditions. Further description is available here: https://github.com/uchicago-cs/chiventure/wiki/Skill-Trees-~-The-Difficulty-of-the-Reading-the-World

anchristensen:
This is a pull request for the issue implement reader effect: https://github.com/uchicago-cs/chiventure/issues/1340

This code includes implementation of the reader effect, a supporting feature used by conditional skills (a type of complex skill) to check the state of the game, for instance whether a player has x item or if the opponent is "undead". This implementation is inherently limited in scope because we cannot ensure we will cover every single condition a game developer would possibly want to check to then execute the conditional skill. Therefore we aimed for a broad, surface level implementation of useful conditions which can be further fleshed out later. Further description is available here: https://github.com/uchicago-cs/chiventure/wiki/Skill-Trees-~-The-Difficulty-of-the-Reading-the-World


KabirDubey (approved):
Everything looks good to me. The meticulous nature of accessing the struct data makes me think that some sort of refactoring may be in order down the line (like a lower level module should handle this stuff). Also I think the title of the wiki page should be changed to something that makes it more obvious that this corresponds to "readers" e.g. "Readers in Conditional Skills" or perhaps literally "Explanation of Readers."

jacknugent1529:
@anchristensen @arya32hub @nbilaver, it looks like there are a few merge conflicts. Can you guys resolve these and then do a more general sweep to make sure this integrates with code that has since been merged

anchristensen:
The merge conflicts have been resolved and our code is ready for review!

src/skilltrees/src/reader.c
```diff
@@ -0,0 +1,375 @@
...
+}
+
+/*See reader.h*/
+int attr_reader_effect_init(attr_reader_effect_t* reader,char *value,
+                            int str_len, reader_location_t location){
+    assert(reader != NULL);
+
+    reader->value = malloc((str_len + 1)*sizeof(char));
+    strcpy(reader->value, value);
+    // reader->value = value;
```
jacknugent1529: Remove this comment before the code is merged into dev

src/skilltrees/src/reader.c
```diff
@@ -0,0 +1,375 @@
...
+                                    }
+                                }
+                            }
+
+                        }
+                    }
+                }
+            }
+        }
+        return 1;
```
jacknugent1529, line 180: Yes, this doesn't look great, but I agree it is largely unavoidable. If you want, you might be able to clean up the syntax by using short-circuit logical operators. This also has its own problems, so I don't think there's ultimately a great solution.
```c
return !(ctx->game 
 && ctx->game->battle_ctx
 && ctx->game->battle_ctx->game
 ...)
```

tests/skilltrees/reader_test.c
```diff
@@ -0,0 +1,309 @@
...
+/* Helper ctx object for below tests */
+chiventure_ctx_t *ctx_example(void) 
+{
+    class_t *class = malloc(sizeof(class_t));
+    if (class == NULL) {
+        fprintf(stderr, "ctx_error: memory allocation failed for class\n");
+        return NULL;
+    }
+    char* name = "HUMAN";//(char*)malloc(sizeof(char)*6);
+    // strcpy(name, "HUMAN");
```
jacknugent1529: clean up these comments before merging into dev

jacknugent1529 (commented):
Excellent work! The code is very well written and documented, and the tests are also exceptional. There's just a few things to clean up before it's merged

tests/skilltrees/reader_test.c
nbilaver: Comments have been removed.

src/skilltrees/src/reader.c
nbilaver: Comment has been removed.

src/skilltrees/src/reader.c
nbilaver: I'm fairly certain that this route would still result in the crash issues that this is attempting to avoid, as I think C will attempt to read each of those pointers even if one earlier in the && chain is NULL. This is a moment where lazy evaluation would be handy. So sadly, this eyesore looks like it has to stay.

jacknugent1529: Not as powerful as Haskell certainly, but C does in fact have short circuit evaluation for logical operators (`&&`, `||`). This response explains it well https://softwareengineering.stackexchange.com/questions/201896/what-is-short-circuiting-in-c-like-languages/201899#201899. That said, the syntax I suggested is still confusing, and the `if` version is fine

jacknugent1529 (approved):
Looks good to me!

jacknugent1529:
Issue Score: Excellent

Comments: Fantastic work on this PR! The code was well written and your description is detailed.

PR #1477, opened 2022-05-19
Quests/Add_Tree_Management

JSerf02:
This PR adds tree management functionality to a number of quest functions. Specifically, adding a task now adds all tasks in the task list in the prereq, and completing a task now adds the next tasks in the quest and checks for task completion. Additionally, this PR adds some new tests, removes the redundant `get_parent()` function that was supposed to do the same thing as some other helper functions but actually didn't work, and adds a function to add rewards from a reward struct to the player.


JSerf02:
Also note that this PR also includes the changes from the missions PR and should be reviewed after that one is accepted

include/quests/quests_state.h
```diff
@@ -526,23 +487,23 @@ int prereq_add_quest(prereq_t *prereq, char *quest_id);
...
 
-/* Returns the task's reward item if the task has been completed.
+/* Checks if a task's prereqs are met and if they are, completes the task, 
+ * returning the task's reward on success. After completing the task, checks 
+ * if the task completion also completed the task's quest, adds any new tasks
+ * from the tree if not and accepts the quest's rewards if so.
  *
  * Parameter:
- * - task: pointer to a task
+ * - tree: pointer to a task tree who's immediate task is getting completed
```
elizabethli31: parameters should be task_id and quest_hash

src/quests/src/quests_state.c
```diff
@@ -323,47 +305,36 @@ int compare_tasks(task_t *a1, task_t *a2)
...
- */ 
-task_tree_t *find_parent(task_tree_t *tree, char *id)
+ * - pointer to the tree immediately containing the task, OR
+ * - NULL if task cannot be found along a valid path
+ *
+ * Note: Traversal no longer relies on task completion, so 
+ *       runtime is now O(T) where T is the number of tasks
+ *       in the game
+ */
+task_tree_t *find_task_in_tree(task_tree_t *tree, char *id)
```
elizabethli31: function name is a little confusing and required actuallt reading the code to know what the function was meant to do. I'd assume this would just return a task and not the task and all its branches. Something like find_task_subtree would be more intuitive

elizabethli31 (commented):
All the code looks really good. The new task_tree functionality is much better implemented than before. I left two comments that I think should be addressed before this is pulled into dev. The parameters definitely need to be updated to actually match the function.

philipchenaux (approved):
I'm happy with how this PR looks. A few tricky parts in the implementation, but it looks like you navigated this well and backed it up with some good tests. Well done! I did not see some failed checks, but once those are fixed, I'll be happy to merge!

elizabethli31 (approved):
It looks good with everything merged and the two comments I made last time being changed!

philipchenaux:
Issue Score: Excellent

Comments:
Excellent job on this PR!

PR #1480, opened 2022-05-20
Quests/add room and item task checks

Castel-Jaloux:
As part of our goal to enable comprehensive quest completion status checking in chi venture, we made several important changes to the player, game, and quest modules.

First, we added a string to the player class to store the current room, updating the move_room function (among others) to account for this change in structure. We then added a function to check if missions are complete and a function to update the quest status for all players, which would be necessary to act upon the information regarding a completed quest. We also updated move_room and the function for adding an item to the player's inventory to call the new quest function and manage quests, thereby properly integrating the quests module into chiventure.

In order to enable this functionality, we also needed to move the function to add an item to a player's inventory to the game module to access necessary information. We ensured that this did not cause any circular dependency issues, both in our modules but also the modules of other teams, which we slightly tweaked to comply with the new structure.


include/quests/quests_state.h
```diff
@@ -526,23 +495,23 @@ int prereq_add_quest(prereq_t *prereq, char *quest_id);
...
-/* Returns the task's reward item if the task has been completed.
+/* Checks if a task's prereqs are met and if they are, completes the task, 
+ * returning the task's reward on success. After completing the task, checks 
+ * if the task completion also completed the task's quest, adds any new tasks
+ * from the tree if not and accepts the quest's rewards if so.
  *
  * Parameter:
- * - task: pointer to a task
+ * - tree: pointer to a task tree who's immediate task is getting completed
  * - player: pointer to player completing the task
```
elizabethli31: parameters here are not correct. should be task_id and quest_hash

elizabethli31 (approved):
Everything looks good! I added one minor comment that was also found in another pull request about the comments needing to be changed. Only other comment is that I'm confused what item_list_t is used for (I'm assuming this is for a later issue). Also, I'm not sure if you can make these changes to the other features' modules even though they are very minor details, but that's up to Philip to decide.

JSerf02:
item_list_t is used in completed_mission to store the player's inventory

philipchenaux (approved):
Looks good to merge, quite a tricky one when you have to modify game.h and player struct, but you did well to ensure no slip-ups with the other parts of the code.

philipchenaux:
Issue Score: Excellent

Comments:
Excellent job on this PR!

PR #1484, opened 2022-05-20
Quest/free

jakeziqiwei:
Previously task tree wasn't allocated properly which was causing tests in other branches to fail. There was also no function to free the task trees at all. We have fixed the function that adds tasks to a tree by using the new find_task_in_tree function and revisited how trees are supposed to work.  We also wrote the function to free the trees and wrote a test for it

### DIFF
--- a/src/quests/src/quests_state.c
+++ b/src/quests/src/quests_state.c
@@ -246,7 +246,7 @@ int task_tree_free(task_tree_t *task_tree)
     if (task_tree == NULL){
         return SUCCESS;
     } else {
-    free(task_tree->task);
+    task_free(task_tree->task);
     task_tree_free(task_tree->parent);
     task_tree_free(task_tree->rsibling);
     task_tree_free(task_tree->lmostchild);
--- a/src/quests/src/quests_state.c
+++ b/src/quests/src/quests_state.c
@@ -241,6 +241,20 @@ int task_free(task_t *task)
 }
 
 /* Refer to quests_state.h */
+int task_tree_free(task_tree_t *task_tree)
+{
+    if (task_tree == NULL){
+        return SUCCESS;
+    } else {
+    task_free(task_tree->task);
+    task_tree_free(task_tree->parent);
+    task_tree_free(task_tree->rsibling);
+    task_tree_free(task_tree->lmostchild);
+    free(task_tree);
+    }
+}
+
+/* Refer to quests_state.h */
 int quest_free(quest_t *q)
 {
     assert(q != NULL);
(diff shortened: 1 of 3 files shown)


src/quests/src/quests_state.c
```diff
@@ -373,6 +370,36 @@ task_tree_t *find_task_in_tree(task_tree_t *tree, char *id)
...
+ *
+ * Returns:
+ * - pointer to the desired task, OR
+ * - NULL if task cannot be found along a valid path
+ *
+ * Note: Traversal no longer relies on task completion, so 
+ *       runtime is now O(T) where T is the number of tasks
+ *       in the game
+ */
+task_t *find_task_tree_of_task_in_tree(task_tree_t *tree, char *id)
```
JSerf02, line 389: Something funky is going on with the typing here. This function is supposed to return a task_tree_t while find_task_in_tree is supposed to return a task_t. Also, it doesn't appear that find_task_in_tree is called anywhere as far as I can tell so that function can probably just be deleted.

JSerf02 (changes requested):
The freeing changes from this push look really good, but I'm concerned that the merge conflicts were resolved a little bit weird and may cause issues with other branches. It seems like this branch is really behind many of the other branches and some of the merges didn't work 100% (in fact, I have no idea how it's compiling considering there are some clear type issues, such as on line 418). Is it possible to scrap this branch and copy paste the free changes into an offshoot of the refactor or status enum branches so both the newer branches and then afterward these free changes can be pushed with less merge conflict issues? The refactor and status enum branches should already be up to date with all of the most recent changes.

If you do opt to keep this branch, find_task_tree_of_task_in_tree needs to return a task tree and find_task_in_tree needs to be defined after that function and needs to return a task. All calls to both functions (I think only the first one is called even though the second one should be called at some point too) need to be updated as well to have the correct types

src/quests/src/quests_state.c
```diff
@@ -240,6 +240,20 @@ int task_free(task_t *task)
     return SUCCESS;
 }
 
+/* Refer to quests_state.h */
+int task_tree_free(task_tree_t *task_tree)
```
JSerf02, line 244: Looks good! Nice job catching this error and adding in this function!

src/quests/src/quests_state.c
```diff
@@ -325,45 +339,65 @@ int compare_tasks(task_t *a1, task_t *a2)
...
- */ 
-task_tree_t *find_parent(task_tree_t *tree, char *id)
+ * - pointer to the tree immediately containing the task, OR
+ * - NULL if task cannot be found along a valid path
+ *
+ * Note: Traversal no longer relies on task completion, so 
+ *       runtime is now O(T) where T is the number of tasks
+ *       in the game
+ */
+task_tree_t *find_task_in_tree(task_tree_t *tree, char *id)
```
JSerf02, line 357: This function shouldn't exist anymore as far as I can tell and it also has the wrong type

src/quests/src/quests_state.c
```diff
@@ -381,14 +415,14 @@ int add_task_to_quest(quest_t *quest, task_t *task_to_add, char *parent_id)
         quest->task_tree = tree;
         return SUCCESS;
     }
-    tree = find_parent(quest->task_tree, parent_id);
+    tree = find_task_tree_of_task_in_tree(quest->task_tree, parent_id);
```
JSerf02, line 418: This line doesn't type check and shouldn't compile. It should be correct once find_task_tree_of_task_in_tree is fixed though

src/quests/src/quests_state.c
```diff
@@ -535,7 +539,7 @@ int add_quest_to_hash(quest_t *quest, quest_hash_t **hash_table)
 task_t *get_task_from_hash(char *id, quest_hash_t *hash_table) {
     task_t *task = NULL;
     for(quest_t *cur = hash_table; cur != NULL; cur = cur->hh.next) {
-        task = find_task_in_tree(cur->task_tree, id);
+        task = find_task_tree_of_task_in_tree(cur->task_tree, id);
```
JSerf02, line 542: This does type check but only because find_task_tree_of_task_in_tree has incorrect typing.

jakeziqiwei: This is now PR #1531

jakeziqiwei:
This is now #1531

PR #1493, opened 2022-05-20
pullrequesting adapted version of raylib

cyruszzhou:
Multiwindow display was not a core feature of raylib. Nevertheless, raysan had provided a testing version of raylib source code that enables multiwindow display on macOS 11.0 and below. I followed the steps he made for the old version and adapted the changes to the newest version of raylib, which would support the latest OS versions. I have also figured out a format of Makefile that will enable us use this version of raylib, which I will put in other pull requests so we can see it working.


dhirpalshah (approved):

cyruszzhou:
Let's ignore this for now

PR #1499, opened 2022-05-20
Cli/multiple commands

mrestuccia1:
Previously there was no great way to list multiple commands for testing purposes. We added the ability to input any number of commands separated by semicolons, as this is an intuitive method to test the entire game more easily, to make sure it works. We added many unit tests that span many cases and passed them. We believe this code is good to go.


include/cli/cmd.h
```diff
@@ -202,7 +202,7 @@ void cmd_show(cmd *c, chiventure_ctx_t *ctx);
  * Returns:
  * - pointer to command struct, NULL if parse fails
```
bzeng6, line 203: this returns line should be updated

tests/cli/test_game.c
```diff
@@ -82,15 +82,123 @@ Test(game, go)
...
     /* Check that current room has changed */
     cr_assert_eq(ctx->game->curr_room, room2,
                  "GO NORTH did not change the current room as expected");
 
+    free(cmd_str);
+    game_free(ctx->game);
+    chiventure_ctx_free(ctx);
+}
+
+/* Creates a sample game and runs the LOOK;go north commands toghether */
```
bzeng6, line 99: "toghether"

bzeng6 (approved):
I think this looks really good! Good job freeing all the mallocs. The only thing is that I think the documentation in the header file of what the new cmd_from_string function does could be more detailed.

mrestuccia1:
After reviewing this PR further, we decided to go in a different direction. In this PR, "AND" and ';' both serve the same function, allowing for multiple commands. However, we decided that 'AND' should be a conditional executer, meaning that if one of the multiple commands fails none of them will be executed. On the other hand, ';' will be an unconditional executer, meaning that even if a command fails, the other commands will still execute.

PR #1502, opened 2022-05-20
Dsl/player class grammar

### LINKED ISSUE #1479: dsl: implement player_class grammar in dsl_grammar.lark for player_class implementation
While implementing player class, the grammar needs to be updated for the conversion to happen

wzeng0:
We've implemented grammar definitions for player_class within the ```grammars/dsl_grammar.lark``` file for player_class. We've defined concise rules to represent an player_class, its location, its properties, any attributes, and its base_stats. We also created a new file called ```examples/dsl/player-class-grammar-ex.dsl``` to test if player_class attributes is properly generated


alexlee0422 (commented):
It would probably be good for you to add header comments in the beginning of the player-class-grammar-ex.dsl file to describe what the file is doing and testing. You should also put the description in one line instead of entering into a new line in lines 6 and 7. Also you defined room C but do not have the info for it so you should also add that. Otherwise looks good!

alexlee0422 (approved):
Looks good now

src/dsl/examples/dsl/player-class-grammar-ex.dsl
```diff
@@ -0,0 +1,28 @@
...
+    long desc: "The shelves filled with books from all over the world, containing vast amounts of knowledge"
+    connections: " "
+
+# testing: player class with id of knight. Testing if attributes and base_stats are generated correctly
+PLAYER_CLASS KNIGHT
+    short desc: "this a knight"
+    long desc: "he is out to save the world"
+    attributes: 
+        noble set TRUE
+    base_stats:
```
jacknugent1529: I'm slightly concerned about consistency within the language. The goal is to create a simple language for game authors, so the difference between "short desc" separated by spaces and "base_stats"/"PLAYER_CLASS" with the underscores might be confusing. This isn't specifically requesting a change, but I'm wondering what your thoughts are for the design of the dsl.

src/dsl/grammars/dsl_grammar.lark
```diff
@@ -25,12 +35,17 @@ location: phrase
 
 phrase: STRING+
 
+bool: phrase
```
jacknugent1529: `phrase` seems like it might be a bit general here. If this represents a boolean value it might make more sense to match only "true" and "false"

jacknugent1529 (commented):
It looks good! There's a few points of clarification I mentioned. Also, based on the grammar, it looks like rooms and classes can be written in any order. In the example file, I think it would be useful to have a class before a room to verify that this works.

src/dsl/grammars/dsl_grammar.lark
```diff
@@ -10,6 +11,15 @@ room: _NL "ROOM" id (property | connections)* item*
 item: _NL "ITEM" id "IN" location property* action*
     | _NL "ITEM" id property* action*
 
+
+class: _NL "PLAYER_CLASS" id property* attribute* base_stats*
+
+attribute: _NL "attributes:" attribute_ids*
+attribute_ids: _NL id "set" bool
+
+base_stats: _NL "base stats:" stat_setting*
```
jacknugent1529, line 20: The `stat_setting*` implies there could be one or more `stat_setting`'s. Is this intentional, and would it make sense that all of them would use "CURRENT" and "MAX" (i.e. are these the only possible fields for a `stat_setting`)?

baigh1: I think what was meant there was that we can have multiple types of stats identified by their id (i.e. "Health" or "Mana") and each of these has a CURRENT and a MAX, nothing else. Is there anything specific you would like to see changed?

jacknugent1529: Got it, I just wanted to clarify the nature of the stats

jacknugent1529 (approved):
Looks good to me!

jacknugent1529:
Issue Score: Excellent

Comments: Great work on this PR! When receiving feedback from a reviewer, try to directly respond to their comments rather than just making the changes silently.

PR #1508, opened 2022-05-21
Pullrequesting quest_graphics struct, functions, and tests

cyruszzhou:
We have written struts to store and represent quest graphics. We have also written new, init, and free functions of them. We tested those functions and they are indeed working properly. We hope to pullrequest these codes and will soon pullrequest the draw function.


cyruszzhou:
Makefile issue, restarting a pull request

PR #1408, opened 2022-05-16
Playerclass/Add skill trees to prefabricated player classes and create testing for them

vanolson:
There were a number of `TODO` if statements in class_prefabs.c, which were meant to be filled with skill trees for a specific prefabricated player class. The only one that had already been created was Warrior. We created these missing prefab skill trees, notably, Bard, Mage, Ranger, Rogue, and Wizard, each of which being a three-tiered linear skill tree, with increasing damage dealt by each increasing skill. We also wrote tests to make sure that these skill trees were being properly created an assigned  in test_class_prefabs.c. In addition, we deleted several obsolete classes that are no longer part of the prefabs. The skill trees team will likely be working on bringing back these deleted classes, but since they aren't fully implemented as classes, we also decided to remove their unfinished skill trees at this time.

PR #1522, opened 2022-05-22
Update class_prefabs.c with functions to create prefabricated skilltrees

### LINKED ISSUE #1066: Update class_prefabs.c with functions to create default weapons and prefabricated skilltrees
Currently, `class_prefabs/c` contains two main functions: `class_prefab_new()` and `class_prefab_add_skills()`. The former function creates a new prefab class with a short and long description, and base stats. The latter function takes in a class, checks if it is one of the prefab classes, and adds a skill tree, although it doesn't actually add any skill trees yet. I would like to:
 - Complete the `class_prefab_add_skills()` function:
     - Create skilltrees for the prefab classes 
     - Document them on the wiki
     - Add these skilltrees into the code
     - Add tests
 - Add a `class_prefab_add_items()` function which adds some default items to each class (e.g. a wand for the wizard, a rusty sword for the warrior)
     - With the same steps as above
 - Finally, I would like to separate the existing `class_prefab_new()` function so that it is possible to create classes which only contain a short and long description. 
      - Similarly, I would like to create a function where the dev can specify which fields of the prefab class they would like filled out, which will be more relevant as there are more ways to prefab classes (i.e. prefab items)

KabirDubey:
`class_prefabs.c` contains two main functions: class_prefab_new() and class_prefab_add_skills(). The former function creates a new prefab class with a short and long description, and base stats. The latter function takes in a class, checks if it is one of the prefab classes, and adds a skill tree, although it doesn't actually add any skill trees yet. In issue #1066, we:

1. Completed the class_prefab_add_skills() function:
2. Create skilltrees for the prefab classes
3. Documented them on the wiki
4. Added tests for (2)

This work was partially done by the player class team, see PR #1408. The collaboration and contribution from skilltrees team is documented in issue #1066 comments.


anchristensen (commented):
This looks pretty good! I like the long descriptions you picked. I do have one question though; in testing the Druid, Knight, and Elementalist classes, you only checked that the first skill name was correct with cr_assert_str_eq (e.g. line 203 of test_class_prefabs.c). Is there a reason that the other skills (each class was initialized with ~3 skills) weren't also checked/is this important? Looking at the check_skill_presence function, doesn't this already check whether initialized skills match the expected list? Otherwise everything looks good from what I can tell.

vanolson (changes requested):
This looks great! I really like the descriptions and skills you chose. There are a couple changes I'd suggest before you merge. First, in lines 287 and 303, the long_desc of elementals and knight, there's a space at the start of their string that I assume shouldn't be there. In addition, in some of these descriptions of the prefab class and the skill trees, the string goes beyond 80 characters in a line, so I'd recommend splitting up the strings for style reasons. Also, the other classes' skilltrees have a little more variation in damage, rather than just the 3, 12, 25, progression, so I would suggest adding a little variation in there like there are in the first 5. In addition, and this isn't something strictly necessary, but that could be good: all of the skill trees you added have the same min_xp progression (100 -> 200 -> 400) for unlocking it, and I think it could be good to add some slight variations in there just for the sake of making the classes different. But that's just a suggestion. All in all, looks good!

KabirDubey:
Re: anchristensen

Thanks for your review. 

> I do have one question though; in testing the Druid, Knight, and Elementalist classes, you only checked that the first skill name was correct with cr_assert_str_eq (e.g. line 203 of test_class_prefabs.c). Is there a reason that the other skills (each class was initialized with ~3 skills) weren't also checked/is this important? Looking at the check_skill_presence function, doesn't this already check whether initialized skills match the expected list? Otherwise everything looks good from what I can tell.

I think you're correct that the `check_skill_presence` function verifies that every skill matches the expected skilltree. For reference, here are lines 69-77 of `test_class_prefabs.c`. 

```
/* Checks whether skill fields are present, and whether the initialized skills 
 * match the expected list */
void check_skill_presence(class_t* c, int num_skills, char** names) {
    cr_assert_not_null(c->skilltree, "failed to initialize skilltree");
    cr_assert_not_null(c->starting_skills, "failed to initialize skill inventory");

    for(int i = 0; i < num_skills; i++)
        cr_assert_str_eq(c->skilltree->nodes[i]->skill->name, names[i], "failed to add skill");
}
```
I think the reason that we have a separate line that verifies if the first skill is correct is that we want to flag a distinct/special error message for when the _first_ skill in the list fails to initialize. In the original conceptualization for these examples, the first skill played a special role in linking to the other skills (i.e. the skills are deployed according to the ordering of the list (e.g. the warrior can't do a triple slash before it does a double slash (line 164, `test_class_prefabs.c`))).

anchristensen:
In response to your reply, it does make sense then that you have checked all the skills and also that it was initialized properly. Therefore aside from some minor fixes that @vanolson mentioned, like some spaces at the beginning of descriptors and breaking up some 80+ char lines, everything looks good to me!

KabirDubey:
Re: vanolson

Thank you for your review.

> First, in lines 287 and 303, the long_desc of elementals and knight, there's a space at the start of their string that I assume shouldn't be there. 

Right, fixed.

> In addition, in some of these descriptions of the prefab class and the skill trees, the string goes beyond 80 characters in a line, so I'd recommend splitting up the strings for style reasons. 

Fixed

> Also, the other classes' skilltrees have a little more variation in damage, rather than just the 3, 12, 25, progression, so I would suggest adding a little variation in there like there are in the first 5. 

Right, I changed the druid, knight, and elementalist classes damage numbers.

> In addition, and this isn't something strictly necessary, but that could be good: all of the skill trees you added have the same min_xp progression (100 -> 200 -> 400) for unlocking it, and I think it could be good to add some slight variations in there just for the sake of making the classes different. 

Okay, I made every class have difference min_xp levels.

vanolson (approved):

anchristensen (approved):

jacknugent1529 (approved):
Nice work! It looks good to me

KabirDubey:
Closing statement: Arya initiated the task. Both met with player class. Skill tree example writing was evenly split between Kabir and Arya. Kabir updated the wiki and addressed reviews.

jacknugent1529:
Issue Score: Excellent

Comments: Excellent work on this PR! The opening statement is very detailed and the reviews are also very well done.

PR #1523, opened 2022-05-22
Quests/add status enum

JSerf02:
This PR adds a quest status enum so quest status values are easier to interpret (before, they were just ints from -1 to 2), adds a new `update_task()` function that calls `accept_reward(complete_task())` with the proper arguments, and modifies the `complete_task()` function to remove tasks from sibling paths to the completed task from the player's task list so the player becomes committed to their current path on the quest.


elizabethli31 (approved):
Everything looks good. The enums are descriptive and concise and all the quest functions were updated accordingly. I also like the new create sample functions - makes testing a lot easier. I like the addition of update_task() which was really intuitive to follow

include/quests/quests_state.h
```diff
@@ -532,28 +536,68 @@ int remove_quest_in_hash(quest_hash_t *hash_table, char *quest_id);
...
+ *
+ * Parameter:
+ * - quest_ctx: The quest_ctx to be freed
+ * 
+ * Returns:
+ * - SUCCESS if freed successfully, FAILURE if an error occured
+*/
+int quest_ctx_free(quest_ctx_t *quest_ctx);
+
+/* Checks a task for completion and accepts th reward if it is
```
elizabethli31: minor typo

include/quests/quests_state.h
```diff
@@ -505,17 +505,28 @@ int prereq_add_task(prereq_t *prereq, char *task_id);
...
 int remove_quest_in_hash(quest_hash_t *hash_table, char *quest_id);
 
+/* Removes a task from a player hash table
+ *
+ * Parameter:
+ * - pointer to a player task hash table
+ * - quest ID, 
+ * 
+ * Returns:
+ * - 0 if the removal was failure, 1 if successful
```
philipchenaux: This and the function above should ideally stick to the convention of returning the macros FAILURE or SUCCESS instead of 1 and 0.

philipchenaux (changes requested):
As Elizabeth mentioned, I agree that the new enum is very helpful. I commented about a minor issue with return values, and also please address Elizabeth's comment about the typo.

include/quests/quests_state.h
Castel-Jaloux: I just made the edits, and it turns out that the actual functions were already returning SUCCESS or FAILURE - for some reason the header comments indicated differently. At any rate, both that issue and the typo are fixed now.

philipchenaux (approved):
Looks good to merge, great job!

philipchenaux:
Issue Score: Excellent

Comments:
Excellent job on this PR!

PR #1527, opened 2022-05-22
Quest/refactor

### LINKED ISSUE #1349: quest/Refactor quest_state
The current quest_state.c is not the best and it is really confusing for people to understand the data structs and code. We will be refactoring it to make it more user friendly.

elizabethli31:
Previously, the quest feature was divided into two modules: quests_structs and quests_state making it very unorganized and lengthy. We refactored the code into prereqs, tasks, quest, quests_hash, and quests_state. Prereqs, tasks, and quest deal with the structs and initialization functions while quests_hash handles anything related to changing the quest or player hash table. Quests_state holds all the quest functionality functions. There is also a samples module which is used for testing (shared across all the test modules). All tests were refactored accordingly and this branch is merged with add_status_enum.


src/quests/CMakeLists.txt
```diff
@@ -1,8 +1,14 @@
-add_library(quests
-            src/quests_state.c)
+cmake_minimum_required(VERSION 3.10)
 
-target_include_directories(quests PRIVATE src/)
```
JSerf02, line 4: I don't really understand what most of the CMake changes do and why some of these lines were replaced, but everything seems to work and build correctly so it's all good with me

JSerf02 (approved):
This is a spectacular pull request that fixes many of the problems we've been dealing with this quarter! Looking through the changes, everything appears to be fine, since this push mainly restructures existing code instead of adding new functionality (though it is DEFINITELY a necessary restructure). My one comment is that the test suite names and test names are a little bit odd, particularly in test_quests_hash.c, but this is extremely minor and doesn't really matter.

include/quests/quests_hash.h
```diff
@@ -0,0 +1,207 @@
...
+int add_task_to_player_hash(task_t *task, quest_ctx_t *qctx);
+
+/* Removes a quest from a hash table
+ *
+ * Parameter:
+ * - pointer to a quest hash table
+ * - quest ID, 
+ * 
+ * Returns:
+ * - 0 if the removal was failure, 1 if successful
```
philipchenaux: Is there any reason why the return values here are 1 or 0 instead of FAILURE and SUCCESS?

philipchenaux (changes requested):
This is an excellent PR, very well written and must have taken a lot of work. I added one minor comment that should be resolved, but besides that I am happy with how this looks.

include/quests/quests_hash.h
jakeziqiwei: Fixed!

philipchenaux (approved):
Great work!

philipchenaux:
Issue Score: Excellent

Comments:
Excellent job on this PR!

PR #1534, opened 2022-05-23
Cli/conditional mult cmds

### LINKED ISSUE #1396: Implement multiple user commands in a single CLI command
team: cli
members: Miles, Matteo, Kate

task: add the ability for the command line to take multiple commands at once i.e. "go south, go east."  The commands would be separated by a symbol such as a comma or semicolon, and the parser would take multiple commands (more than 2).

This functionality would be very useful, as other teams may want to test their changes, and won't have to type in each command individually.  Instead, they would be able to do their entire test in one command separated by semicolons.

mrestuccia1:
Currently, there is only one way to input multiple commands, and that is the "and" operator. In this pull request, we added an extra tool to input multiple commands, via the ';' character. We changed the "and" operator to be conditional, meaning that if we input "look and go south", the command "go south" will only execute if "look" is successful. On the other hand, the ';' operator will always execute both commands, regardless of if one fails. This will be useful, as there may some commands that you want to execute no matter what, while some others you might only want to do if the previous command succeeds

Team: Matteo, Miles


include/cli/cmd.h
```diff
@@ -202,7 +202,7 @@ void cmd_show(cmd *c, chiventure_ctx_t *ctx);
  * Returns:
  * - pointer to command struct, NULL if parse fails
```
bzeng6: update the return value in header file if needed

src/cli/src/cmd.c
```diff
@@ -235,23 +235,38 @@ cmd *cmd_from_tokens(char **ts, lookup_t **table)
...
-        LL_APPEND(ctx->cli_ctx->command_history, new_command);
-    }
+    char* currcmd;
+    cmd** actions;
+    actions = (cmd**)malloc(sizeof(cmd) * TOKEN_LIST_SIZE);
+    int count = 0;
     
-    char **parsed_input = parse(s);
-    if (parsed_input == NULL)
+    // Tokenizes input by "and" string
```
bzeng6, line 254: Do you think separating commands by "and" string might prove problematic for kind3 actions commands (see operations.c) like "take chair and book", wherein the same action is executed on two separate items?

src/ui/src/print_functions.c
```diff
@@ -186,19 +186,39 @@ void print_cli(chiventure_ctx_t *ctx, window_t *win, int *retval)
          * it to be executed
          */
         tokenized_cmds* temp;
+        // Tokenized commands separated by ';'
```
bzeng6, line 189: What is the functional difference of splitting up command line input by ";" and by "and"? Do you think you need to test whether it works if there are both ";"s and "and"s in the input string?

src/cli/src/parser.c
```diff
@@ -73,16 +73,16 @@ tokenized_cmds *parse_r(char *input)
     char **save_ptr = &input;
     
     //puts tokenized segments of the inputted string in 
-    //the command line with a delimiter "AND" into a 
+    //the command line with a delimiter ";" into a
```
bzeng6, line 87: You should document this change in the header file, too.

src/cli/src/parser.c
```diff
@@ -73,16 +73,16 @@ tokenized_cmds *parse_r(char *input)
     char **save_ptr = &input;
     
     //puts tokenized segments of the inputted string in 
-    //the command line with a delimiter "AND" into a 
+    //the command line with a delimiter ";" into a 
     //utlist until the end of the string
-    char *token = strtokstr_r(input, "AND", save_ptr);
+    char *token = strtokstr_r(input, ";", save_ptr);
```
bzeng6, line 89: Do you think instead of changing the implementation of parse_r to separate commands by ";" instead  of "and", you could have used parse_r as it was in your cmd_from_string function? I'm mostly just wondering why you decided to switch from using ";" as the delimiter to using "and" in cmd_from_string.

bzeng6 (commented):
This looks really good, but I'm a little confused about the reasons for some of your changes. 

Also, I get this when I try it in chiventure—is this expected behavior?
Screen Shot 2022-05-22 at 11 33 33 PM

src/cli/src/parser.c
mrestuccia1: I changed it because the way the functions where written, it was more convenient to have parse_r separate by ':', since it creates a linked list, and cmd_from string tokenize by "AND", and it creates an array. That way, I can iterate through the array for each member of the linked list, and make sure that the commands before in the array did not fail. This makes it such that "and" commands only execute when there is no error, while ';' commands execute without exception.

src/ui/src/print_functions.c
mrestuccia1: The difference is that "and" commands only execute when there is no error, while ';' commands execute without exception.

include/cli/cmd.h
mrestuccia1: You are right, so we have now edited the documentation to reflect the changes in the header file function and the .c file function.

mrestuccia1:
All changes that have been requested were added, the main comment is right, we had problems with our output. These have now been fixed, and have the correct effect. "and" commands will only execute when they both succeed, while ';' commands will execute unconditionally. See example below:

Screenshot (11)

include/cli/cmd.h
```diff
@@ -193,16 +193,17 @@ void cmd_show(cmd *c, chiventure_ctx_t *ctx);
...
+ * Builds a cmd (as defined above) from a string, separates inputs
+ * into tokens separated by "AND"
+ * Almost unneeded, but will stay so that AND, ';' are working commands.
  *
  * Parameters:
  * - array of characters, pointer to chiventure context struct
  *
  * Returns:
- * - pointer to command struct, NULL if parse fails
+ * - arrays of arrays of command structs, NULL if parse fails
```
bzeng6, line 204: array of arrays or array of pointers to cmd structs?

src/cli/src/cmd.c
```diff
@@ -235,23 +235,38 @@ cmd *cmd_from_tokens(char **ts, lookup_t **table)
...
 {
-
-    if (s != NULL) 
-    {
-        command_list_t *new_command = new_command_list(s);
-        LL_APPEND(ctx->cli_ctx->command_history, new_command);
-    }
+    char* currcmd;
+    cmd** actions;
+    actions = (cmd**)malloc(sizeof(cmd) * TOKEN_LIST_SIZE);
```
bzeng6: should it be sizeof(cmd) or sizeof(cmd*)? when you are assigning actions[count] on line 263, you call cmd_from_tokens which i think returns a pointer to a cmd?

bzeng6 (commented):
The documentation is really thorough, but I think one of us, either you or me, is mixing up the return type of cmd_from_string

include/cli/cmd.h
mrestuccia1: You are right, I just changed that.

src/cli/src/cmd.c
mrestuccia1: This is a mistake, it was supposed to be sizeof(cmd*). I juts changed this

bzeng6 (approved):

src/cli/src/cmd.c
jacknugent1529: @mrestuccia1 From what I see, it doesn't appear to have been changed

jacknugent1529: Also, is there a reason you use `TOKEN_LIST_SIZE` here? Is there a reason to prevent people from combining as many commands as they would like (using a linked list implementation)?

src/ui/src/print_functions.c
```diff
@@ -186,19 +186,46 @@ void print_cli(chiventure_ctx_t *ctx, window_t *win, int *retval)
...
         tokenized_cmds* parsed_cmds = parse_r(cmd_string);
+        // For each token
         LL_FOREACH(parsed_cmds,temp)
         {
-            cmd *c = cmd_from_string(temp->cmds, ctx);
-            if (!c)
+            // Get command and potentiallyy tokenize by 'and'
+            cmd **c = cmd_from_string(temp->cmds, ctx);
+            int is_malformed = 0;
```
jacknugent1529: This looks like it would be better represented as a boolean

src/ui/src/print_functions.c
```diff
@@ -186,19 +186,46 @@ void print_cli(chiventure_ctx_t *ctx, window_t *win, int *retval)
...
          */
         tokenized_cmds* temp;
+        // Tokenized commands separated by ';'
         tokenized_cmds* parsed_cmds = parse_r(cmd_string);
+        // For each token
         LL_FOREACH(parsed_cmds,temp)
         {
-            cmd *c = cmd_from_string(temp->cmds, ctx);
-            if (!c)
+            // Get command and potentiallyy tokenize by 'and'
```
jacknugent1529: spelling

src/ui/src/print_functions.c
```diff
@@ -186,19 +186,46 @@ void print_cli(chiventure_ctx_t *ctx, window_t *win, int *retval)
...
+                {   
+                    // if command execution (do_cmd) does not return an error
+                    // and the input is not too long
+                    if (rc != CLI_CMD_CALLBACK_ERROR || !is_malformed) 
+                    {    
+                        // do command, return CLI_CMD_CALLBACK_ERROR if
+                        // there is an error printing a message to the screen
+                        rc = do_cmd(c[i], cli_ui_callback, NULL, ctx);
+                        i++;
+                    // If there is an error in do_cmd
```
jacknugent1529: This comment should probably be moved to the `else` branch (or fixed to match the indentation)

jacknugent1529 (commented):
It looks good and it seems like you've fixed most of the issues already! I just pointed out a fe more minor items to clarify or change.

src/cli/src/cmd.c
mrestuccia1: I fixed the sizeof(cmd*) problem now. As for the TOKEN_LIST_SIZE issue, the way the code was written before required using the cmd_from_strings function, and the best way to edit that without changing too much of the code structure was to return an array of pointers to cmds. Changing this would lead to changing how print_cli works, and likely many other functions. As such, I have now changed the code to simply allow for an arbitrary (right now 10000) number of commands, instead of creating a linked list.

src/ui/src/print_functions.c
mrestuccia1: fixed :)

src/ui/src/print_functions.c
mrestuccia1: fixed :)

src/ui/src/print_functions.c
mrestuccia1: This is now fixed, I moved it into the else branch.

src/cli/src/cmd.c
jacknugent1529: I see. In that case, you should use define a constant with `#define` instead of just writing 10,000. Also, I would lean towards a much smaller number (could be something like 5), since it seems wasteful to malloc a 10,000 element array (of 8 byte pointers) each time in order to just parse the command line

mrestuccia1: Perfect, I just fixed this.

jacknugent1529:
I'm running into a few issues when trying to run the `src/dsl/examples/wdl/demo.wdl` file. Since these don't exist on the dev branch, I assume something is wrong with splitting the commands. Note that the command `GO NORTH` should work since previously went south. Even if it wasn't a valid direction, though, the error message is not correct.
image

mrestuccia1:
The issue referenced has been fixed, the strtok_r function was causing problems, so we switched to the strtokstr_r functions, which was function alike strtok written by previous cs220 years, presumably because they had the same issue as us.

jacknugent1529 (approved):
Looks good to me!

jacknugent1529:
Issue Score: Excellent

Comments: Nice work on this PR! The code was well-written and you responded to the comments well. However, I think more thorough testing should be done earlier in the process so that some of the errors I pointed to would be caught by an earlier review or before the PR is made.

PR #1539, opened 2022-05-23
Cli/levenshtein suggestions

mrestuccia1:
Currently, when a command is not found in Chiventure a function is called to give suggestions. These suggestions are calculated simply by comparing characters. However, this is not a very accurate method, as length can also play a part in similarity. As such, we aim to implement a fuzzy search system (specifically Levenshtein's Distance) in order to better predict what word a user might have been trying to input and suggest it.


src/cli/src/operations.c
```diff
@@ -49,38 +50,62 @@ int compare(char* word, char* action)
...
+{
+    int input_len = strlen(action_input);
+    int action_len = strlen(action);
+    char* tail_inp = action_input+1;
+    char* tail_act = action+1;
+    if (action_len == 0) {
+        return input_len;
+    }else if (input_len == 0) {
+        return action_len;
+    }else if (tolower(action_input[0]) == tolower(action[0])) {
```
majorsylvie: Add a comment here about the tolower function. 
I had not heard of it before and had to look up what it was since I thought it was your own function that somehow wasn't documented.

src/cli/src/operations.c
```diff
@@ -49,38 +50,62 @@ int compare(char* word, char* action)
...
+        int both_tails = levenshtein(tail_inp, tail_act);
+        return both_tails;
+    }else{
+        int tail_one = levenshtein(tail_inp, action);
+        int tail_two = levenshtein(action_input, tail_act);
+        int tail_both = levenshtein(tail_inp, tail_act);
+        return 1 + mini(tail_one,tail_two,tail_both);
+    }
```
majorsylvie, line 133: simple implementation, good job sticking with the definition of the Levenshtien distance.

src/cli/src/operations.c
```diff
@@ -49,38 +50,62 @@ int compare(char* word, char* action)
...
+
 /* 
  * This function returns a string which is the suggestion
- * It finds the suggestion by comparing 
- * each possible action to the input
- * using the compare helper function
- * 
+ * It finds the suggestion by using Levenshtein's
+ * Distance formula linked here:
+ * https://en.wikipedia.org/wiki/Levenshtein_distance
```
majorsylvie: good inclusion of the link to make it easy for people to find your source.

src/cli/src/operations.c
```diff
@@ -49,38 +50,62 @@ int compare(char* word, char* action)
...
  */
 char* suggestions(char *action_input, char** actions)
```
majorsylvie, line 138: add note about what is actually going to be in this inputted array, I personally know that we will be passing the global `actions_for_sug` array, but make it clear for other teams to easily know what is happening.

This is a byproduct of this function not actually being included in the `operations.h` file.

To be a little more formal, write a full `.h`esque header comment to detail your function. Specifically:
* Brief statement of what the `suggestions` function does
* Brief statement on HOW it does it (computing Levenshtien distance)
  * include small note about what the Levenshtien distance is, something as simple as "the distance is a metric to compare string similarity"
* List of your inputs, and please be specific to how suggestions is actually called in the `actions_error_operation` function in `operations.c`
* List of your outputs 

This is because it the function is already not well documented before this, so adding another layer of complexity warrants proper documentation.

src/cli/src/operations.c
```diff
@@ -49,38 +50,62 @@ int compare(char* word, char* action)
...
-            temp = compare(strdup(action_input), strdup(actions[i]));
-            if (temp > initial) 
-            {
-                index = i;
-                initial = temp;
-            }
+            min = temp;
+            index = i;
         }
     }
```
majorsylvie, line 153: Very clean for loop!
Either add a note about how the for loop works inline, or add it to the previously mentioend `.h`esque header comment. 
It is _pretty_ obvious, however I would much rather it be clear to the reader.

tests/cli/test_cmd.c
```diff
@@ -2,6 +2,7 @@
 #include <stdlib.h>
 #include <stdbool.h>
 #include "cli/cmd.h"
+#include "cli/operations.h"
```
majorsylvie, line 5: why include operations.h here?

majorsylvie (changes requested):
Overall, great job! I like the idea and implementation of using the Levenshtein distance since it's a pretty easy algorithim to use when it comes to string similarity. 

All you need now is stellar documentation for this function and I think it can hit the books. 

One thing to note, according to the linked wiki page, there is a more efficient way to calculate the distance involving pre-loading a matrix of distances, that would be a neat thing to mention in some kind of write-up or suggestions for future years doc.

src/cli/src/operations.c
mrestuccia1: I agree, I have added this change now.

tests/cli/test_cmd.c
mrestuccia1: It was included before (in the dev branch) and I accidentally deleted it, so I readded it.

src/cli/src/operations.c
mrestuccia1: This change has been made as requested, a detailed header has been added.

src/cli/src/operations.c
mrestuccia1: This header has now been added, and clearly explains what the function is.

tests/cli/test_cmd.c
mrestuccia1: Added by mistake, is now removed

src/cli/src/operations.c
```diff
@@ -49,38 +50,81 @@ int compare(char* word, char* action)
     return current;
 }
 
+/* Calculates the minimum between three values, 
+   helper to levenshtein function*/
+int mini (int a, int b, int c) {
+    if (a < b && a < c) {
+        return a;
+    }else if (b < a && b < c) {
```
jacknugent1529: style: space after curly brace (fix this throughout the code)

src/cli/src/operations.c
```diff
@@ -49,38 +50,81 @@ int compare(char* word, char* action)
...
+  linked below is here:
+  https://en.wikipedia.org/wiki/Levenshtein_distance
+  Helper funtion to suggestions function.*/
+int levenshtein(char *action_input, char* action) 
+{
+    int input_len = strlen(action_input);
+    int action_len = strlen(action);
+    char* tail_inp = action_input+1;
+    char* tail_act = action+1;
+    if (action_len == 0) {
```
jacknugent1529: Minor style issue, but this check should probably be moved to before you define `tail_inp`. The reason is that if `action_input` is length 0, then `tail_inp` will not be well defined. You obviously don't reference it here, but in case someone adds something in the future, it's better to guarantee that invalid pointers won't be created.

jacknugent1529 (commented):
It looks good! The documentation is very thorough and the implementation looks solid. Are there any tests written for this function, or, if not, how can you verify that it works?

src/cli/src/operations.c
mrestuccia1: I agree, and have just implemented this change

src/cli/src/operations.c
mrestuccia1: Fixed all of these style mistakes

mrestuccia1:
I have added tests, and they all pass. To check my answer, I used this Levenshtein Calculator tool online (https://planetcalc.com/1721/)

tests/cli/test_cmd.c
```diff
@@ -2,6 +2,13 @@
...
+
+char* actions1[29] = {"OPEN", "CLOSE", "PUSH", "PULL", "TURNON", "TURNOFF", 
+                        "TAKE", "PICKUP", "DROP","CONSUME","USE","DRINK",
+                        "EAT", "GO", "WALK", "USE_ON", "PUT", "QUIT","HIST", "HELP",
+                        "CREDITS", "LOOK", "INV", "MAP", "SWITCH", "LOAD_WDL", "NAME", 
+                        "PALETTE", "ITEMS"};
```
majorsylvie, line 40: Change this to `extern char** actions_for_sug;` and do the same for the `NUM_ACTIONS` constant which are both contained in `operations.c`

This way you don't have to update these separately as you test.

https://stackoverflow.com/questions/6792930/how-do-i-share-a-global-variable-between-c-files

majorsylvie (approved):
Thank you for adjusting your documentation! Makes the code super readable :)

jacknugent1529 (approved):
Looks good to me!

jacknugent1529:
Issue Score: Excellent

Comments: Excellent work on this PR! Everything is documented in a detailed and well-organized fashion. One thing to note is to be careful with using force push, since it can overwrite the git history. Methods such as "revert" can often be used instead.

PR #1552, opened 2022-05-23
Quests/wdl

JSerf02:
Added a WDL parsing function for quests and fixed a ton of bugs in quest functions. The function is also tested extensively. One note, though, is that many aspects of quests objects that can be in a WDL file are not required in order for quests to function, but if they are not included, parsing treats it as an error and prints it into stderr. This doesn't prevent the program from running (you can see from the tests that everything works 100% correctly) so it shouldn't really be a problem, but it just means stderr will have a ton of error messages. You can see this from the test output as well.


src/wdl/src/load_quests.c
```diff
@@ -0,0 +1,278 @@
...
+    else if (strcmp(mission_type, "Collect Item") == 0)
+        type = 2;
+    else if (strcmp(mission_type, "Visit Room") == 0)
+        type = 3;
+
+    mission_t *mission = mission_new(target, type);
+    return mission;
+}
+
+task_t *load_task(obj_t *task_obj, game_t *game) {
```
elizabethli31, line 129: load_task should have a header like the other functions

elizabethli31 (changes requested):
WDL is very comprehensive and includes all the objects which are thoroughly included step by step in load_quests.c. This needs to be updated with dev so it doesn't have any merge conflicts. Also load_task should have a header.

include/quests/quests_hash.h
```diff
@@ -193,6 +238,28 @@ int remove_quest_in_hash(quest_hash_t *hash_table, char *quest_id);
...
 int remove_quest_all(quest_hash_t **hash_table);
 
+/* Removes a task from a task hash table
+ *
+ * Parameter:
+ * - pointer to a task hash table
+ * - task ID, 
+ * 
+ * Returns:
+ * - 0 if the removal was failure, 1 if successful
```
philipchenaux: Fix return values for this function and the one below to SUCCESS or FAILURE

philipchenaux (changes requested):
Minor comment on return values again, but once that is addressed this looks good to merge! Also please address Elizabeth's comment.

elizabethli31 (approved):
Changed made and looks good to me!

philipchenaux (approved):
Looks good to merge!

philipchenaux:
Issue Score: Excellent

Comments:
Excellent job on this PR!

PR #1571, opened 2022-05-24
Battle crit change

### LINKED ISSUE #1569: Bring calculate crit out of battle_ai.c and instead have it be calculated in battle_flow to have crits be communicated to the user better.
The goal of this PR is to change our code to calculate critical damage in battle_flow.c. When creating the demo, we found that our print statements do not differentiate if our damage if calculated with or without a crit. We will also make it so that a different message will print if critical damage is successful. This will primarily change three functions, battle_flow_move, enemy_make_move, and print_battle_damage.

emiliolevinspage:
Crit damage calculation has been moved to battle_flow_move() and print_battle_damage adds a clause for moves that crit. This allows the user to see when their move critical strikes.


include/battle/battle_flow.h
```diff
@@ -85,6 +85,14 @@ battle_t *set_battle(battle_player_t *ctx_player, npc_t *npc_enemy,
...
 int calculate_accuracy(int user_accuracy, int move_accuracy);
 
+/*
+* Calculates Critical Damage
+ * Parameters:
+ * - crit_chance : the crit chance of the user using the move
+ * returns: the critical damage multiplier
+ */
+ double crit_modifier(int crit_chance);
+
```
aliu65803, line 97: This is an incredibly minor detail, but to be consistent with the other descriptions you might want to have new lines between the description, parameter, and returns.

aliu65803 (approved):
I am done with my review. It looks good, and I am approving this because it does not seem like there is any code that needs to be changed. The only comment I made was to add a new line in one of the descriptions of the function in battle_flow.h so the description can look like all the others. Great job!

include/battle/battle_flow.h
emiliolevinspage: good catch! updated

lilyehsani (approved):
LGTM! Approved.

nicholaswlee:
This issue is now closed because the PR has been merged. This PR allowed critical damage to be calculated outside of the damage function. As a result, this allows battle_flow_move() to indicate when a move does critical damage rather than just normal damage. This also allowed us to create a new printed statement when a crit is achieved.

lilyehsani:
Issue Score: Excellent

Comments:
Great work on this issue! The title of an issue (or PR) has to convey what the issue will be producing. In the future, try using simple titles with a "Verb-Object" structure that clearly states what you will be doing in the issue (or PR) (e.g., “Design and implement a ‘circle’ module”, “Research libraries for drawing simple geometric shapes”, etc.)

PR #1575, opened 2022-05-24
Skilltrees/integrate skilltrees into view action

### LINKED ISSUE #1510: rpg-skilltrees: meet with action management, integrate skilltrees into a view action
The action management module works with the conditions and effects involved to manipulate a player in chiventure. They designed a fourth "kind" of action (called VIEW) that presumably makes a player view items (further documentation pending).  This utilizes a `do_self_action` function that executes VIEW actions when given various player attributes. From lines 157-174 of the header file:

```
/* A function that executes KIND 4 actions (ACTION <player>)
 *
 * Parameters:
 * - a: An action type struct
 * - p: A player struct
 * - obj: An enum describing what self-related object (STATS, etc)
 *        needs to be acted on
 * - ret_string: A pointer to a string describing the result of the function
 *   - NOTE: THIS STRING IS MALLOCED AND MUST BE FREED BY USERS OF THIS FUNCTION
 *
 * Returns:
 * - 0 upon success, success string as an out parameter
 */
int do_self_action(chiventure_ctx_t *c, action_type_t *a, player_t *p,
                  self_action_object obj, char **ret_string);
```

We will meet with the action management team to discuss what is needed from the skilltrees team to make this possible for skilltrees. We will create functions in the skilltrees module that return skilltree information appropriately for action mangement.

anchristensen:
This pull request will finalize issue #1510. As described in the issue, the action management team described to us a module of theirs which works with the conditions and effects involved to manipulate a player in chiventure. They designed a fourth "kind" of action (called VIEW) that makes a player view items/info. They needed two functions: The first, given a skilltree, returns a char * with basic information about that skilltree. The second takes in a skill and prints its description. These two functions provide high-level information for the described module. This PR includes the code for both functions, as well as tests for both of them.


nbilaver (approved):
Everything looks good to me! The code is clean and well written.

src/skilltrees/src/skilltree.c
```diff
@@ -364,3 +364,62 @@ int inventory_skill_acquire(skill_tree_t* tree, skill_inventory_t* inventory,
...
+    // Calculating size of string we need to store all necessary info
+    for (int j = 0; j < tree->num_nodes; j++){
+        // count size of skill name
+        size += strlen(tree->nodes[j]->skill->name);
+        // add size of prereq_level and current level
+        // uses helper function to calculate their lengths
+        size += count_digits(tree->nodes[j]->prereq_level);
+        size += count_digits(tree->nodes[j]->skill->level);
+        // add size of other text around each info piece below, ex "Skill Name"
+        size += 55;
```
jacknugent1529: I'm a bit confused on why 55 exactly was chosen. Can you expand on this comment a bit?

include/skilltrees/skilltree.h
```diff
@@ -307,4 +307,28 @@ skill_t** skill_prereqs_missing(skill_tree_t* tree,
...
+/*
+ * Prints useful, high-level information about a skilltree
+ * Support for integration with action management for view action
+ *
+ * Parameters: 
+ *  - tree: pointer to a skilltree whose info will be printed
+ * 
+ * Returns:
+ *  - string to be printed
+ */
```
jacknugent1529, line 328: I think this documentation could be expanded a bit to describe what sort of information will be displayed (maybe provide an example)

src/skilltrees/src/skilltree.c
```diff
@@ -364,3 +364,62 @@ int inventory_skill_acquire(skill_tree_t* tree, skill_inventory_t* inventory,
...
+    }
+    char buffer[size];
+    char* temp = buffer;
+
+    for (int i = 0; i < tree->num_nodes; i++){
+        sprintf(temp, "Skill Name: %s, Prereq Level: %d, Current Level: %d\n", tree->nodes[i]->skill->name, tree->nodes[i]->prereq_level, tree->nodes[i]->skill->level);
```
jacknugent1529: Dealing with strings in C is tricky and the cause of many errors, so I think it would be helpful for future teams if you were to elaborate a bit more on how this string formatting is related to the size you calculated above

jacknugent1529 (commented):
It looks good! I think there are a few places where the documentation can be expanded, since manipulating strings in C can easily cause errors

src/skilltrees/src/skilltree.c
anchristensen: It's 55 because the length of this string from line 408: "Skill Name: %s, Prereq Level: %d, Current Level: %d\n" is 55 and since it's using this surrounding info in the buffer for each node, this space needs to be saved. Let me know if that doesn't make sense

include/skilltrees/skilltree.h
anchristensen: This is a great point, I tried to be more specific for both functions on what exactly they return with an example for each of what they would return.

src/skilltrees/src/skilltree.c
anchristensen: I attempted to address this comment by adding an explanation of the 55 char addition to size within a comment in the code. Please let me know if this doesn't sufficiently address your comment or if I should improve this explanation with more details (lines 403-406).

src/skilltrees/src/skilltree.c
jacknugent1529: That makes sense, but I still have a few concerns. For one, it doesn't look like you're explicitly accounting for the null terminator (instead, you're overestimating the total size and in that way accounting for it). 

More importantly, though, when this integrates with action management, should there be a maximum string length? You can imagine the case where an enormous skill tree which creates an incredibly large string. If you haven't, I think you should clarify with action management how this would work. You could also change the arguments to the function so that action management provides you with a buffer and a buffer size, and you fill the buffer until the tree is complete or the buffer is full

anchristensen: I see what you're saying, that makes sense. Action management agreed with what you said about taking in a buffer size and not exceeding that to avoid the case of an incredibly large string; however, they said that they wanted to avoid passing in an actual buffer to the function. Specifically they said, "I think the previous teams avoided passing our buffer string pointer around too much to avoid potential errors, so we might stick with that. So just a string return value would be great. We can definitely pass in our buffer size though, so your parameters would be something like skilltree_t tree, int buffer." Assuming you agree with this decision, I will update the function and tests accordingly, and will comment again once the function and tests are updated. I will also update the size 55 addition to take into account the null character but not the %d/%s marks since these don't count.

jacknugent1529: Ok, that sounds good. What I meant for accounting for the null terminator, is that it should likely go outside the loop. Use the loop to calculate the characters in the string, and then afterwards add one for the null terminator. Also, I noticed you were using `buf` to denote the length (I think). Could you change this to something slightly more descriptive, like `max_buf_len`?

anchristensen: I followed your advice on the more descriptive buf length and the null terminator calculation. The function is now complete and passing tests--I added another test to check for when the given skilltree description would exceed the max_buf_len and it passed. One other major change happened; action management let me know that they wanted me to change the skill_description function to take in a skilltree and a skill name (instead of just the name which was what it used to be), and print out the skill description of the skill within the given tree which matches the given name. This means that you will probably have to review this function again since it is very different from the initial implementation.

jacknugent1529: Alright, that's fine. @nbilaver should re-review then as well

src/skilltrees/src/skilltree.c
```diff
@@ -364,3 +364,116 @@ int inventory_skill_acquire(skill_tree_t* tree, skill_inventory_t* inventory,
...
+    } else {
+        char buffer[max_buf_len];
+        char* temp = buffer;
+
+        int count = 0;
+        int curr = 0;
+        
+        for (int i = 0; i < tree->num_nodes; i++){
+            // count of current node size
+            curr = 49 + strlen(tree->nodes[i]->skill->name) +
```
jacknugent1529: The 49 here is different than the 47 on line 407. Is this intentional?

I think the best way to avoid confusion is at the beginning of the function to define `skill_desc_template = "Skill Name: %s,..."` and then let the next line be `template_len = strlen(skill_desc_template)` (I know this overestimates it slightly, but that's fine in this case). Then, if I understand correctly, you can use that variable on lines 407 and 433

src/skilltrees/src/skilltree.c
```diff
@@ -364,3 +364,116 @@ int inventory_skill_acquire(skill_tree_t* tree, skill_inventory_t* inventory,
...
+
+            sprintf(buffer, "Skill Description: %s\n", tree->nodes[i]->skill->desc);
+
+            char *display = strdup(buffer);
+            return display;
+        }
+    }
+    
+    char buffer[33];
+    sprintf(buffer, "No skill matches the given name");
```
jacknugent1529: This would be more clear with just `char buffer[] = "No skill matches...";` so that you don't have to worry about the length.

jacknugent1529 (commented):
Just a few small changes. Dealing with strings in c is always tricky

KabirDubey:
Closing statement: Kabir initiated the task. Both met with action management. Annika did most of the coding. Kabir helped debug.

src/skilltrees/src/skilltree.c
anchristensen: You're correct, it wasn't intentional. I originally just added 49 in both spots but then realized your comment about the null terminator separation was important here (and then forgot to add this change to both places). Regardless, I used your idea about the template_len variable (this is a great way to make it even more clear to future readers what's happening with the size calculation)

src/skilltrees/src/skilltree.c
anchristensen: Fixed

nbilaver (approved):
After the changes Jack requested, I can't see anything else of note. Approved!

jacknugent1529 (approved):
Excellent work on this PR! It looks good to me

jacknugent1529:
Issue Score: Excellent

Comments: Great work on this PR! The opening summary is very detailed and you responded to all the peer feedback very well.

PR #1578, opened 2022-05-24
Action management/1290 make view stats actions

### LINKED ISSUE #1463: Make "view" action more flexible, coordinate with CLI team to make interface more usable
As it stands, the do_self_action function is inefficient, because it has redundant parameters and uses an enum to check the target of the view action (e.g. STATS in view stats). This makes it difficult for the CLI team to call our function given the input from the user's command line. We need some way to make this smoother. Currently, the only action this function can handle is view, which may change in the future.

HarrisonD123:
We made several changes to our new `do_self_actions` function to make it easier for the CLI team to call and to make it easier to have our function support many new viewing actions.

Our do_self_actions function now only contains the parameters `chiventure_ctx_t *c`, `action_type_t *a`,
`char* target`, and `char **ret_string`
The target parameter decides what the user wants to view. We were able to remove the previous version's `player_t` parameter because we realized that the `chiventure_ctx_t` struct already contains a player.
We also implemented a deeper if-else statement structure to allow for viewing of other interfaces (i.e. skills, quests, inventory, etc.) in the future.


include/action_management/actionmanagement.h
```diff
@@ -170,7 +169,7 @@ int do_item_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *direct,
  * - WRONG_KIND if the action type has the wrong kind, failure string as an out parameter
  * - 6 if conditions for the action haven't been met, failure string as an out parameter
  */
-int do_self_action(chiventure_ctx_t *c, action_type_t *a, player_t *p,
-                  self_action_object obj, char **ret_string);
+int do_self_action(chiventure_ctx_t *c, action_type_t *a,
+                  char* target, char **ret_string);
```
sabinecodes15, line 173:
```suggestion
                  char *target, char **ret_string);
```

include/action_management/actionmanagement.h
```diff
@@ -159,9 +159,8 @@ int do_item_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *direct,
  * Parameters:
  * - c: A context struct encapsulating the shared state in chiventure
  * - a: An action type struct
- * - p: A player struct
- * - obj: An enum describing what self-related object (STATS, etc)
- *        needs to be acted on
+ * - target: An enum describing what self-related object (STATS, etc)
```
sabinecodes15, line 162:
```suggestion
 * - target: A string describing what self-related object (STATS, etc)
```

src/action_management/src/actionmanagement.c
```diff
@@ -346,12 +346,13 @@ int do_item_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *direct,
 
 /* KIND 4
  * See actionmanagement.h */
-int do_self_action(chiventure_ctx_t *c, action_type_t *a, player_t *p, self_action_object obj, char **ret_string)
+int do_self_action(chiventure_ctx_t *c, action_type_t *a,
+                   char* target, char **ret_string)
```
sabinecodes15, line 350:
```suggestion
                   char *target, char **ret_string)
```

sabinecodes15 (changes requested):
A few cosmetic changes, but otherwise looks good!

ecoble:
I am removing the milestone and this PR will remain ungraded since it was mistakenly merged and a new graded PR was opened in its place.

PR #1580, opened 2022-05-24
Action management/do self action errors

### LINKED ISSUE #1455: Update Error Messages For View Stats in Action-Management and CLI Modules
The `"view"` action created by the action management team (and used by the CLI team) needs to have errors indicating when functions are called with the wrong action target (e.g. `view iphone`). These will be implemented in the `do_self_action` function in `actionmanagement.c.` At the conclusion of this ticket, the `do_self_action` function will return the correct error message on receiving the wrong action target, which will be picked up by the CLI team and sent to the command line.

HarrisonD123:
Given that our do_self_action function now has an if-else logic chain to test what the user wants to view (based on strcmp's), we now have a way to error check the item that the user wants to view. This means that if the user types view porta-potty into the command line, we will be able to return an error message. We also have an error message to check for the action itself in the rare case in which CLI passes us an invalid action (currently anything other than "view").

We've also added a new nested if-statement under the conditional that checks for "view", that supports the statement view effects. Upon this command, the do_self_action function will now call the do_stat_effect function from the game-state/stats module. For this specific command, we did not need to collaborate with any other teams. We currently have plans to support even more view actions in the near future in collaboration with other teams.


include/action_management/actionmanagement.h
```diff
@@ -170,7 +169,7 @@ int do_item_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *direct,
  * - WRONG_KIND if the action type has the wrong kind, failure string as an out parameter
  * - 6 if conditions for the action haven't been met, failure string as an out parameter
  */
-int do_self_action(chiventure_ctx_t *c, action_type_t *a, player_t *p,
-                  self_action_object obj, char **ret_string);
+int do_self_action(chiventure_ctx_t *c, action_type_t *a,
+                  char* target, char **ret_string);
```
jbrouwer5: In the function description you have that target is an enum, but then you give it a char* type instead. If it is supposed to be an enum, make sure the function actually takes a value from a defined enum. If it is supposed to take a string just make sure the description aligns with that.

src/action_management/src/actionmanagement.c
```diff
@@ -346,12 +346,13 @@ int do_item_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *direct,
 
 /* KIND 4
  * See actionmanagement.h */
-int do_self_action(chiventure_ctx_t *c, action_type_t *a, player_t *p, self_action_object obj, char **ret_string)
+int do_self_action(chiventure_ctx_t *c, action_type_t *a,
+                   char* target, char **ret_string)
```
jbrouwer5: Make sure to apply any changes to the function header relating to the enum here as well.

src/action_management/src/actionmanagement.c
```diff
@@ -367,14 +368,25 @@ int do_self_action(chiventure_ctx_t *c, action_type_t *a, player_t *p, self_acti
     }
 
     if (strncmp(a->c_name, "view", BUFFER_SIZE) == 0) {
-        if (obj == STATS) {
+        if (strcmp(target, "stats") == 0) {
```
jbrouwer5, line 372: Just checking that CLI will always pass an all lowercase string to target. If not you should let them know that it needs to be lowercase or add the string adjustments in here.

include/action_management/actionmanagement.h
```diff
@@ -170,7 +169,7 @@ int do_item_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *direct,
  * - WRONG_KIND if the action type has the wrong kind, failure string as an out parameter
  * - 6 if conditions for the action haven't been met, failure string as an out parameter
```
jbrouwer5: I don't see the return value of 6 being used in the function.

jbrouwer5 (changes requested):
Once the target parameter is changed in the function header and the other comments are checked over this code is good to go!

ecoble (approved):
Once Jackson's changes are addressed, this looks good to me!

src/action_management/src/actionmanagement.c
HarrisonD123: Yep! CLI changes all strings to lowercase

src/action_management/src/actionmanagement.c
HarrisonD123: Fixed

include/action_management/actionmanagement.h
HarrisonD123: That was left over from a previous function's comments. Just removed that line!

jbrouwer5 (approved):
Looks good!

ecoble:
Issue Score: Excellent

Comments:
Nice job adding error checking and reviewing this PR! Please include closing statements.

PR #1583, opened 2022-05-24
sound: fixing sound module warnings pull request

### LINKED ISSUE #1551: sound: fixing sound module warnings
The sound.c file that was pushed to dev accidentally compiles with warnings. the task is to fix the code in sound.c so the warnings are removed. By removing the warnings while maintaining the structure and integrity of the code, we will be able to write more functions to increase functionality of the sound module while keeping the same coding philosophy used in the sound load/free/play functions. Also, later groups and other groups that want to use the sound module will be able to use these functions without running into unresolved issues within the code.

Assignees: Jason, Victor, Jake

JaStarbuck:
Summary: after merging the load/free/play functions for sounds, it turns out there were warnings in the code. To get rid of these warnings, we needed to create a new union so that different sound types could be returned from and passed between functions. This was very involved and required another deep dive into the demo.c code. This work was necessary, however, because it will allow us to more accurately create new functionality for the sound module and will also be accurate and reliable for other groups that intend to use the sound module. #1551


sangerr (approved):
Given that there are no more warnings or errors occurring, it seems like this pull request is good to be merged. The commit names are very funny but I think that adding a comment to the PR that details how/why you made the changes you did to fix the warnings would be great, as it is difficult to understand that from the commits. After that, I think that it is good to merge!

JaStarbuck:
As Rhea said, this comment will be about what was changed that helped with the error: Originally, we were passing in the string name of the music to the function to play the music. This was incorrect as it needed the music struct to be passed into it. Then, the problem we experienced was that we needed to distinguish between the music struct and sound effect struct being passed into the built in SDL functions. This is why we created the union so that either the music or sound effect struct was contained and could thus be conditionally passed into the SDL functions. Then any other changes were just fixing the functions so that they were returning/taking in the correct structs and unions.

src/sound/src/sound.c
```diff
@@ -22,23 +24,25 @@ int load_wav(sound_t *sound)
         if (!mus)  // error check
         {
             printf("Background music is not loaded\n");
-            return 1;
+            return NULL;
         }
+        loaded->backgroundSound = mus;
```
ecoble, line 29: Union types are accessed with the following syntax: `loaded.backgroundSound`

src/sound/src/sound.c
```diff
@@ -22,23 +24,25 @@ int load_wav(sound_t *sound)
...
     else if (sound->type == SOUND_EFFECT)
     {
         Mix_Chunk *wav = Mix_LoadWAV(sound->name); // loads the sound effect to be played
         if (!wav)  // error check
         {
             printf("Sound effect is not loaded\n");
-            return 1;
+            return NULL;
         }
+        loaded->soundEffect = wav;
```
ecoble, line 39: Same syntax here.

src/sound/src/sound.c
```diff
@@ -47,12 +51,12 @@ int sound_free(sound_t *sound)
     }
     else if (sound->type == BACKGROUND)
     {
-        Mix_FreeMusic(sound->name); // frees resources assocateed with the music
+        Mix_FreeMusic(soundType->backgroundSound); // frees resources assocateed with the music
```
ecoble, line 54: Here as well, and line 59.

src/sound/src/sound.c
```diff
@@ -61,21 +65,26 @@ int sound_free(sound_t *sound)
...
     {
         fprintf(stderr, "no sound objects present\n");
         exit(1);
     }
     else if (sound->type == BACKGROUND)
     {
+        sound_type_t *music = load_wav(sound);
         SDL_Delay(delay); // do a delay if there is one
-        Mix_PlayMusic(sound->name, -1); // play the music
+        Mix_PlayMusic(music->backgroundSound, -1); // play the music
```
ecoble, line 78: Syntax

ecoble (changes requested):
Great job fixing the warning and creating the union type! One small syntax issue with using the union, otherwise looks good to me.

ecoble (approved):

JaStarbuck:
Closing statement: In this pull request the union and slight function restructuring around the union was a success. We accidentally made a change to the union syntax which was changed back during the pull request. Now, these functions will be able to be used in the demo and by future groups integrating sound into chiventure!

ecoble:
Issue Score: Excellent

Comments:
Great job debugging this and creating the new union struct!

PR #1589, opened 2022-05-24
Changing inventory position elements in item_t struct from pointers to unsigned int

### LINKED ISSUE #1588: Changing inventory position elements in item_t struct from pointers to unsigned int
The position elements in the item_t struct were originally for holding the coordinates of an item within an inventory, but the graphics team no longer requires these elements so we are removing them.

KendrickXie:
The position elements in the item_t struct were originally for holding the coordinates of an item within an inventory, but the graphics team no longer requires these elements so we are removing them.


KendrickXie:
Due to some miscommunication, I initially thought the elements for the positions in item_t should be removed entirely. Instead, the graphics team needs them to be unsigned ints rather than pointers, so that change has been made.

HarrisonD123 (approved):
Fairly simple change here. Looks good to me!

ecoble (approved):

ecoble:
Issue Score: Excellent

Comments:
Nice job on a small fix! This will be considered a minor PR since the change was very simple. Please include closing statements for PRs.

PR #1590, opened 2022-05-24
Dsl/graph-validator

baigh1:
We have created the graph verification module located in dsl/src which contains example_graphs.py and graph_validator.py. The former Python file has the various graphs we have designed for testing specific cases (including both normal and edge cases). The latter Python file contains the implementation of the graph validator algorithms. Running this file should produce a list of results identifying whether each graph is valid or not (TRUE or FALSE). We hope for this to get merged into the main branch, and it is a relatively isolated piece of work from the rest of the DSL modules for now.


src/dsl/examples/dsl/new-demo.dsl
```diff
@@ -0,0 +1,99 @@
...
+    attributes:
+        studious set TRUE
+        procrastinator set FALSE
+        inquisitive set TRUE
+    base_stats:
+        health
+            CURRENT 20
+            MAX 100
+        grade
+	        CURRENT 80
```
wzeng0: Indentation error. The Current and Max attributes should have 1 less indent since they belong to the attribute "grade"

src/dsl/examples/dsl/new-demo.dsl
```diff
@@ -0,0 +1,99 @@
...
+
+ROOM storage room
+    short desc: "This is a storage room."
+    long desc: "The storage room has many things in it but only one thing seems to be useful: the dsl documentation. From the storage room, head east to the quad."
+    connections: EAST TO quad
+                 
+    ITEM dsl documentation IN storage room
+        short desc: "A dsl documentation caked in dust."
+        long desc: "The dsl documentation gives you everything you need to know about DSL.”
+    action: CONSUME
```
wzeng0: If you follow your code style from the code lines above, you have a new line between ITEM and action. You should pick a style and stick with it

wzeng0 (approved):
Great job! Other than styling, I don't see any other things you should change. Great descriptions and great story line going on here. Your implementation to check graphs are spot on and you hit every necessary components that make a valid graph. You had adequate tests for each condition and makes sure that they all pass. Again, great job guys.

PR #1603, opened 2022-05-25
modify skill node struct to include class prerequisites

### LINKED ISSUE #1568: Modify skill_node struct to include class prerequisites
Lets say someone was a warrior and a wizard. They can use fire and swords. Maybe they could light their sword on fire. Going along this intuition, it would therefore be cool to provide an implementation for skills that require more than one class. I therefore think it would be beneficial to include a field in the skill node struct for the required classes to use a skill, as well as the number of said classes. We then would need to change the corresponding functions that act on skill_nodes in order to maintain functionality

wellingtj:
Certain skills only make sense by combining skills from multiple classes, so we modified the skill node struct to include class prerequisites. We also changed all corresponding code to include the new class parameter. We also fixed all of the tests to account for these new parameters


wellingtj:
As for a summary of what everybody did, Nik and I (I was sitting right next to him so all the changes were done on his computer) were working on modifying the skill_node struct and relevant functions. Kabir added the new fields to the add_skill function and refactored the calls to add_skill accordingly

tests/skilltrees/skilltree_test.c
```diff
@@ -592,8 +598,8 @@ Test(skilltree_test, inventory_skill_acquire_has)
         2, 5, defusebombeffect);
     skill_t* skill2 = skill_new(1001, ACTIVE, "chop tree", "chops a tree",
         3, 6, choptreeeffect);
-    skill_node_t* bomb_node = skill_node_new(skill1, 0, 12, 47);
-    skill_node_t* chop_node = skill_node_new(skill2, 1,14,  31);
+    skill_node_t* bomb_node = skill_node_new(skill1, 0, 12, NULL, 0, 47);
+    skill_node_t* chop_node = skill_node_new(skill2, 1,14, NULL, 0, 31);
```
katie-gao, line 587: It's good that you guys have updated all the functions to add the new parameters of playerclasses and num of classes. My only question is that while the test functions technically have the two new parameters, it's always just NULL and zero for playerclasses and num of classes. I think the next steps could be to see what would happen once there are playerclasses specified and see how that works.

src/skilltrees/src/skilltree.c
```diff
@@ -10,19 +10,35 @@
 
 /* See skilltree.h */
 skill_node_t* skill_node_new(skill_t* skill, unsigned int num_prereq_skills,
-                             unsigned int prereqlevel,
-                             unsigned int size) {
+                            unsigned int prereq_level, char** playerclasses, int num_classes,
```
katie-gao: Check that this is in the 80 character line limit.

src/skilltrees/src/skilltree.c
```diff
@@ -10,19 +10,35 @@
...
     node->num_prereq_skills = 0;
 
+    node->playerclasses = (char**) malloc(sizeof(char*)*(num_classes + 1));
+    if (node->playerclasses == NULL) {
+        fprintf(stderr, "skill_node_new: memory allocation failed\n");
+        return NULL;
+    }
+
+    for(int i = 0; i < num_classes; i++){
+        node->playerclasses[i] = (char*) malloc(sizeof(char)*(strlen(playerclasses[i])+1));
```
katie-gao: Also, check if this is in the limit - from a glance, it looks kind of long.

src/skilltrees/src/skilltree.c
```diff
@@ -36,6 +52,11 @@ int skill_node_free(skill_node_t* node) {
         free(node->prereqs);
     }
 
+    for(int i = 0; i < node->num_classes; i++){
+        free(node->playerclasses[i]);
```
katie-gao: Nice job freeing -- I would not have remembered doing that

tests/skilltrees/skilltree_test.c
```diff
@@ -20,7 +20,13 @@ Test(skilltree_test, skill_node_new_test)
...
     skill_t* skill = skill_new(1000, ACTIVE, "defuse bomb", "defuses a bomb",
         2, 5, defusebombeffect, NULL);
-    skill_node_t* node = skill_node_new(skill, 0, 24, 76);
+
+    char** classes = (char**)malloc(sizeof(char*)*3);
+    classes[0] = "bard";
+    classes[1] = "beserker";
+    classes[2] = "mage";
+
+    skill_node_t* node = skill_node_new(skill, 0, 24, classes, 3, 76);
```
katie-gao, line 29: Nice job actually implementing the classes function and the num of classes. At the end of the review, I mentioned that it might be beneficial to have a few more tests similar to this to know the changed parameters fully work, but it is nice to know that one does work.

katie-gao (approved):
The changes were really good and I liked the variable names! All the functions were updated and geared towards having playerclasses as detailed in the summary. The only thing is maybe having more tests that has multiple classes beyond the one written to ensure it is all working, but it looks good for now!

tests/skilltrees/skilltree_test.c
wellingtj: these tests are commented out for this years end result of chiventure as they cause a crash

src/skilltrees/src/skilltree.c
wellingtj: I fixed this

src/skilltrees/src/skilltree.c
wellingtj: this is a few characters over, but given that it's one statement I don't know if I can just put it on a new line like that. If I can let me know and I'll for sure take care of that right away though

tests/skilltrees/skilltree_test.c
wellingtj: I have converted this into an issue, but given that we need this pull request for the demo we will be working on soon, I believe that's better left as an improvement for next year. I do agree though with what you're saying

include/skilltrees/skilltree.h
```diff
@@ -63,6 +63,12 @@ struct skill_node {
     // This is an ADDITIONAL pre_req to the lists of skill_nodes preceding the tree
     unsigned int prereq_level;
 
+    //List of playerclasses node belongs to.
+    char** playerclasses;
```
lilyehsani: Move ** over like `char **playerclasses`.

src/skilltrees/src/skilltree.c
```diff
@@ -10,19 +10,35 @@
...
+        return NULL;
+    }
+
+    for(int i = 0; i < num_classes; i++){
+        node->playerclasses[i] = (char*) malloc(sizeof(char)*(strlen(playerclasses[i])+1));
+        if (node->playerclasses[i] == NULL) {
+            fprintf(stderr, "skill_node_new: memory allocation failed\n");
+            return NULL;
+        }
+        strcpy(node->playerclasses[i], playerclasses[i]);
```
lilyehsani: Use `strncpy` instead as it is safer.

chelseaegbarin (approved):

chelseaegbarin:
I think your modifications were warranted and well-executed! I don't have any major suggestions, however for readability/consistency purposes I would recommend renaming the variable playerclasses according to snake case (player_classes) because I noticed you did this for prereq_level but not playerclasses.  Otherwise great job!

lilyehsani (approved):
LGTM! Approved.

wellingtj:
We have included a class prerequisite list called player_classes in the skill node struct. We have also included another parameter in that struct called num_classes to account for the length of that list. The rest of this PR was fixing functions to account for the new parameters in tests and other files

lilyehsani:
Issue Score: Excellent

Comments:
Great work on this PR!

PR #1607, opened 2022-05-25
Fixing ut hash handle Mistake regarding NPCs in Rooms

### LINKED ISSUE #1599: Fixing ut_hash_handle Mistake regarding NPCs in Rooms
One of the bugs I have come across while playing with `battle_move_example_2021` is that NPCs don't seem to respond very well to some move commands, and I was able to trace the issue to discrepancies between the hash_handle being used during some functions dealing with NPCs in rooms. Currently, there are two hash_handles included in the `npc_t` struct, the usual `hh` as well as `hh_room`, which the previous developers designated to be used for dealing with NPCs in rooms. I found that the use of `hh_room` has been neglected in `operations.c`, and intend on fixing that in this issue.

bvvill:
As described in #1599


src/game-state/src/room.c
```diff
@@ -16,6 +16,9 @@ int room_init(room_t *new_room, char *room_id, char *short_desc,
     case_insensitize(new_room->room_id);
     strncpy(new_room->short_desc, short_desc, strlen(short_desc)+1);
     strncpy(new_room->long_desc, long_desc, strlen(long_desc)+1);
+    
+    item_hash_t *items = NULL;
```
srydbergcox, line 20: So, then are the items always NULL when a room is initialized? I am not too familiar with the room struct, but I want to double check that this is correct.

src/npc/src/npc.c
```diff
@@ -275,6 +277,13 @@ int change_npc_hp(npc_t *npc, int change)
     return npc->npc_battle->stats->hp;
 }
 
+/* See npc.h */
+int move_npc(npc_t *npc)
+{
+    return move_npc_mov(npc->movement);
```
srydbergcox, line 283: What is the purpose of this function? It seems that it is redundant since it can just be called using move_npc_mov(npc->movement).

srydbergcox (approved):
Nice work! All of the code here looks good. I don't have any specific changes or comments about your code, I do have a couple clarifying questions that I left just to confirm a couple of things. Other than that though, I think this looks nice and I will approve it.

src/game-state/src/room.c
bvvill: > So, then are the items always NULL when a room is initialized?

Yeah when a room is created through `room_new()` the only inputs are the `room_id` and descriptions, and then those are the only values that get passed on to `room_init()` after the `room_t` is malloc'd, so no other members of the struct are given values upon a room's creation. I believe that C by default is supposed to initialize these other members to `NULL`, but I don't think we should rely on that. Items are added to a room after its creation through `add_item_to_room()`, which is used religiously in in-memory games/examples and tests.

srydbergcox: Ok, perfect sounds good then!

src/npc/src/npc.c
bvvill: Yeah this function is purely for minimizing complexity where possible. I think people are going to more often use functions dealing with higher-level structs such as an `npc_t` (as opposed to its movement struct/member) and a function for moving an NPC given solely it's `npc_t` didn't exist yet, so I figured I'd quickly implement it. It's also a lot easier to read/decipher when used in other functions, and I don't think performance loss is an issue. Also, while right now the function only makes changes to an NPC's movement member, perhaps developers in the future could add more functions inside this one that change something else related to the NPC if they move rooms, such as dialogue or appearance, but this honestly had nothing to do with why I decided to implement `move_npc()`.

lilyehsani (approved):
LGTM! Approved.

bvvill:
comments from #1599:

One of the bugs I have come across while playing with `battle_move_example_2021` is that NPCs don't seem to respond very well to some move commands, and I was able to trace the issue to discrepancies between the hash_handle being used during some functions dealing with NPCs in rooms. Currently, there are two hash_handles included in the `npc_t` struct, the usual `hh` as well as `hh_room`, which the previous developers designated to be used for dealing with NPCs in rooms. I found that the use of `hh_room` has been neglected in `operations.c`, and intend on fixing that in this issue.

Another thing I've done in this branch is optimize `npc_one_move()` in `room.c`, such that it takes a `strcmp()` of the room_id that occupied an NPC's `npc->movement->track` before and after a call to `move_npc`, so that if an NPC stays in the same room, it will just return `SUCCESS` and not have to unnecessarily go through a couple hash functions.

Also, in trying to prevent mysterious seg faults resulting from improperly-initialized hash structs, I've went ahead and explicitly NULL-initialized the `npc_hash_t` member of the `npcs_in_room_t` struct in the `npcs_in_room_init()` function, like so:
```
npc_hash_t *npcs = NULL;
npcs_in_room->npc_list = npcs;
```

in summary: the goal of this pr was to make sure the correct `ut_hash_handle`s were being used regarding items held by NPCs, items in rooms, and NPCs in rooms, as well as making sure explicit NULL hash table initialization was consistent in these areas. This issue/PR achieved this goal and has been closed/merged

lilyehsani:
Issue Score: Excellent

Comments:
Great work on this issue! The closing statement is very thorough. The summary for this issue is too terse. A good rule of thumb is that you should be writing at least one full paragraph in your summary (see https://uchicago-cs.github.io/cmsc22000/project/issue-pr-grading.html#descriptive-summary for examples).

PR #1613, opened 2022-05-25
Cli/strengthen unit tests

mrestuccia1:
Currently, the unit tests in Chiventure do not cover many cases and are thus quite weak. This is cause for worry, as they might be missing errors that could create issues down the line. We aim to bolster these tests by adding more both to the previously written functions of the CLI code, as well as to the new functions we are adding (multiple words in quotes, removing fillers, multiple commands...).


tests/cli/test_game.c
```diff
@@ -94,4 +168,35 @@ Test(game, go)
     free(cmd_str);
     game_free(ctx->game);
     chiventure_ctx_free(ctx);
-}
\ No newline at end of file
+}
+
+/* Creates a sample game, runs the GO "North to south" command and verifies that
```
bzeng6: is it supposed to run 'go "north to south"' or 'go "north to the south"'?

tests/cli/test_parser.c
```diff
@@ -259,35 +270,169 @@ Test(parse_r, two_words_quote)
...
-    char str[] = "\"GO\" EAST";
+    char str[] = "\"south\"";
     char **words = parse(str);
-    cr_assert_str_eq(words[0],"go", "parse() did not create first token");
-    cr_assert_str_eq(words[1],"east", "parse() did not create second token");
+    cr_assert_str_eq(words[0],"south", "parse() did not create first token");
+}
+
+
+//Tests the parsing of input which is fully lowercase.
```
bzeng6: small thing but i think the [style guide](https://uchicago-cs.github.io/dev-guide/style_guide_c.html#comments) suggests to use block comments for function headers, `/* like this */`

tests/cli/test_parser.c
```diff
@@ -259,35 +270,169 @@ Test(parse_r, two_words_quote)
...
+    cr_assert_str_eq(words[1], "at", "parse() did not create second token");
+    cr_assert_str_eq(words[2], "me", "parse() did not create third token");
+    cr_assert_null(words[3],"parse() should point to NULL for empty tokens");
+}
+
+// Tests whether more than 4 words returns NULL as expected
+Test(parse, too_many_words)
+{
+    char str[] = "LOOK LOOK LOOK LOOK LOOK";
+    cr_assert_null(parse(str), "parse() did not create first token");
```
bzeng6, line 348: since it should point to NULL, i think the printed error message if the assertion is false should be changed

tests/cli/test_parser.c
```diff
@@ -259,35 +270,169 @@ Test(parse_r, two_words_quote)
...
+
+// Tests whether more than 4 words in quotes are treated as one token
+Test(parse, too_many_words_in_quotes)
+{
+    char str[] = "\"LOOK LOOK LOOK LOOK LOOK\"";
+    char** words = parse(str);
+    cr_assert_str_eq("look look look look look", words[0], "parse() did not create first token");
 }
 
+//Tests the parsing of an input of 2 words connected by AND which should return the 2 words.
```
bzeng6: I think the header for this test function should be changed, to my understanding it doesn't describe the test properly.

tests/cli/test_parser.c
```diff
@@ -259,35 +270,169 @@ Test(parse_r, two_words_quote)
...
+Test(parse_r, start_and)
+{
+    char* expecting_words[1];
+    expecting_words[0] = " PUSH";
+    char str[] = "AND PUSH";
+    check_comparison(str, 1, expecting_words);
+}
+
+//Tests input in which the and is before any command
+Test(parse_r, and_sandwhich)
```
bzeng6: spelled sandwich wrong but idk if that's intentional lol

bzeng6 (approved):
Looks great

tests/cli/test_game.c
mrestuccia1: It was meant to be GO "North to the South", I fixed it now.

tests/cli/test_parser.c
mrestuccia1: I changed all of the comments from the tests we made from // to the style you suggested.

tests/cli/test_parser.c
mrestuccia1: Changed!

tests/cli/test_parser.c
mrestuccia1: You are correct, I changed this.

tests/cli/test_parser.c
mrestuccia1: fixed

jacknugent1529 (approved):
Looks good to me!

jacknugent1529:
Issue Score: Excellent

Comments: Excellent work with this PR! Testing is always a good thing, especially with integral features such as the cli. This PR looks really good, one nitpick (not a rubric item) is that your commit messages don't provide as much distinguishing information as they could.

PR #1623, opened 2022-05-25
Added sound_new and sound_init functions

### LINKED ISSUE #1622: sound: Add initializing functions for sound type declaration
This issue is created because we found that there are no initialization functions of any kind for `sound_t` module, which comes to be a huge hassle while writing for tests, because the same lines of code will have to be repeated (specifically, `malloc` of `sound_t` and assigning values and pointers within `sound_t`.

jiaoboxuan:
Summary: Existent functions such as `load_wav` and `sound_free` require already allocated `sound_t` structs, and while writing tests, I would have to allocate the structs in separate lines of code, and repeat the process per new test. Therefore, I added two functions, `sound_init` and `sound_new`, descriptions of which are added in `sound.h` file. The code compiles successfully in the workflow run, and these two simple functions could significantly ease the test writing process, as well as improve the readability of the module.


include/sound/sound.h
```diff
@@ -31,12 +31,34 @@ typedef struct {
...
+
+/* Initializes the 
+ * 
+ * Parameters:
+ *      None
+ * Return:
+ *      An allocated sound structure, with name and type
+ */
+int sound_init(sound_t *sound, SoundType type, char *name);
+
```
JaStarbuck, line 63: I like your documentation here!

src/sound/src/sound.c
```diff
@@ -1,6 +1,26 @@
 #include "common/uthash.h"
 #include "sound/sound.h"
 
+/* See sound.h for details */
+sound_t *sound_new(SoundType type, char *name)
+{
+    sound_t *sound = (sound_t*)malloc(sizeof(sound_t));
+    sound->wavBuffer = (uint8_t*)malloc(sizeof(uint8_t));
```
JaStarbuck, line 8: not sure if this needs to be malloced

JaStarbuck (approved):
This looks really good! I've gone through and left comments on some things. I wonder if once you malloc the struct, you actually have to malloc the things inside the struct. Other than that, it looks good!

JaStarbuck (approved):
Oops think I accidentally re-review--approving again

include/sound/sound.h
```diff
@@ -31,12 +31,34 @@ typedef struct {
...
+ *      type: enum SoundType
+ *      name: name of the sound, 
+ *            or more specifically, the path to a sound file.
+ *            Must be already allocated in the heap
+ * Return:
+ *      An allocated sound structure, with name and type
+ */
+sound_t *sound_new(SoundType type, char *name);
+
+/* Initializes the
```
ecoble: Initializes the what? Also it looks like function does have parameters, and it is returning an int.

src/sound/src/sound.c
ecoble: Mallocing this is correct since wavBuffer is a pointer.

ecoble (changes requested):
Nice work! Just a quick change on the header comment.

jiaoboxuan:
Update: changed the `char` structure from pointer form to stack form, and updated the documentation.

ecoble (approved):

jwschroeder21:
Closing statement: sound_init and sound_new were added with descriptions of them in the sound.h file. These functions ease the test writing process and allow  for no re-allocation of the sound_t structs in each process of a repeated test.

ecoble:
Issue Score: Excellent

Comments:
Great work! Nice job identifying the need for new and init functions!

PR #1626, opened 2022-05-25
Adding and Updating Functions dealing with Items in Rooms and held by NPCs so that Duplicate Items Cannot Exist within a Singular Hash Table

### LINKED ISSUE #1614: Adding and Updating Functions dealing with Items in Rooms and held by NPCs
There are currently a few bugs involving items within rooms or held by NPCs, one being that duplicate items are allowed to be added to a room's or NPC's `item_hash_t`, which is an error according to UTHash's documentation. There are also a couple of useful functionalities that have yet to be implemented, such as "getting" an item from an NPC's inventory given the `item_id`. There are also a couple hash implementations regarding items that could use optimization

bvvill:
As described in #1614


include/npc/npc.h
```diff
@@ -258,6 +269,26 @@ int add_item_to_npc(npc_t *npc, item_t *item);
...
+ *
+ * Returns:
+ * SUCCESS upon completion
+ */
+int delete_all_items_from_npc(npc_t *npc);
+
+/* Deletes and frees all items in an NPC's inventory
+ * 
+ * Parameters:
+ *  npc: the npc whose inventory we're freeing
```
junasano: add deleting as well, so becomes "... inventory we're deleting and freeing"
To make your intentions stated in the issue more explicit.

src/game-state/src/game.c
```diff
@@ -21,6 +21,22 @@ game_t *game_new(char *desc)
 
     /* read from the file using interface from WDL team */
 
+    room_hash_t *rooms = NULL;
```
junasano, line 24: Im not sure if it works like this but is there a reason why you stated each variable before hand and set that to NULL?  If not, maybe just go straight into game->all_rooms = NULL; and etc.

src/game-state/src/item.c
```diff
@@ -17,7 +17,14 @@ int item_init(item_t *new_item, char *item_id, char *short_desc,
     case_insensitize(new_item->item_id);
     strcpy(new_item->short_desc, short_desc);
     strcpy(new_item->long_desc, long_desc);
-    new_item->stat_effects = NULL;
+
+    game_action_hash_t *actions = NULL;
```
junasano, line 21: Same thing here with declaring the NULL beforehand.

tests/npc/test_npc.c
```diff
@@ -301,16 +300,14 @@ Test(npc, remove_item_from_npc)
     cr_assert_eq(rc, SUCCESS, "add_item_to_npc failed to "
                  "add an item to npc");
     rc = add_item_to_npc(npc, dup_item);
-    cr_assert_eq(rc, SUCCESS, "add_item_to_npc failed to "
-                 "add an item to npc");
+    cr_assert_eq(rc, FAILURE, "add_item_to_npc added duplicate item to npc");
```
junasano: I feel like testing for duplicate items on the remove_item_from_npc test is redundant. Since it's already being tested in add_item_to_npc and this test is specifically for removing the item rather than adding.

tests/game-state/test_rooms.c
```diff
@@ -120,7 +120,7 @@ Test(room_item, add_duplicate_item_to_room)
     cr_assert_eq(rv, SUCCESS, "item not added to room correctly");
 
     int check = add_item_to_room(new_room, test_item2);
-    cr_assert_eq(check, SUCCESS, "duplicate item not added to room correctly");
+    cr_assert_eq(check, FAILURE, "duplicate item added to room incorrectly");
```
junasano, line 123: isn't this check and the check right below this pretty much doing the same thing?

tests/game-state/test_rooms.c
```diff
@@ -168,16 +168,15 @@ Test(room_item, remove_item_from_room)
     cr_assert_eq(rc, SUCCESS, "add_item_to_room failed to "
                  "add an item to room");
     rc = add_item_to_room(room, dup_item);
-    cr_assert_eq(rc, SUCCESS, "add_item_to_room failed to "
-                 "add an item to room");
+    cr_assert_eq(rc, FAILURE, "add_item_to_room added duplicate item");
```
junasano: Same thing as comment from the other test file. No need to test for duplicate again and since the test is for the remove item test.

src/game-state/src/room.c
```diff
@@ -345,63 +355,38 @@ int auto_gen_movement(npc_t *npc, room_list_t *all_rooms)
...
-    r2 = add_npc_to_room(new_npc_room, npc);
-    r3 = delete_npc_from_room(old_npc_room, npc);
-    if ((r1 == FAILURE) || (r2 == FAILURE) || (r3 == FAILURE))
+    if(npc->movement == NULL)
     {
         return FAILURE;
     }
-    else
+    char *old_room_id = npc->movement->track;
+    assert(move_npc(npc) != 0);
```
junasano: change the 0 to SUCCESS so that it is more explicit

include/npc/npc.h
```diff
@@ -311,6 +342,23 @@ int add_battle_to_npc(npc_t *npc, stat_t *stats, move_t *moves,
...
+ * Returns:
+ * FAILURE: if a move does not occur, 
+ * SUCCESS: if successful move to the next room, with the only exception
+ * being if an indefinite NPC has more than one room in its path, and then
+ * reaches the end of its path, in which case its path direction is flipped,
+ * and its time_ray for the current step in its path is reset, but it does not
+ * technically change rooms.
+ *
+ */
+int move_npc(npc_t *npc);
```
junasano: I'm not really sure what the point of this function is except that it is in the npc.h/npc.c module rather than npc_move.h/npc_move.c. Especially since they are pretty much identical and is only used once so far (and in that place, move_npc_mov can be used).

src/game-state/src/room.c
```diff
@@ -17,6 +17,9 @@ int room_init(room_t *new_room, char *room_id, char *short_desc,
     strncpy(new_room->short_desc, short_desc, strlen(short_desc)+1);
     strncpy(new_room->long_desc, long_desc, strlen(long_desc)+1);
 
+    item_hash_t *head = NULL;
```
junasano, line 23: same thing will NULL here

src/npc/src/npc.c
```diff
@@ -17,6 +17,8 @@ int npc_init(npc_t *npc, char *npc_id, char *short_desc, char *long_desc,
     npc->hostility_level = hostility_level;
     npc->npc_battle = NULL;
     npc->movement = movement;
+    item_hash_t *items = NULL;
```
junasano: NULL once again

src/npc/src/npc_battle.c
```diff
@@ -48,9 +48,7 @@ npc_battle_t *npc_battle_new(stat_t* stats, move_t* moves,
                                 hostility_level, class_type, items,
                                 armor, accessory, weapon);
 
-    if (npc_battle == NULL || npc_battle->stats == NULL ||
-            npc_battle->moves == NULL || npc_battle->class_type == NULL ||
-            npc_battle->items == NULL || check != SUCCESS)
+    if (npc_battle == NULL || check != SUCCESS)
```
junasano, line 51: Is there a reason why you deleted the NULL checks for the other stuff? Because I feel like a battle can't start if those are NULL and just checking to see if npc_battle == NULL isn't enough to check for everything

src/npc/src/rooms_npc.c
```diff
@@ -5,7 +5,9 @@ int npcs_in_room_init(npcs_in_room_t *npcs_in_room, char* room_id)
 {
     assert(npcs_in_room != NULL);
     strncpy(npcs_in_room->room_id, room_id, MAX_ID_LEN);
-    npcs_in_room->npc_list = NULL;
+
+    npc_hash_t *npcs = NULL;
```
junasano, line 9: NULL here too

junasano (changes requested):
Overall, the logic seems good. I left some comments here and there but I'm not 100% sure for some of them so feel free to correct me if you had a different intention.

include/npc/npc.h
bvvill: i agree, will do

src/game-state/src/room.c
bvvill: omg this is actually a major bug lol thank you for pointing this out, since the refactoring of `move_npc_mov()` to simply out `FAILURE` or `SUCCESS`, this statement was not updated to say `assert(move_npc(npc) != FAILURE);`, so I will change that now

src/npc/src/npc_battle.c
bvvill: checking npc_battle for being NULL is checking if the memory allocation of the struct worked, and then checking if `check` equals `SUCCESS` makes sure that the initialization of the struct went okay. In some of the examples/tests, there are many parameters that `NULL` is being passed into, and because of that, the return value for `npc_battle_new()` is NULL, which obviously wasn't the intention of whoever was creating those battle structs in those cases. It's also to stay consistent with the other struct creation implementations throughout chiventure, as I haven't come across any other `..._new()` function that expects every single struct member to not be `NULL` and returns `NULL` if not.

tests/game-state/test_rooms.c
bvvill: Yup you're right, didn't think about that, the second one shall be removed

tests/game-state/test_rooms.c
bvvill: Yeah I agree, I will remove it now

tests/npc/test_npc.c
bvvill: Yeah I agree, removing this as well

include/npc/npc.h
bvvill: Yeah this function is purely for minimizing complexity where possible. I think people are going to more often use functions dealing with higher-level structs such as an `npc_t` (as opposed to its movement struct/member) and a function for moving an NPC given solely it's `npc_t` didn't exist yet, so I figured I'd quickly implement it. It's also a lot easier to read/decipher when used in other functions, and I don't think performance loss is an issue. Also, while right now the function only makes changes to an NPC's movement member, perhaps developers in the future could add more functions inside this one that change something else related to the NPC if they move rooms, such as dialogue or appearance, but this honestly had nothing to do with why I decided to implement `move_npc()`.

src/game-state/src/game.c
bvvill: https://github.com/uchicago-cs/chiventure/issues/1614#issuecomment-1137860606

src/game-state/src/item.c
bvvill: https://github.com/uchicago-cs/chiventure/issues/1614#issuecomment-1137860606

src/game-state/src/room.c
bvvill: https://github.com/uchicago-cs/chiventure/issues/1614#issuecomment-1137860606

src/npc/src/npc.c
bvvill: https://github.com/uchicago-cs/chiventure/issues/1614#issuecomment-1137860606

src/npc/src/rooms_npc.c
bvvill: https://github.com/uchicago-cs/chiventure/issues/1614#issuecomment-1137860606

junasano:
The changes look good. Just make sure to merge with the dev branch since there seems to have been new PRs approved.

junasano (approved):

src/game-state/src/room.c
```diff
@@ -67,11 +70,19 @@ int room_free(room_t *room)
 /* See room.h */
 int add_item_to_room(room_t *room, item_t *item)
 {
-    int rc;
-
-    rc = add_item_to_hash(&(room->items), item);
```
lilyehsani: What is the reason you moved away from using this helper function? I notice that the implementation here is very similar to the implementation for add_item_to_npc, so a helper function makes sense to me.

lilyehsani (changes requested):
Looks really good! One comment.

src/game-state/src/room.c
bvvill: While the helper function is correctly used here, I think that it's worth noting that it assumes the hash handle being used is `hh`, and because there are two possible hash handles for items, I thought it was better to be explicit as to which one was being used. That being said, I think I've become overly paranoid about hash table errors, so if you feel this was an unnecessary change, I will swiftly undo it.

lilyehsani (approved):
LGTM! Approved.

bvvill:
comments from #1614:

There are currently a few bugs involving items within rooms or held by NPCs, one being that duplicate items are allowed to be added to a room's or NPC's `item_hash_t`, which is an error according to UTHash's documentation. There are also a couple of useful functionalities that have yet to be implemented, such as "getting" an item from an NPC's inventory given the `item_id`. There are also a couple hash implementations regarding items that could use optimization

As I said I would do in my opening comment, I've implemented `item_t *get_item_from_npc(npc_t *npc, char *item_id)` in npc.h/.c, which returns an item in an NPC's inventory given the `item_id`, if it exists.

I've also implemented `delete_all_items_from_npc` and `free_all_items_from_npc`, which delete and (delete, but also) free the given NPC's inventory, respectively. The difference in the use of these functions is that solely deleting an NPC's inventory doesn't free the items that were in it, so that the items can still be used in other places or conveniently added to other hash tables, as is done in `transfer_all_npc_items()` (found in room.h/.c)

in `game.c` and `item.c`, I've also went ahead and explicitly initialized every hash table member of the `game_t` and `item_t` structs to be `NULL` when either of those two structs is created. Doing so is incredibly important for ut_hash table maintenance and stability, and can lead to mysterious seg faults if not done properly. The new lines added to `game_new()` are:
```
    room_hash_t *rooms = NULL;
    item_hash_t *items = NULL;
    quest_hash_t *quests = NULL;
    npc_hash_t *npcs = NULL;
    stats_global_hash_t *stats = NULL;
    effects_global_hash_t *effects = NULL;
    class_hash_t *classes = NULL;

    game->all_rooms = rooms;
    game->all_items = items;
    game->all_quests = quests;
    game->all_npcs = npcs;
    game->curr_stats = stats;
    game->all_effects = effects;
    game->all_classes = classes;
```
And the new lines in `item_init()` are:
```
    game_action_hash_t *actions = NULL;
    attribute_hash_t *attributes = NULL;
    effects_hash_t *stat_effects = NULL;

    new_item->actions = actions;
    new_item->attributes = attributes;
    new_item->stat_effects = stat_effects;
```


in `add_item_to_npc()` (in npc.h/.c), I've also explicitly implemented the function instead of using `add_item_to_hash()`, since it assumes the hash handle to be hh and doesn't check that the input parameters aren't NULL.

in `game-state/test_rooms.c` and `npc/test_npc.c`, I've updated the tests regarding adding duplicate items to rooms and NPCs, such that the functions now appropriately return `FAILURE` if a duplicate item is attempted to be added

lilyehsani:
Issue Score: Satisfactory

Comments:
Great work on this PR! The closing statement is very thorough. Two comments:
1. The summary for this PR is too terse. A good rule of thumb is that you should be writing at least one full paragraph in your summary (see https://uchicago-cs.github.io/cmsc22000/project/issue-pr-grading.html#descriptive-summary for examples).
2. When you create a PR, you are requesting a review of the code in the branch you want to merge. As such, you should not push any commits to the branch right after creating the PR; you should only do so in response to a code review. Make sure that, when you create a PR, the code is ready for review and passes all the tests. Merging dev is acceptable, but there are other commits after the creation of the PR and before the review.

PR #1631, opened 2022-05-25
updated call to `do_self_action` in `kind4_action_operation`

### LINKED ISSUE #1565: cli: implement our end of `view quests`
follow our end of the implementation as defined in the documentation of `view quests`

majorsylvie:
AM updated their implementation so we can change the function call in `kind4_action_operation`

Specifically, AM and us have updated the kind4_action_operation to allow for the quests teams group of:
```
view quests
view quests "[quest ID]"
view task "[task ID]"
```
We needed to update the call in order to pass an array of strings that would include all arguments to view


jacknugent1529:
Can you clarify what the expected behavior is? When I try these commands, nothing happens, not even an error message.
image

majorsylvie:
> Can you clarify what the expected behavior is? When I try these commands, nothing happens, not even an error message. image

As of now, just recognition (so not telling your the command is unrecognized)
For view stats, if there is not a player with stats in the context, it will not print anything as of now (which is a decision that game-state made in their implementation)

For the quests trio, quests has yet to finish implementing all their commands, and AM is expecting us to change the call to `do_self_action` which is what we're doing here.

jacknugent1529 (approved):
Looks good to me!

milesgardner (approved):
Looks great!

jacknugent1529:
Issue Score: Excellent

Comments: Nice work with this minor pull request! Everything is well organized and documented.

PR #1639, opened 2022-05-26
sound: add functionality for pausing/stopping sound pull request

### LINKED ISSUE #1464: sound: add functionality for pausing/stopping sound
Summary: As seen in the demo.c file, they include functionality past load/free/play sound. Specifically, it looks like they are able to introduce sound effects, stop music, pause music, etc from the press of a button or through other inputs. In this issue, the goal is to add functions in the sound module that specifically allow the pausing/stopping of music or sound effects. It looks like there are built in SDL functions in the demo that will be helpful in the process, and we will need to decide how and where to implement button pressing in order to pause/stop the sounds. We hope that this will further enhance the chiventure experience and will make it easier and more appealing for other groups to use the sound module in their own projects.

assignees: Jason, Victor, Rhea

JaStarbuck:
Summary: For this pull request, we've made 3 simple functions using functions from the SDL library. It may seem redundant to make these functions since they're so small, it makes it so that other groups and future teams don't have to do their own research and find these SDL functions themselves--they will already be in the sound module. We've also made small edits to the play_sound function to get rid of potentially function breaking code. We realized that freeing the sound in the play function could lead to the sound not playing. Like other pull requests this sprint, these functions haven't been tested and will need to be in the future. The goal of this pull request is so that, like mentioned before, increase ease of use for future groups and other teams.


src/sound/src/sound.c
```diff
@@ -88,4 +88,22 @@ int play_sound(sound_t *sound, int delay)
...
     }
     return 0;
-}
\ No newline at end of file
+}
+
+/* See sound.h for details */
+void stop_sound()
+{
+    int x = Mix_HaltMusic();
```
jwschroeder21: Looks good

src/sound/src/sound.c
```diff
@@ -88,4 +88,22 @@ int play_sound(sound_t *sound, int delay)
...
+}
+
+/* See sound.h for details */
+void pause_sound()
+{
+    Mix_PauseMusic();
```
jwschroeder21, line 102: Looks good

src/sound/src/sound.c
```diff
@@ -88,4 +88,22 @@ int play_sound(sound_t *sound, int delay)
...
+}
+
+/* See sound.h for details */
+void resume_sound()
+{
+    Mix_ResumeMusic();
```
jwschroeder21, line 108: Brilliant work!

jwschroeder21 (approved):

src/sound/src/sound.c
ecoble: If you're not doing anything with the int, do you need to save it? Or maybe you can return it as a signal for success/failure?

ecoble (changes requested):
Small change!

ecoble (approved):
Nice job!

JaStarbuck:
Closing statement: In this pull request we were able to get the function headers for pause/stop/resume in sound.h and the functions in sound.c working. The only edit that was requested was to get rid of an unused variable. We will need to test these functions so that future groups and other teams can use them when they access the sound module.

ecoble:
Issue Score: Excellent

Comments:
Great job researching the SDL library and implementing this!

PR #1640, opened 2022-05-26
Actions fireball and slash was added

### LINKED ISSUE #1638: Adding the actions "fireball" and "slash" to the Chiventure actions list
The custom-scripts team would like to add the actions "FIREBALL" and "SLASH" so that we may include these actions on a WDL file. Without the addition of these actions in the main Chiventure actions list, we cannot load in the WDL file properly. These two actions should be added to the list of actions on the load-actions.c file as well as the header file action-struct.h

jackjackkimkim:
This pull request is to add actions fireball and slash to the list of actions in Chiventure


include/custom-scripts/get_custom_type.h
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * bool_t_get() returns a bool from an object_t struct
+ * Parameters:
+ * - ot: an obj_t
+ * - d: operating data type
+ * Returns:
+ * - bool from ot
+ */
+data_t arg_t_get(object_t *ot);
```
sabinecodes15, line 14: The comment in this header file doesn't match the function declaration.

include/custom-scripts/arg_custom_type.h
```diff
@@ -0,0 +1,64 @@
...
+ * Creates an argument structure with the specified boolean
+ * Helper function for obj_add_arg_bool
+ */
+arg_t *arg_t_init(data_t d, data_type_t t);
+
+/**
+ * Adds an argument structure to the end of the argument linked list
+ * Helper function for all obj_add_arg_<type> functions
+ */
+arg_t *arg_t_add(arg_t *head, arg_t *add);
```
sabinecodes15, line 62: Even though the functions in this file are helper functions, you should still be more descriptive in the header comments. What are the input parameters representing? What is the output?

include/custom-scripts/lua_custom_type.h
```diff
@@ -0,0 +1,30 @@
+#ifndef _LUA_CUSTOM_TYPE_H_
+#define _LUA_CUSTOM_TYPE_H_
+
+#include "obj_custom_type.h"
+
+/**
+ * Helper function used to push the arguments in the linked list
+ * to the Lua vritual stack
```
sabinecodes15:
```suggestion
 * to the Lua virtual stack
```

include/custom-scripts/lua_custom_type.h
```diff
@@ -0,0 +1,30 @@
...
+ * Calls the Lua function in the directory specified, stores the data
+ * in the Lua virtual stack for the getter functions to extract and return
+ * Parameters:
+ * - the object_t struct that contains the argument list to be passed to Lua
+ * - the Lua script directory
+ * Returns:
+ * - the Lua state with the return value of the Lua script extracted and placed
+ *   in the virtual stack
+ */
+lua_State *callLua(object_t *ot, char *lua_path);
```
sabinecodes15, line 28:
```suggestion
lua_State *call_lua(object_t *ot, char *lua_path);
```
The convention for c is snake case rather than camel case; good to keep this style consistent across function names

include/custom-scripts/lua_custom_type.h
```diff
@@ -0,0 +1,30 @@
...
+ * Parameters:
+ * - the Lua state created
+ * - the object containing the argument linked list
+ * Returns:
+ * - the number of arguments in the linked list
+ */
+int push_args(lua_State *L, object_t* ot);
+
+/**
+ * Helper function for the getter functions below
```
sabinecodes15, line 18: where are the getter functions below in this file?

include/action_management/action_structs.h
```diff
@@ -13,6 +13,8 @@
 enum actions {
     // KIND 1 ACTIONS - ACTION <item>
     OPEN,
+    FIREBALL,
```
sabinecodes15, line 16: As a design note, rather than including new actions as specific as "slash" and "fireball", you could consider theses actions  as USE-ing the item (the sword, the staff, etc.). This implementation could set the precedent of including a new weapon specific action every time a new weapon type is added, and that could get really unnecessarily convoluted. You could instead specify that the action is a "slash" or "fireball" in the short/long descriptions of the item.

sabinecodes15 (commented):
I made some comments on the custom scripts stuff, but ultimately the one thing I'd like to point out is [this](https://github.com/uchicago-cs/chiventure/pull/1640/files#r882342598). 

I won't explicitly approve this since I'm only needed for the action management stuff- a custom scripts member should be the one to give the approving review.

include/action_management/action_structs.h
ecoble: This is a great point, you can also see in the action management code that there is already a set up for action aliases, i.e. eat and drink are both just aliases for consume.

jackjackkimkim:
Ramarko was able to find a way to monkey patch the two custom actions "fireball" and "slash" so this pull request is no longer necessary.

PR #1641, opened 2022-05-26
Skilltrees/design and implement random skills

### LINKED ISSUE #1460: Design and Implement Random Skills
Design and implement random skills, another complex skill type which has an element of randomness to determine how, or if, it is executed. For instance, one skill could be "All-Out Attack: Deal 100 damage; misses 30% of the time." This effect succeeds or fails randomly. Another example would be "Chaos Beam: Equal chance to Stun, Freeze, or Burn target." This executes one of three potential outcomes. This will involve adding functions and documentation supporting these skills, and possibly adding some effects, such as random chance and/or random range effect. We will also need to evaluate how best to support a skill such as the chaos beam, which may require nesting or another sort of supporting effect.

There is further description of random skills here: https://github.com/uchicago-cs/chiventure/wiki/Skill-Trees-~-Integrating-complex-skill-(combined,-random,-sequential,-etc.)-implementation

team members: Annika, Arya, Chelsea

anchristensen:
This pull request will end issue #1470. In this PR, we designed and implemented a new type of complex skills, random skills as initially described here: https://github.com/uchicago-cs/chiventure/wiki/Skill-Trees-~-Integrating-complex-skill-(combined,-random,-sequential,-etc.)-implementation. This involved creating three new structs, one for each type of random skill (chance, range, switch), which contain a complex skill as well as individualized parameters such as, for random chance, a float with the chance of failure. Then, we wrote new, init, and free functions for each of these structs, as well as execute functions for each (which also involved writing some random generation code using the C rand() function, which may not be as perfectly random as a third party library, but works for now). Finally we added testing for all of these functions.


include/skilltrees/skilltrees_common.h
```diff
@@ -125,6 +125,41 @@ typedef struct complex_skill{
...
+    //Complex skills struct, type must be RANDOM_CHANCE
+    complex_skill_t* complex_skill;
+
+    // chance of failure of this skill
+    float chance_failure;
+
+} random_chance_type_t;
+
+/* Random Range complex type */
+typedef struct random_range_type {
```
nbilaver, line 139: What specifically does the random_range_type do in random skills? It isn't quite obvious from looking at the struct alone. Looking at its execute function I think it executes a random amount of the skills within the skills list of the complex struct, but I'm not quite sure. Maybe a comment to describe it could be helpful.

src/skilltrees/src/complex_skills.c
```diff
@@ -166,4 +166,270 @@ int complex_skill_xp_up(complex_skill_t* complex_skill, unsigned int xp_gained){
...
+        return FAILURE;
+    }
+
+    float total = 0;
+    for (int i = 0; i < switch_skill->complex_skill->num_skills; i++){
+        total += switch_skill->chances[i];
+    }
+
+    int value = random_float_generator(total);
+
```
nbilaver, line 420: Why is the first skill checked individually in it's own if statement, while the rest are in a for loop? Couldn't the for loop just start at i = 0 and do the same thing without this extra if statement?

include/skilltrees/complex_skills.h
```diff
@@ -142,4 +145,167 @@ int complex_skill_level_up(complex_skill_t* complex_skill);
...
 
+/*
+ * Allocates a random chance type in the heap.
+ *
+ * Parameters:
+ *  - chance_failure: A float representing the chance of failure
+ *  - complex_skill: the associated complex skill
+ *
+ * Returns:
+ *  - A pointer to the random chnace type, or NULL if one cannot be
```
jacknugent1529: spelling

include/skilltrees/complex_skills.h
```diff
@@ -142,4 +145,167 @@ int complex_skill_level_up(complex_skill_t* complex_skill);
...
+ * Returns:
+ *  - Always returns 0
+ */
+int random_range_free(random_range_type_t* random_range_skill);
+
+ /*
+ * Allocates a new random switch type in the heap.
+ *
+ * Parameters:
+ *  - chances: List of percentages for each subskill
```
jacknugent1529: Even if the length of the `chances` array should be determined by the number of subskills, I think you should still pass around a length parameter in order to verify that this is actually the case. Maybe you could just verify in the `init` function that the length is correct. Currently, if the game author mistakenly provides an array of the wrong size, it appears it will just crash.

include/skilltrees/complex_skills.h
```diff
@@ -142,4 +145,167 @@ int complex_skill_level_up(complex_skill_t* complex_skill);
...
+ */
+random_switch_type_t* random_switch_new(complex_skill_t* complex_skill, float* chances);
+
+ /*
+ * Initializes a random switch type.
+ *
+ * Parameters:
+ *  - random_switch_type_t: must point struct already allocated with 
+ *    random_switch_new
+ *  - chances: List of percentages for each subskill
```
jacknugent1529: see above

include/skilltrees/skilltrees_common.h
jacknugent1529: I agree. I think the header comments for each struct in this file could be expanded a bit

src/skilltrees/src/complex_skills.c
```diff
@@ -166,4 +166,270 @@ int complex_skill_xp_up(complex_skill_t* complex_skill, unsigned int xp_gained){
...
+    random_switch_type_t* random;
+
+   random = (random_switch_type_t*)malloc(sizeof(random_switch_type_t));
+
+    if (random == NULL) {
+        fprintf(stderr, "random_switch_new: memory allocation failed for random\n");
+        return NULL;
+    }
+
+    float* list = (float*)malloc(sizeof(float)*complex_skill->num_skills);
```
jacknugent1529: This is an array, so labelling it a list might be somewhat misleading

src/skilltrees/src/complex_skills.c
```diff
@@ -166,4 +166,270 @@ int complex_skill_xp_up(complex_skill_t* complex_skill, unsigned int xp_gained){
...
+
+/*See complex_skills.h */
+int execute_random_chance_complex_skill(random_chance_type_t* chance_skill, chiventure_ctx_t* ctx){
+    if (chance_skill->complex_skill->type != RANDOM_CHANCE){
+        return FAILURE;
+    }
+
+    int value = random_float_generator(100.0);
+    if (value < (chance_skill->chance_failure * 100)){
+        for (int i= 0; i < chance_skill->complex_skill->num_skills; i++){
```
jacknugent1529: style: spacing around `=`

src/skilltrees/src/complex_skills.c
```diff
@@ -166,4 +166,270 @@ int complex_skill_xp_up(complex_skill_t* complex_skill, unsigned int xp_gained){
...
+            skill_execute(chance_skill->complex_skill->skills[i], ctx);
+        }
+    }
```
jacknugent1529: In the documentation, I think you should specify that either none of the skills are executed, or all of them are executed.

jacknugent1529 (commented):
This is fantastic work! I'm sure documentation is being written for another task, but in the `complex_skills.h`, can you briefly explain at the top of the file what the various types of complex skills? Just a sentence or two for each type would helpful for others to understand the code.

chelseaegbarin:
Addressed Jack's suggestions by adding brief descriptions for each type of complex skill to the complex_skill.h file. Fixed various typos and style errors. Added a chances_len (length of chances array) parameter to the random switch functions and tests and accounted for the new parameter by checking that it is equal to th random skill's complex skill's num_skills field (because there should be a chance float for each skill).

jacknugent1529 (approved):
Looks good to me! Nice work on this issue!

src/skilltrees/src/complex_skills.c
anchristensen: Good point, I just fixed this!

nbilaver (approved):
Everything looks good now!

anchristensen:
Since this PR was reviewed, several things have changed. Descriptions of each type of random skill were added to the complex_skills.h file to explain what each type of random skill does, and documentation of the random chance skill was expanded to more clearly explain that either none or all of its subskills are executed. Some minor spelling and spacing fixes were made, and some variable names changed to clarify their purpose (ex: “list” changed to "chances_array”). A new parameter was added to the switch type to list the length of the chances array, to verify that the author is providing an array of the correct size. Finally, two loops that could have been in a single for loop were combined.

jacknugent1529:
Issue Score: Excellent

Comments: Excellent work on this PR! The code was very well written when you opened the PR, and you addressed the feedback well.

PR #1647, opened 2022-05-26
Sound: tts struct pull request

parkerbroz:
Created a tts struct which enables easier and simpler use of text to speech. We think long-term this struct will be useful for integrating the tts functionality into chiventure, as it is an effective middleman which takes on all the strain of interfacing with the espeak libraries. There is definitely room for expansion of the struct. Espeak has many fields and options which can be set, most of which were left as default here. Adding support for these would be fairly simple, essentially just modifying the data that tts_t stores and adjusting a few function calls in speak().


include/sound/tts.h
```diff
@@ -0,0 +1,29 @@
...
+/* This represents a text-to-speech object */ 
+typedef struct {
+    char* voicename; // The name of the speaker
+    /* voicename cannot be anything - you can find a list of all
+    possible voices by running espeak --voices. */
+    char* text; // The text which will be spoken
+    int buflength; // The buffer length in ms of sound buffers
+} tts_t;
+
+// malloc and fill a tts_t struct with the necessary data
```
jwschroeder21: This looks really good. I think it may be cleaner to have the comments all be in the same style. Maybe we should choose either /* */ style comments or // ?

src/sound/examples/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+if (Espeak_FOUND)
+    set(SOUND_EXAMPLES test-tts)
+
+    add_executable(test-tts
+                   test-tts.c)
+
+    # Link with chiventure libraries
```
jwschroeder21, line 7: Nicely done

jwschroeder21 (changes requested):
This all looks really great. I think some of the commenting on the code could be cleaned up and I'm happy to help with that. Otherwise great job!

include/sound/tts.h
parkerbroz: I generally tried to use /**/ for multiline comments or important headers. I think it gives a nice distinction that it's more significant than //. If you'd like, I can go through and change them but I think it just comes down to stylistic preference.

jwschroeder21: Okay, I'm convinced. Sounds good.

jwschroeder21 (approved):

src/sound/src/tts.c
```diff
@@ -0,0 +1,31 @@
+#include "sound/tts.h"
+#include <espeak-ng/speak_lib.h>
+#include <string.h>
+
+espeak_AUDIO_OUTPUT output = AUDIO_OUTPUT_SYNCH_PLAYBACK;
+char *path = NULL;
+void* user_data;
```
ecoble: Move asterisk over

include/sound/tts.h
ecoble: With header comments, please follow the format of:

Brief description of functions
Parameters:
- param1: description
- param2: description
- ...

Returns:
- brief description description of what is returned

ecoble (changes requested):
Looks good! Small changes!

ecoble (approved):
Nice job!

jwschroeder21:
Closing statement: we created a tts struct which enables easier and simpler use of text to speech. In the long term, this struct will be useful for integrating the tts functionality into chiventure. It is an effective intermediary which takes on all the hard work of interfacing with the e-speak libraries. 

There is still room for improvement. Adding support would be fairly simple, essentially just modifying the data that tts_t stores and adjusting a few function calls in speak().

ecoble:
Issue Score: Excellent

Comments:
Great job on these e-speak structs! This is great foundation for future use in Chiventure.

PR #1662, opened 2022-05-27
1625/am view skills

### LINKED ISSUE #1625: Implement a new action "view skills" along with the skilltrees team
Our goal here is to allow the user to view the names of the skills in their skilltree.

In order to implement this, the skilltrees team would have to create a display skilltree function in the game-state/player module that we can call from the action_management/actionmanagement module. We would have to modify the logic within the if-statement of our do_self_action function to accommodate this and to call this display function.

Because we won't be able to display all skills and their descriptions in one print to the terminal, we also want to implement a `"view skills "description""` command, which will call the skilltree team's `display_skill_description` function and print the description of a specific skill.

HarrisonD123:
We modified our `do_self_action` function to first take in multiple string inputs by converting `string *target` to an array of strings, `string **target`. We then modified our if-statement logic:

If the first string in the array is `"skills"` and the second element is null, we call the skilltree modules function `display_tree,` which prints all skill names out.

If the first string in the array is `"skills"` and the second element is a string, we call the skilltree function `display_skill_description,` passing in the second string as the skill name. This displays the description of the skill and will be used by the player to see what a particular skill does/means.

Harrison changed the parameters of the `do_self_action` function. Jackson and Sabine coded up the if-statement logic and called the skilltree team's functions.


include/action_management/actionmanagement.h
```diff
@@ -159,16 +159,17 @@ int do_item_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *direct,
...
  * - c: A context struct encapsulating the shared state in chiventure
  * - a: An action type struct
- * - target: A string describing what self-related object (STATS, etc)
+ * - target: An enum describing what self-related object (STATS, etc)
  *           needs to be acted on
  * - ret_string: A pointer to a string describing the result of the function
  *   - NOTE: THIS STRING IS MALLOCED AND MUST BE FREED BY USERS OF THIS FUNCTION
  *
  * Returns:
  * - 0 upon success, success string as an out parameter
```
chelseaegbarin: By "out parameter" do you mean what is printed to the terminal? Not a suggestion just trying to understand !

include/action_management/actionmanagement.h
```diff
@@ -159,16 +159,17 @@ int do_item_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *direct,
...
  * - WRONG_KIND if the action type has the wrong kind, failure string as an out parameter
+ * - 6 if conditions for the action haven't been met, failure string as an out parameter
```
chelseaegbarin: Why does it return 6? Do you mean 1 or am I misunderstanding

chelseaegbarin: Okay wait I see why now! I would make these descriptions more uniform though because you specified WRONG_KIND in the previous return option but in this one you just used the number, and it wasn't until I went to the source file that I understood what that meant

chelseaegbarin (approved):
See comments :) Other than those it looks great. I think your changes were warranted and well-executed. I also appreciate how you left comments throughout your code, it made it easier to follow. My only suggestions/points of confusion were in the header file and are easy fixes :)

KendrickXie (approved):
Looks good to me.

include/action_management/actionmanagement.h
ecoble: Great point!

ecoble (changes requested):
Looks great overall, Chelsea brought up great points about making the header comments more clear.

include/action_management/actionmanagement.h
HarrisonD123: Just that the string is passed in by reference by the CLI team! They have a string whose address they pass in as a parameter. We assign the string, which allows CLI to use the value without our having to actually return it.

include/action_management/actionmanagement.h
HarrisonD123: Just got rid of that comment line. This function never returns 6.

ecoble (approved):
Great job!

ecoble:
Issue Score: Excellent

Comments:
Great job reviewing and updating this PR! Be sure to include closing statements.

Issue #1612, opened 2022-05-25
Create a Battle Item Stats Wiki Document to describe stat multipliers for each playerclass + item combination


Vincent-Zhang4135:
While working on Integrating Player Class/Item Interactions with Battle Logic, we realized that there are a lot of different specifications that need to detailed in regarding how certain battle items could affect **depending** on which playerclass the battle item is being used on. We should expect different multipliers for all the different stats depending on which playerclass-battleitem combination was interaction with, and so I have created a separate issue to develop the documentation for this.

Vincent-Zhang4135:
Jonathan has created a basic template for the warrior class stat effects after using a Strength Up consumable, and I have linked below the Player Class Battle Item Stat wiki that will include these multipliers below, and decided for simplicity, that I will work only with Warrior and Wizard player classes for now to make sure we have a work stats functionality first:

[Battle Item Stats](https://github.com/uchicago-cs/chiventure/wiki/Battles-~-Battle-Item-Stats)

jroven:
Vincent, Katie, and I met synchronously to come up with eight class-based multipliers for stat changes due to items. Integrated classes include warrior, wizard, bard, and rogue. Integrated items include the consumables strength up, defense up, elixir of life, and healing potion.

lilyehsani:
Issue Score: Excellent

Comments:
Great work on this issue!

PR #1665, opened 2022-05-27
Integrate player class/item interactions with battle logic: implement multipliers in `apply_item_stat_changes()` in battle_logic.c

### LINKED ISSUE #1024: Integrating Player Class/Item Interactions with Battle Logic
In Issue #959, player classes and items were given more detailed interactions through the addition of multipliers, which allow certain player classes to use an item more (>1.0x) proficiently, less (<1.0x) proficiently, or be completely restricted (0x) from using an item.

Although this code is functional in the player class module, integration with other modules of chiventure is necessary in order for the multipliers to have an effect on game flow. The most pertinent use of items currently exists in the battle module, so my goal is to successfully integrate these interactions with pre-existing code in this module.

jroven:
Vincent, Katie, and I have implemented class-dependent multipliers on stat changes due to items as detailed in issue #1024. We split the function `apply_stat_changes()` into `apply_item_stat_changes()` and `apply_movement_stat_changes()` because we needed to change the parameters and are not dealing with movement at all. We also created a `class_item_stat_multipliers_t` struct, which has parameters of doubles that represent multipliers of each stat. The bulk of the changes occurred in the new function `class_multipliers()`, which changes the multipliers within the struct based on the combination of player class and item. Then, these multipliers are applied to the bonuses in `apply_item_stat_changes()`. We implemented eight multipliers as detailed on the [Battle Item Stats](https://github.com/uchicago-cs/chiventure/wiki/Battles-~-Battle-Item-Stats) wiki page in issue #1612.


src/battle/src/battle_logic.c
```diff
@@ -199,7 +199,104 @@ int apply_stat_changes(stat_t* target_stats, stat_changes_t* changes)
...
+    if (strcmp(class->name, "warrior") == 0) {
+        if (strcmp(item->name, "Strength Up") == 0) {
+            mults->phys_atk = 1.5;
+        }
+        if (strcmp(item->name, "Defense Up") == 0) {
+            mults->phys_def = 1.2;
+        }
+    }
+    if (strcmp(class->name, "wizard") == 0) {
+        if (strcmp(item->name, "Strength Up") == 0) {
```
wellingtj, line 343: why do certain items decrease stats? I'm a bit confused about that. Otherwise I don't really have any comments to give this. this is a good PR

wellingtj (approved):
solid code. I honestly looked through but I only had one really minor comment. This seems good

lilyehsani:
I have added @nmkhan100 from the battle team to review the changes to battle code.

src/battle/src/battle_logic.c
```diff
@@ -199,7 +199,104 @@ int apply_stat_changes(stat_t* target_stats, stat_changes_t* changes)
     return SUCCESS;
 }
 
-/* See Battle_logic.h */
+/* see battle_logic.h */
+int apply_item_stat_changes(class_t* class, stat_t* target_stats, battle_item_t* item)
```
nmkhan100, line 274: Could you not have used the stat_changes_t struct for the apply_item_stat_changes() function? It might make it easier for other people to understand what's going on if you changed your code to use the stat_changes_t struct and improve readability (just some more coding to implement this).

nmkhan100 (approved):
I have reviewed your code and approved these changes because I think everything looks good. I just made a comment for a part of the code that looks like it could be improved, but it won't change the functionality of the code, so it isn't really a serious issue for this PR (could be improved in the future to make the code easier to understand and reduce code clutter).

src/battle/src/battle_logic.c
Vincent-Zhang4135: its actually because this is a multiplier that describes how greatly an additional stat adds to the current stat. So for example, if the multiplier is 0.8, and your HP is 100, that is not actually saying that it will make your HP 80, instead, it is saying that it will make the multiple 100 + 0.8 * (some stat buffer), so its always increasing, but the intention is that some stats are less effective on some users than others

lilyehsani (approved):
This looks good! However, since it deals with battle code rather than playerclass code, I am going to hold off on merging it until after their demo is in and working.

Vincent-Zhang4135:
Jonathan, Katie, and I, worked on integrating `apply_stat_item_changes`, and successfully made this function add multipliers to the stat changes based on whichever item and whichever class the item is being applied to. The documentation for this can be found in more detail in [Battle Item Stats](https://github.com/uchicago-cs/chiventure/wiki/Battles-~-Battle-Item-Stats). Additionally, some of our tests had to be commented out to be functional for now, and we have included a separate issue for this: #1718

lilyehsani:
Issue Score: Excellent

Comments:
Great work on this PR! For a task of this size, we would've expected to see more status updates and/or status updates that were more detailed. I have applied this rubric item because it looks like the review comments were never responded to.

PR #1675, opened 2022-05-27
1628/am view quests

### LINKED ISSUE #1628: Implement a new action "view quests" along with the quests team
Our goal here is to allow the user to view the names of the quests.

In order to implement this, the quests team would have to create a display quests function in the `game-state/player` module that we can call from the `action_management/actionmanagement` module. We would have to modify the logic within the if-statement of our `do_self_action` function to accommodate this and to call this display function.

The quests team also wants the ability to view all of the tasks within a specific quest, using their quest id string. In other words, we also need to implement a `"view quests "quest-id""` command.

HarrisonD123:
We modified our do_self_action function to first take in multiple string inputs by converting string *target to an array of strings, string **target. We then modified our if-statement logic:

If the first string in the array is "quests" and the second element is null, we call the quests module's function show_quests, which prints all quest names out.

If the first string in the array is "quests" and the second element is a string, we call the quests function show_task_tree, passing in the second string as the quest name. This displays the tree of tasks that completing the specified quest entails.

If the first string in the array is "task" and the second element is a string, we call the quests function show_task, passing in the second string as the task name. This displays one specific task's of a particular quest and its description, so that the player knows what it entails.


include/action_management/actionmanagement.h
```diff
@@ -159,7 +159,8 @@ int do_item_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *direct,
  * Parameters:
  * - c: A context struct encapsulating the shared state in chiventure
  * - a: An action type struct
- * - target: A string describing what self-related object (STATS, etc)
+ * - target: An array of strings describing what self-related object
+ *           (quests, etc), and secondary-level object ("quest-id", etc)
```
Don-Osipov, line 163: Maybe a bit more clarification on what a secondary-level object is?

src/action_management/src/actionmanagement.c
```diff
@@ -389,18 +391,37 @@ int do_self_action(chiventure_ctx_t *c, action_type_t *a,
...
             // retrieve inventory from the player
             // TO BE IMPLEMENTED
-        } else if (strcmp(target, "skills") == 0) {
+        } else if (strcmp(target[0], "skills") == 0) {
             // retrieve skill tree from the player
             // TO BE IMPLEMENTED
+        } else if (strcmp(target[0], "quests") == 0) {
+            // retrieve quests from game
+            if(target[1] == NULL) {
+                string = show_quests(c->game->curr_player);
```
Don-Osipov: You use c->game->curr_player very often, maybe make it a variable that you reuse

Don-Osipov: You also have this line: game_t *game = c->game;
which goes unused as far as I see

Just in general try to make utilize variables some more in this function to improve readability.

Don-Osipov (commented):
Looks pretty good other than some minor documentation and code style changes. Nice job documenting what the code you added does - it's very clear.

src/action_management/src/actionmanagement.c
```diff
@@ -367,15 +369,14 @@ int do_item_item_action(chiventure_ctx_t *c, action_type_t *a, item_t *direct,
...
-                   char *target, char **ret_string)
+                   char **target, char **ret_string)
 {
     assert(c);
     assert(c->game);
     assert(a);
     assert(target);
-    
-    game_t *game = c->game;
-    target = case_insensitized_string(target);
```
ecoble, line 378: do you still need to case desensitize target?

ecoble (commented):
Just a quick question, otherwise looks good!

src/action_management/src/actionmanagement.c
JSerf02: Hi, I'm from the quest team and I was messing around with the changes from this PR in another branch to test our final demo. 

So I'm 90% sure you don't need to make it lowercase again since the strings in target are already lowercase. In fact, I'm pretty sure this line is actually a mistake that needs to get removed since target is a char ** while case_insensitized_string takes a char * as input (this causes segfaults in-game). Another issue with this happens later in the function as well. I posted fixes in your Slack channel yesterday if anyone can take a look.

sabinecodes15: Thanks for pointing this out! Just updated the code to reflect those changes.

src/action_management/src/actionmanagement.c
```diff
@@ -389,18 +390,37 @@ int do_self_action(chiventure_ctx_t *c, action_type_t *a,
...
+            // display the description of a specified task from a quest
+            if(target[1] == NULL) {
+                string = "Error: Please provide task name";
+            }
+            else {
+                string = show_task(target[1], c_player,
+                               c->game->all_quests);
+            }
         } else {
             sprintf(string, "%s cannot be viewed", target);
```
JSerf02: This is the other line with incorrect typing (char ** instead of char *). Should be target[0] I think

src/action_management/src/actionmanagement.c
HarrisonD123: Those issues have all been fixed!

HarrisonD123: CLI already lowercases all strings

src/action_management/src/actionmanagement.c
```diff
@@ -389,18 +390,37 @@ int do_self_action(chiventure_ctx_t *c, action_type_t *a,
...
+        } else if (strcmp(target[0], "skills") == 0) {
             // retrieve skill tree from the player
             // TO BE IMPLEMENTED
+        } else if (strcmp(target[0], "quests") == 0) {
+            // retrieve quests from game
+            if(target[1] == NULL) {
+                string = show_quests(c_player);
+            }
+            // retrieve the task tree from a specific quest
+            else {
```
JSerf02, line 413: It would likely make more sense in terms of quest interface if this were its own condition for when the user enters "quest" to distinguish more between printing multiple quests and an individual quest (so add another if statement just for this, print an error message if there is no second input in that statement, and call the first statement in the current "quests" if statement with no other checks). 
This is not the biggest deal though and it is definitely more important to get this PR in than to fix this if there's no time.

JSerf02 (changes requested):
Changes look good! I commented 2 changes. The bottom one is far more important since it causes segfaults if not fixed, but it would be nice if you can get to the top one as well.
Will approve once the mistyping error on line 425 is fixed.

src/action_management/src/actionmanagement.c
HarrisonD123: Ah thanks for catching that. It was fixed in another branch, but we have a lot of branches working on the same function, so it's hard to keep track of what was changed where. Fixed!

src/action_management/src/actionmanagement.c
HarrisonD123: Just added it, will push it soon.

JSerf02 (approved):

src/action_management/src/actionmanagement.c
JSerf02: Thank you!

HarrisonD123: Done!

Don-Osipov (approved):
All changes that were commented on by me earlier have been addressed. Looks good!

ecoble (approved):
Great job reviewing and updating this PR!

ecoble:
Issue Score: Excellent

Comments:
Great job on this issue, and especially with reviewing the updating a cross-team PR! Please be sure to include closing statements for PRs as well as issues.

PR #1686, opened 2022-05-28
Playerclass: created function, tests for prefab items associated with classes

### LINKED ISSUE #1656: Write new, modify existing helper functions for class_prefabs.c
Issue #1066 mentioned a number of functions that could be added to `class_prefabs.c`, which, since that issue is closed, I am pasting in here to start a new issue. I will add a `class_prefab_add_items()` function which adds some default items to each class (e.g. a wand for the wizard, a rusty sword for the warrior), by creating items, documenting them on the wiki, coding these items, and adding tests. I will also separate the existing `class_prefab_new()` function so that it is possible to create classes which only contain a short and long description. I may also create a function where the dev can specify which fields of the prefab class they would like filled out, which will be more relevant as there are more ways to prefab classes (i.e. prefab items), but this might get pushed to a later issue depending on time.

vanolson:
In `class_prefabs.h` and `class_prefabs.c`, I wrote a function, `class_prefab_add_items()`, which creates and return an item based on an inputted playerclass. The function creates a starter item for each non-NPC playerclass in the prefabs (it also includes a condition for basic, but doesn't create an item. This could be filled later.). For NPCs, or classes not found in the prefabs, it should return `NULL`. I also wrote tests for this function for each prefab condition in the `test_class_prefabs.c` file, which contains a helper function, `check_item_presence()`, to reduce duplicate code. The item creation could later be expanded to include more fields, but the function is complete.


wellingtj:
Van did most of the work for this one. He made and coded all of the prefabs, but I helped come up with some of the ideas, and I also handled documentation

jroven (approved):
This looks great! I really like the item prefabs and descriptions that you made

src/playerclass/src/class_prefabs.c
```diff
@@ -936,3 +938,168 @@ int class_prefab_add_skills(class_t* class) {
...
+    else if (!strncmp(temp_name, "sorceror", MAX_NAME_LEN)) {
+        /* Create item */
+        item_t* item_0 = item_new("healing stone", "A healing stone", 
+                                  "An ancient stone that rejuvenates the "
+                                  "sorceror from damage.");
+
+        return item_0;
+    }
+
+    else {
```
lilyehsani: You should either add items for all the prefabs (including the new NPC ones) or make this clause not error.

lilyehsani (changes requested):
Very cool addition! One comment.

src/playerclass/src/class_prefabs.c
vanolson: removed the error message, though I'm not fully sure that's what you intended?

lilyehsani (approved):
LGTM! Approved.

vanolson:
In the `class_prefabs` header and C files, I created the `class_prefab_add_items()` function, which has base items for the non-NPC player classes we have prefabricated. Jake Wellington helped me with the ideas for these items, as well as helping to document this new part of the prefab module. The function returns `NULL` for NPCs and classes not found in the prefabs. NPC items could be created as part of a future issue. In addition to creating the items, I also wrote tests for all of them and a helper function for those tests. In responding to feedback, I removed the error message that came along with an unfound class and resolved a merge conflict.

lilyehsani:
Issue Score: Excellent

Comments:
Great work on this PR!

PR #1695, opened 2022-05-28
Quests/wdl file branch

JSerf02:
Made gaining XP increase the player's level and added our final presentation WDL document


JSerf02:
Don't pull this in yet, there are strange bugs with the demo (will be done by the end of tomorrow)

include/game-state/game_state_common.h
```diff
@@ -38,7 +38,4 @@
 #define ROOM_SRC_NULL 2
 #define ROOM_DEST_NULL 3
 
-//Macros for quests
-#define QUEST_NAME_MAX_LEN 100
```
elizabethli31: shouldnt this still exit and be max length 44?

include/npc/dialogue.h
```diff
@@ -3,6 +3,7 @@
 
 #include "game-state/condition.h"
 #include "cli/util.h"
+//#include "quests/quests_state.h"
```
elizabethli31: remove this comment

include/game-state/game_state_common.h
JSerf02: That's a change from a prior branch that came up again because of merging issues I think. What happened with that was we actually have to use QUEST_NAME_MAX_LEN in a bunch of different modules and none of them actually include this file. Maybe we should change it so they do include this file instead of defining it directly, but the direct defines were already in the files and we didn't want to introduce more dependency issues so we just removed it here

src/quests/examples/quest_game/steve.wdl
```diff
@@ -0,0 +1,704 @@
...
+                {
+                    "Task Name": "Kill Steve",
+                    "Mission": {
+                        "Target Name": "Steve",
+                        "Type": "Meet NPC"
+                    },
+                    "Prerequisites": {
+                        "Level": 8,
+                        "Quests": ["Find Steve"],
+                        "Tasks": ["Obtain Knife"]
```
elizabethli31, line 172: just wanted to double check - is obtain knife supposed to be a side task that's not included in any of the quests?

JSerf02: Yeah it's supposed to showcase our system for automatically creating quests for tasks that are in prereqs but not in any task trees

src/wdl/src/validate.c
```diff
@@ -426,6 +426,57 @@ int node_action_type_check(obj_t *obj)
...
+{
+    int id = (obj_get_type(obj, "Task") == TYPE_STR);
+    if (dialogue_type_check(obj_get(obj, "Dialogue")) != SUCCESS)
+    {
+        return FAILURE;
+    }
+    return SUCCESS;
+}
+
+/*int quest_list_type_check(obj_list_t *obj)
```
elizabethli31: is this supposed to all be commented out?

JSerf02: Just removed them. Those are deprecated validation functions from when the npc wdl was going to work a little differently. We can maybe add actual ones for the new version to make sure the string exists, but it really isn't the biggest deal considering these objects only have 2 types of info and one of them is already being type-checked

elizabethli31 (commented):
I don't see any obvious issues with this but I'll wait until there's no bugs or merge conflicts before reviewing again and approving. This is so well thought out and perfectly complex enough to show off all the changes we've made with quests this year and its integration with other features! Storyline was great too - glad steve's baby photos are going to be exposed. Over all, just really really well done!

JSerf02:
Branch is finally done, and with that, the quest module is ready for the final presentation!

elizabethli31 (approved):
Changes look great and the demo we practiced went smoothly without any issues!

philipchenaux (approved):
Looks good to merge! Well done guys, and great work this quarter!

PR #1710, opened 2022-05-30
Fix extraneous print statement in `test_operations.c`

majorsylvie:
removing extranueous testing print statement from test_operations.c


jacknugent1529 (approved):
Thanks for fixing this!

bzeng6 (approved):
good job

PR #1712, opened 2022-05-31
cli: Parser Changes DEMO PR

### LINKED ISSUE #1592: cli: make WDL file that showcases what we've worked on!
wdl file should have player info, 
places with longer names to be parsed as "[longer name here]"

milesgardner:
The CLI team is excited to demo some of the progress we've made on chiventure!  We've made a simple WDL file, which we'll be using to showcase some of the changes we've made to the existing parser module.


tests/wdl/examples/wdl/parser-changes.wdl
```diff
@@ -0,0 +1,98 @@
...
+            "connections": [
+                {
+                    "to": "Home_base",
+                    "direction": "SOUTH"
+                }
+            ]
+        },
+        "Home_base": {
+            "short_desc": "This is Home Base",
+            "long_desc": "This is Home Base. It's empty, but there is an escape to the north.",
```
majorsylvie: Still says escape to the north, when the connection is to the west. 
So the player goes south than is told to go north then can't.

Fix the description!

majorsylvie (changes requested):
Ran it on my own computer, the directions and room connections work.
Adjust the one comment on room description.

Also! I am not sure what you are wanting to test with the Levenshtien difference. I would really appreciate some kind of README or directions of what is actually meant to be run in this part of the demo. You can add this to the demo2022 folder for example. So it's clear what is actually wanted for testing.

tests/wdl/examples/wdl/parser-changes.wdl
mrestuccia1: Fixed comment

jacknugent1529 (approved):
Looks good to me!

majorsylvie (approved):

